Issue #2, opened 2020-04-27
Operator Provisions an Always On Virtual Office Space on AWS


zspencer:
```features/video-workspaces.feature
Feature: Video Workspaces
  In order discuss things face-to-face within my organization
  As Colin Collaborator and Felicia Faciliator
  I want a Video Workspace

  @0.1-alpha
  Scenario: Operator Provisions an 8 person Video Workspace on AWS
    Given an Operator has ran `jitsi/build` for {{clientDomain}}
    When an Operator runs the `jitsi/provision` command with:
      | arguments                        |
      | --client-domain={{clientDomain}} |
    Then a JITSI meet instance is available at https://{{clientDomain}}

  @0.1-alpha
  Scenario: Operator Builds an 8 person Video Workspace on AWS
    When an Operator runs the `jitsi/build` command with:
      | arguments                        |
      | --region=us-west-1               |
      | --provider=aws                   |
      | --client-domain={{clientDomain}} |
    Then a convene-jitsi-{{clientDomain}} is available within the us-west-1 region
```

Deploy Instance: 
- [x] Automate launch an instance
- [x] Automate apply a ssl certificate
- [x] Automate configure network setting ( Look into Terraform )

Release Instance:
- [x] Automate point DNS record to the instance

Install Jitsi
- [x] Automate interact with Jitsi interface

maximegalon5:
Who is the Operator? Does the Operator have a _technical_ background? If so, then this looks straightforward to me. I'm guessing the intention is to translate arguments into UI equivalents making it easier for non-technical Operators too?

user512:
@zspencer Do you know if we have a ssl certificate to test with?

zspencer:
You will probably want to self-sign a certificate for testing purposes (unless that doesn't work, in which case we should buy a wildcard *.zinc.coop cert.

zspencer:
@user512  - Would you be willing to go through feature test end-to-end by hand and confirm that you're able to spin up a brand new convene installation without any trouble? And if you _do_ experience trouble, could you add some check-boxes for us to pick off over the next week or two?

maximegalon5:
@zspencer I'm reading this issue as Vivek can go to meet.zinc.coop and create Vivek's Music room and then share that link. If this is the correct interpretation, what action do you need from me to _accept_ /_confirm_ this?

zspencer:
[...]

PR #3, opened 2020-04-28
Build AMI using Packer

user512:
LinK: https://github.com/zinc-collective/convene/issues/2
* TODO: Inspect error msg when provisioning
* Keep reading: https://www.packer.io/intro/getting-started/provision.html


example.json
```diff
@@ -0,0 +1,36 @@
...
+      "most_recent": true
+    },
+    "instance_type": "t2.micro",
+    "ssh_username": "ubuntu",
+    "ami_name": "packer-example {{timestamp}}"
+  }],
+  "provisioners": [{
+    "type": "shell",
+    "inline": [
+      "sleep 30",
```
user512, line 27: > Note: The sleep 30 in the example above is very important. Because Packer is able to detect and SSH into the instance as soon as SSH is available, Ubuntu actually doesn't get proper amounts of time to initialize. The sleep makes sure that the OS properly initializes.
https://www.packer.io/intro/getting-started/provision.html#configuring-provisioners

jitsi/provision
```diff
@@ -0,0 +1,30 @@
...
+    ;;
+esac
+done
+
+packer build \
+  -var 'instance_type=${INSTANCE_TYPE}' \
+  -var 'ssl_certificate=${SSL_CERTIFICATE}' \
+  -var 'public_domain=${PUBLIC_DOMAIN}' \
+  -var 'include_nginx=${INCLUDE_NGINX}'\
+  example.json
```
user512, line 30: Consider making `jitsi/provision` to take different VM provider as argument and choose different `aws-ebs.json` or `digital-ocean.json` but sticking with AWS only for now.

example.json
```diff
@@ -0,0 +1,36 @@
...
+        "virtualization-type": "hvm",
+        "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
+        "root-device-type": "ebs"
+      },
+      "owners": ["099720109477"],
+      "most_recent": true
+    },
+    "instance_type": "t2.micro",
+    "ssh_username": "ubuntu",
+    "ami_name": "packer-example {{timestamp}}"
```
zspencer, line 22: could we make this `convene-jitsi`?

example.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "builders": [{
+    "type": "amazon-ebs",
+    "access_key": "{{user `aws_access_key`}}",
+    "secret_key": "{{user `aws_secret_key`}}",
+    "region": "us-east-1",
+    "source_ami_filter": {
+      "filters": {
+        "virtualization-type": "hvm",
+        "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
```
zspencer, line 14: Would it be worth updating this to the ubuntu 20.04? or is that tooooo soon to adopt a new LTS?

example.json
```diff
@@ -0,0 +1,36 @@
+{
+  "variables": {
+    "aws_access_key": "{{env `AWS_ACCESS_KEY`}}",
+    "aws_secret_key": "{{env `AWS_SECRET_KEY`}}"
+  },
+  "builders": [{
+    "type": "amazon-ebs",
+    "access_key": "{{user `aws_access_key`}}",
+    "secret_key": "{{user `aws_secret_key`}}",
+    "region": "us-east-1",
```
zspencer, line 10: We may want to parameterize this so that people can set the region when instantiating the AMI

NOTE.md
```diff
@@ -0,0 +1,50 @@
...
+                "ec2:DescribeImageAttribute",
+                "ec2:DeleteSecurityGroup",
+                "ec2:DescribeSubnets",
+                "ec2:DeleteKeyPair"
+            ],
+            "Resource": "*"
+        }
+    ]
+}
+```
```
zspencer, line 50: This is super useful! Perhaps it's something we should add to the README.md?

example.json
zspencer: I wish JSON had comments :D

example.json
```diff
@@ -0,0 +1,36 @@
...
+    "source_ami_filter": {
+      "filters": {
+        "virtualization-type": "hvm",
+        "name": "ubuntu/images/*ubuntu-xenial-16.04-amd64-server-*",
+        "root-device-type": "ebs"
+      },
+      "owners": ["099720109477"],
+      "most_recent": true
+    },
+    "instance_type": "t2.micro",
```
zspencer, line 20: We probably want to make this parameterized as well.

jitsi/install
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+sleep 30
+echo 'deb https://download.jitsi.org stable/' | sudo tee -a /etc/apt/sources.list.d/jitsi-stable.list
+sudo wget -qO -  https://download.jitsi.org/jitsi-key.gpg.key | sudo apt-key add -
+sudo apt-get install apt-transport-https debconf-utils
```
zspencer, line 6: It looks like the package we want is `debconf`: https://packages.ubuntu.com/search?keywords=debconf

zspencer, line 6: We may need to run the `sudo apt-get update` before we do the `sudo apt-get install` in order for it to find the debconf-utils package.

jitsi/install
```diff
@@ -0,0 +1,26 @@
...
+cat << EOF | sudo debconf-set-selections
+jitsi-videobridge	jitsi-videobridge/jvb-hostname string	meet.zinc.coop
+jitsi-meet-prosody	jitsi-videobridge/jvb-hostname string	meet.zinc.coop
+jitsi-meet-web-config	jitsi-meet/cert-choice select I want to use my own certificate
+EOF
+
+sudo apt-get -y install jitsi-meet <<EOF
+test placeholder
+test
+EOF
```
zspencer, line 24: Did this work to cat the string values into the UI?

user512: Unfortunately it doesn't.

zspencer:
I think this is some phenomenal work! I'm going to sprout a few issues to clean up the rough edges!

PR #5, opened 2020-05-01
🔧📝✅ Install cucumber-js and move feature definition to repo

zspencer:
See: https://github.com/zinc-collective/convene/issues/2

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,9 @@ efforts; but requires people or organizations to maintain their own
 infrastructure tooling or pay privately held companies to host the services they
 use.
 
-As a result, much of the software tools to safely, securely, efficiently, reliably and
-self-sufficiently host public-domain services is unavailable to the public.
+As a result, much of the software tools to safely, securely, efficiently,
+reliably and self-sufficiently host public-domain services is unavailable to the
+public.
 
 Convene provides software tools to reliably, safely, and securely provision,
 deploy, host and scale public-domain software services for personal, communal,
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -57,6 +58,15 @@ Paid support is available for \$135 USD per hour<sup>[1][footnote-1]</sup>.
...
 [issue-tracker]: https://github.com/zinc-collective/convene/issues
 
+## Architecture
+
+At present, Convene is split into three modules:
+
+- `infrastructure`, which contains infrastructure management code for Packer,
+  Terraform and Ansible.
+- `web` which provides the human interface for Convene.
+- `api` which stores data long-term and provides a programmatic interface.
```
zspencer, line 68: @cjerozal / @user512 - Does this make sense from a "how do we want to split out responsibilities" for now? I feel like the majority of the work will be in the `infrastructure` module; but I also expect that as we start building out WebRTC UI bits that we will want those to live in something like `web` so folks don't need to know too much about Packer and such.

`api` may be extraneous; as we could most likely bundle that in with web.

features/infrastructure/hosting-jitsi.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Jitsi Hosting
```
zspencer, line 1: @user512 / @cjerozal / @maximegalon5  - I'm trying to use feature-driven development so that it's possible to focus on defining what we want to accomplish at the appropriate level of specificity so folks that aren't particularly code-y can riff on it. It also could serve as a reasonable source of usage documentation on a feature-by-feature basis; even if we don't wind up automating them.

I'm curious if this feels like a useful way for us to communicate more asynchronously about intent?

user512: I think this is useful.

README.md
user512: I think this makes sense.

PR #16, opened 2020-05-22
Automate acquiring an SSL certificate when provisioning a JITSI instance

zspencer:
See: https://github.com/zinc-collective/convene/issues/2

I may have conflated too many things, but as I was working on this it became a bit clearer that we would want to keep the terraformy bits separate out of the other non-infrastructure management bits; so I put them into the `infrastructure` folder that the README references.

The other major thing that changed here is now each "Client" will want a folder in `infrastructure/clients/their-domain.example.com` with their particular terraform configuration.

I _think_ once we have two to five clients, we'll be able to find the patterns in the terraformy configs and consolidate them. 

We may also wind up wanting to move meet.zinc.coop off of AWS and onto a service like Vultr which has _significantly_ less expensive data-transfer costs.


zspencer:
I'm kind of tempted to split the moving and the changing into separate PRs so we don't lose history on the files... Thoughts @user512?

infrastructure/features/hosting-jitsi.feature
```diff
@@ -0,0 +1,21 @@
...
+    Then a JITSI meet instance is available at https://{{clientDomain}}
+
+  @0.1-alpha
+  Scenario: Operator Builds a Clients single-server JITSI AMI on AWS
+    When an Operator runs the `jitsi/build` command with:
+      | arguments                        |
+      | --region=us-west-1               |
+      | --provider=aws                   |
+      | --client-domain={{clientDomain}} |
+    Then a convene-jitsi-{{clientDomain}} is available within the us-west-1 region
```
maximegalon5, line 21: Does _us-west-1 region limit us_ or is there a conflict between it and say "region": "{{env `REGION`}}" in the aws.json?

zspencer: Nope! This is an example for how we populate the `{{env REGION}}` in the `aws.json` file.

user512:
> I'm kind of tempted to split the moving and the changing into separate PRs so we don't lose history on the files... Thoughts @user512?

I think this is fine. 👍

PR #22, opened 2020-05-29
Reframing to center the feature of Video WorkSpaces

zspencer:
Hosting Jitsi isn't the important part, it's an incidental detail.

This reframes to Video Workspaces, which gives us more space to play.

### DIFF
--- a/infrastructure/features/video-workspaces.feature
+++ b/infrastructure/features/video-workspaces.feature
@@ -1,10 +1,10 @@
-Feature: Jitsi Hosting
-  In order to provide video chat to my organization
-  As an Operator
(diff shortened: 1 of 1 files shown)


maximegalon5:
Something to consider from a framing perspective. A video workspace limits us (from a branding and positioning perspective) and while we may be too early, a virtual workspace is really what we are creating, imo.

zspencer:
Basically, I am thinking that 0.1 goes hard on Video, 0.2 hits Scheduling, 0.3 hits Accessibility, and 0.4 expands Dev Tooling, which opens us up to 0.5+ giving us flexibility to create beautifully integrated experiences.

Issue #8, opened 2020-05-08
🌫️ `Spaces`: Purchasing a Space


user512:
## Scenarios:

- [ ] Guest Purchases a Space
  - [ ] Guest submits Down Payment  (See: #186)
  - [ ] Guest Submits Form with Space Details
       - [ ] Space Owner may Embed Form (See: #619)
  - [ ] Operators are emailed when a Space Request is submitted

- [ ] Resident purchases a Space
- [ ] Utilities: Stripe (TODO: Create a Ticket for Stripe Utility)
  - [ ] Stripe Hookup exposes products, plans, and prices 
- [ ] Furniture: Order Form (TODO: Create a Ticket for an Order Form furniture)

zspencer:
This is distinct from #9 in that we can be successful with this purely by allowing our creators to create an always-on video room.

This will build on top of the work we're doing in https://github.com/zinc-collective/convene/issues/2 to automate the provisioning of a Jitsi instance by creating affordances for Organizers to create a video room they can use at any time of day.

#8 is  more _expensive_ than #9 for Organizers, as each Video Room is an independent Jitsi server. AWS's pricing for a T3 micro is about $15/mo, which means we'd need to charge $45/mo; and we don't know how many folks can "fit" in that room.

But it also allows us to package up the work we're doing to automate provisioning and deploying of Jitsi instance(s) in #2 into a Minimal (potentially) Viable Product without adding the scope of scheduling.

I'm considering a few scenarios, and would appreciate thoughts from @zinc-collective/convene-maintainers about which of these (or others!) would be reasonable next steps for this epic.

```features/permanent-video-rooms.feature
Scenario: Organizer creates a tiny paid video room 
  Given an Organizer creates a tiny Video Room
  When Convene programaticaly provisions the tiny Video Room
  Then the Organizer can use the Video Room at any time
```

This scenario is slightly different, in that we assume there is no automation beyond the gathering of the Organizer's payment information and contact information.

The work of provisioning the Video Room is performed manually by an Operator (@user512 or I), and we would replace this scenario with an automated one once we've confirmed the market actually is willing to pay for private, secure video rooms.

```features/permanent-video-rooms.feature
Scenario: Organizer requests we provision a tiny paid video room
  Given an Organizer requests a tiny Video Room
  When an Operator provisions the Video Room
  Then the Organizer can use the Video Room at any time
```

Thoughts @zinc-collective/convene-maintainers?

maximegalon5:
[...]

PR #26, opened 2020-05-31
Automate creation of the Convene pricing tier

zspencer:
See: https://github.com/zinc-collective/convene/issues/8

I was using this primarily to explore what it would look like to build
Product and Pricing setup into Compensated/Support, while also ensuring
we have a papertrail for setting up pricing for Convene.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,4 @@
+# For local development, get these from https://dashboard.stripe.com/test/apikeys
+# For production, get them from https://dashboard.stripe.com/apikeys
+STRIPE_API_KEY=PLACE_STRIPE_SECRET_KEY_HERE
+STRIPE_PUBLISHABLE_API_KEY=PLACE_STRIPE_PUBLISHABLE_KEY_HERE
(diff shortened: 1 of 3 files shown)


zspencer:
@user512 / @cjerozal - Programmatic application of an idempotent pricing configuration feels like something that may be worth moving into the Compensated project at some point. Do you have thoughts on that?

user512:
I am confused on what we are trying to do here and some usage example might help me better understand.
Is `bin/configure-pricing` intended to be ran by developer and copy the output to zinc.coop?

zspencer:
@user512 - Yep! Keep in mind, part of Convene's business strategy is we want folks to be comfortable setting up their own "franchise" of Convene. This way someone could have an Oakland City Convene or a New Zealand Convene.

Franchisees pay Zinc for the right to use Convene within an LLC or other legal entity, while people who are willing to take on all the liability personally can use the Convene code free-of-charge to manage their own instances.

The way I think compensated could fit in with this is documented in [Developer Creates Products and Prices in Stripe](https://github.com/zinc-collective/compensated/issues/79). 

Essentially, we could help folks get started running a business by using compensated to manage their portfolio of products and prices across sales channels (Apple, Stripe, Android, etc).

PR #34, opened 2020-07-02
Releasing meet.zinc.coop no longer requires code changes

zspencer:
Now whenever we do a build for meet.zinc.coop, a provision will use the
updated AMI without requiring us to update the variables.

### DIFF
--- a/infrastructure/clients/meet.zinc.coop/infrastructure.tf
+++ b/infrastructure/clients/meet.zinc.coop/infrastructure.tf
@@ -30,15 +30,8 @@ provider "aws" {
   region = "us-west-1"
 }
 
-variable "jitsi_ami" {
-  type = string
-}
-
 resource "aws_instance" "convene_video" {
-  # TODO: Can we _infer_ this value or somehow detect the AMI based upon it's name?
-  #       Because every time we run `jitsi/build` it creates a new AMI on a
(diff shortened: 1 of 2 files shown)


infrastructure/clients/meet.zinc.coop/infrastructure.tf
```diff
@@ -50,6 +43,17 @@ resource "aws_instance" "convene_video" {
   key_name = "deployer-key"
 }
 
+data "aws_ami" "convene_ami" {
+  most_recent      = true
+  # name_regex       = "^convene-jitsi-*"
```
user512, line 48: Do we intend to remove this line or this is useful?

zspencer: Hmm, good quesiton. I think we can delete the line completely; but I _think_ this is what is sent to AWS to filter down the list of images that are returned. We probably want to use this in place of the `filter` block below; since our selection needs are relatively limited.  But I could be [reading the docs wrong](https://www.terraform.io/docs/providers/aws/d/ami.html#name_regex)

zspencer: Nope I have it backwards, we can delete this and rely on the `filter` block.

Issue #29, opened 2020-06-20
Multiple Operators may SSH into Provisioned Convene instances


zspencer:
```features/managing-convene.feature
In order to not carry the full weight of supporting Convene
As an Operator
I would like other Operators to be able to manage and maintain Convene instances

@skip-ci 
Scenario: Multiple Operators may log in to Provisioned Convene Instances
  Given a running Convene instance that was launched from a packer image with the following ssh authorized keys:
   | user | key |
   | zee  |  https://github.com/zspencer.keys |
   | tomlee |  https://github.com/user512.keys |
  When I ssh into the running instance using my private key
  Then I am logged into the running instance
```

## Technical Notes
Currently, we build the AMI without any SSH keys, and create a public key on AWS when we provision the instance.

Instead, we could update the build script so AMIs built for Convene allow logins from @user512 and @zspencer, ideally using individual usernames and public github keys

Docs: 
- [Ansible's Authorized Key Module Docs](https://docs.ansible.com/ansible/latest/modules/authorized_key_module.html)
- [Terraform's Ansible Provisioner Docs](https://www.packer.io/docs/provisioners/ansible)

zspencer:
This works for Workspaces built for both AWS and Vultr! Great work @user512!

PR #44, opened 2020-07-09
Spike on using Vultr

user512:
* TODO: Automatically fetch snapshot id to provision server
* TODO: Still need to figure out a way to get ssh_key_ids from terraform before packer build?

### DIFF
--- a/infrastructure/clients/convene-test-vultr.zinc.coop/infrastructure.tf
+++ b/infrastructure/clients/convene-test-vultr.zinc.coop/infrastructure.tf
@@ -6,7 +6,6 @@ variable "cloudflare_email" {
   type = string
 }
 
-
 variable "cloudflare_api_key" {
   type = string
 }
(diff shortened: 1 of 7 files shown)


infrastructure/clients/convene-test-vultr.zinc.coop/infrastructure.tf
```diff
@@ -0,0 +1,84 @@
...
+variable "vultr_api_key" {
+  type = string
+}
+
+provider "vultr" {
+  api_key = var.vultr_api_key
+}
+
+resource "vultr_server" "convene_vultr_video" {
+  snapshot_id = "b765f07738a83"
```
zspencer, line 42: Could we parameterize this to a variable? Or can we use a [vultr_snapshot](https://www.terraform.io/docs/providers/vultr/d/snapshot.html) data source to filter to the `convene-jitsi-convene-test-vultr.zinc.coop*` description we use when building the image?

zspencer:
Re: Keys - I wonder if we want to consider implementing #29 sooner rather than later; and using [Ansible's authorized key module](https://docs.ansible.com/ansible/latest/modules/authorized_key_module.html) to create users for each of us instead of relying on logging in as root/ubuntu.

Thoughts?

infrastructure/clients/convene-test-vultr.zinc.coop/infrastructure.tf
user512: Couldn't figured out a way to do `convene-jitsi-convene-test-vultr.zinc.coop*` but parameterize Vultr Snapshot id.

zspencer: That's fine! This is great work!

user512:
What do we gain by creating users instead of root?

zspencer:
For one, it means we can both SSH into the same machine; for two it gives an audit log of who did what on the server when we sudo things.

There's a great thread here explaining the implications of signing in as root - https://askubuntu.com/questions/16178/why-is-it-bad-to-log-in-as-root

user512:
Yes, that makes sense. The SSH into the same machine should be required before Andromeda release.

PR #45, opened 2020-07-17
Allow multiple users to ssh into Vultr instance

user512:
Connects: #29 
Allow operators in `operators.yml` to ssh into Vultr instance.


infrastructure/operators.yml
```diff
@@ -0,0 +1,32 @@
+---
+- hosts: default
```
zspencer, line 2: Could we link back to the host alias documentation in the packer ansible provisioner?

infrastructure/hosts
```diff
@@ -0,0 +1,2 @@
+[default]
+104.238.181.120
```
zspencer, line 2: We don't need to commit this file.

infrastructure/operators.yml
```diff
@@ -0,0 +1,32 @@
...
+      github: zspencer
+
+  become: yes
+  become_method: sudo
+  tasks:
+  - name: "make sure SSH is allowed"
+    ufw:
+      rule: allow
+      port: 22
+      proto: tcp
```
zspencer, line 19: I think the UFW is already configured to allow port 22; we probably don't need this?

PR #52, opened 2020-07-26
Expand view height to entire screen and allow mobile Jitsi video

user512:
Ref: https://github.com/jitsi/jitsi-meet/blob/7bfb2fc21907b2de499b3bed6fe5a4149fecad65/config.js#L489-L491
Screen Shot 2020-07-26 at 12 50 24 PM

### DIFF
--- a/convene-web/app/views/lobbies/show.html.erb
+++ b/convene-web/app/views/lobbies/show.html.erb
@@ -7,4 +7,4 @@
   </ul>
 </div>
 
-<div id="meet"></div>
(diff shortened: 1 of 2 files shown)


convene-web/app/views/lobbies/show.html.erb
```diff
@@ -7,4 +7,4 @@
   </ul>
 </div>
 
-<div id="meet"></div>
+<div id="meet" class="h-screen"></div>
```
zspencer, line 10: Nice work!

PR #54, opened 2020-07-26
Document Locking Rooms feature

zspencer:
I sprouted some initial parameter types so that we can drape additional
behavior for particular actors or rooms and what not.

### DIFF
--- a/features/locking-rooms.feature
+++ b/features/locking-rooms.feature
@@ -36,7 +36,7 @@ Feature: Locking Rooms
   Scenario: Workspace Member who does not know the Room Key cannot enter Locked Room
     Given a Workspace with a Locked Room
     When a Workspace Member provides the wrong Room Key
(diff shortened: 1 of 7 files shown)


features/locking-rooms.feature
```diff
@@ -0,0 +1,49 @@
...
+
+  Scenario: Workspace Member who knows the Room Key enters Locked Room
+    Given a Workspace with a Locked Room
+    When a Workspace Member provides the correct Room Key
+    Then the Workspace Member is placed in the Room
+
+  Scenario: Workspace Member who does not know the Room Key cannot enter Locked Room
+    Given a Workspace with a Locked Room
+    When a Workspace Member provides the wrong Room Key
+    Then the Workspace Member is placed in the Room
```
user512, line 39: I think this should be `Then the Workspace Member is not placed in the Room`.

PR #56, opened 2020-07-27
Discovering Room UI with the ability to collapse video div when hang up

user512:
Not exactly like the mock (haven't implement the back button), but I think this is a mergeable state and would love to get feedback on the direction.

screencast 2020-07-28 15-51-52


convene-web/app/controllers/workspaces_controller.rb
```diff
@@ -0,0 +1,9 @@
+class WorkspacesController < ApplicationController
+  def show
+    @rooms = current_workspace.rooms
+  end
```
zspencer, line 4: I have a relatively strong preference to expose as few instance variables as possible in views. Could we instead expose the `workspace` and eager load the rooms; then rely on `workspace.rooms` to access the rooms in the UI?

This reduces coupling between the template layer and the controller layer.

user512:
There isn't an article that I can reference to using `CustomEvent` but I read [Better Stimulus -> SOLID](https://www.betterstimulus.com/solid/single-responsibility.html) and noticed my original logic was mixing Jitsi API logic and handling enter room click event and end up having a global Stimulus controller which I think broke the single responsibility principle, and somehow I related this to my experience in a giant monolithic Rails app and refactor it into a multi Rails engine app using `ActiveSupport::Notifications` as an event system and drive me into using browser `CustomEvent`.

user512:
Booked a 15 minutes with you on Thursday, but if you have a few to chat, please Slack me, would love to get your thoughts on extracting an adapter module for interacting with Jitsi Meet in the "Convene" way.

Issue #39, opened 2020-07-03
Discovering Rooms


user512:
## This Issue May Be Closed When

- [x] Workspace Members may discover listed rooms
- [x] Workspace Members cannot discover unlisted rooms
- [x] The feature definition is in the repository
- [x] The step definitions are wired in
- [x] ~There is a tutorial for entering rooms somewhere under `https://www.zinc.coop/convene/`~ We can do this later.

## Feature Definition
```features/discovering-rooms.feature
Feature: In order to interact with other people in a Workspace
         I want to be able to explore the Workspace's rooms  

Scenario: Workspace Member may discover Listed Room
  Given a Workspace with a Listed Room
  When a Workspace Member is on the Workspace Dashboard
  Then they see the Room

Scenario: Workspace Member may not discover Unlisted Room
  Given a Workspace with an Unlisted Room
  When a Workspace Member is on the Workspace Dashboard
  Then they do not see the Room

Scenario: Guest may discover Listed Room
  Given a Workspace with a Listed Room
  When a Guest is on the Workspace Dashboard
  Then they see the Room

Scenario: Guest may not discover Unlisted Room
  Given a Workspace with an Unlisted Room
  When a Guest is on the Workspace Dashboard
  Then they do not see the Room
```

## UI Mockup
![Convene - Discovering Rooms - UI Sketch](https://user-images.githubusercontent.com/50284/88491447-b391c280-cf57-11ea-88b3-0d0f2f8112c9.png)

zspencer:
I've pulled out:
- Discovering Internal Rooms to https://github.com/zinc-collective/convene/issues/41
- Discovering Unlisted Rooms to https://github.com/zinc-collective/convene/issues/72

As I believe this particular feature is reaching "useful enough" to consider moving it into the "how do we present this to the world as part of our marketing and education efforts" stage of development.

@melissamcewen - Would you be interested in taking point on that? I don't think we want to tie in too much visuals since I expect the UI will evolve over the next few months; but the concept of "you can navigate between rooms to gather in a self-determining manner" may be worth highlighting on a sub-page or something? What do you think?

zspencer:
Now that we have tests in place, and notes in the CHANGELOG I am going to say this is good to go.

PR #60, opened 2020-08-02
Member only see listed rooms

user512:
Part of https://github.com/zinc-collective/convene/issues/39
Hold off from implementing `accessable_by` because we now treat every guest as members
Screen Shot 2020-08-02 at 10 32 22 AM

### DIFF
--- a/convene-web/Gemfile.lock
+++ b/convene-web/Gemfile.lock
@@ -37,8 +37,10 @@ GEM
       erubi (~> 1.4)
       rails-dom-testing (~> 2.0)
       rails-html-sanitizer (~> 1.1, >= 1.2.0)
-    active_record_union (1.3.0)
-      activerecord (>= 4.0)
+    active_record_extended (1.4.0)
+      activerecord (>= 5.0, < 6.1)
+      ar_outer_joins (~> 0.2)
+      pg (< 2.0)
     activejob (6.0.3.2)
       activesupport (= 6.0.3.2)
       globalid (>= 0.3.6)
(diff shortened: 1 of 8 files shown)


convene-web/Gemfile.lock
```diff
@@ -37,6 +37,8 @@ GEM
       erubi (~> 1.4)
       rails-dom-testing (~> 2.0)
       rails-html-sanitizer (~> 1.1, >= 1.2.0)
+    active_record_union (1.3.0)
```
zspencer, line 40: The README indicates that for folks using PostgreSQL, there is a more potent toolkit called [`ActiveRecordExtended`](https://github.com/georgekaraszi/ActiveRecordExtended).

This may give us more super-powers for the future; thoughts on using that instead of the one that just adds `Union`?

convene-web/app/views/workspaces/show.html.erb
```diff
@@ -9,7 +9,7 @@
 
 <div class="p-6">
   <ul class="mt-3 grid grid-cols-1 gap-5 sm:gap-6 sm:grid-cols-2 lg:grid-cols-4">
-    <% @workspace.rooms.each do |room| %>
+    <% @workspace.rooms.accessable_by.each do |room| %>
```
zspencer, line 12: I think we still want to pass in `current_user` even if we're not using it for anything and it's a `nil` value. That way when we do start implementing users we don't have to find and update usages of accessable by.

Thoughts?

user512: Right, but I believe we don't have `current_user` yet.

convene-web/Gemfile.lock
user512: Nice, didn't know about this tool.

convene-web/app/views/workspaces/show.html.erb
zspencer: We could add a `helper_method def current_user; nil; end` to `ApplicationController` as a seam if we want.

PR #62, opened 2020-08-02
Checkin pending Discovering Rooms features spec

user512:
Connects: #39 
Checking this in while working on discovering rooms feature so we don't forget.

### DIFF
--- a/features/discovering-rooms.feature
+++ b/features/discovering-rooms.feature
@@ -1,0 +1,39 @@
+@wip
+Feature: Discovering Rooms
+  In order to interact with other people in a Workspace
+  I want to be able to explore the Workspace's rooms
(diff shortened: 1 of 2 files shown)


features/discovering-rooms.feature
```diff
@@ -0,0 +1,39 @@
+@wip
```
zspencer, line 1: I think I'd prefer we use `@wip` on scenarios for the work that is actually in flight; and maybe use `@unstarted` for the scenarios that have not yet been picked up. We can use `@unscheduled` for feature definitions that we are contemplating but have not yet decided which release they go in to.

We'll probably also want to add `and not @unstarted` to the `test` script in `package.json`

Issue #59, opened 2020-07-31
Entering Rooms


zspencer:
### This Issue May Be Closed When
- [x] Move feature definition into repo
- [x] Wire in stepsf or non-branded domain scenarios
- [x] Entering Rooms via Slug in a Branded Domain
- [x] Entering Rooms via Full URL
- [x] Entering Rooms via Room Picker on Dashboard
- [x] Entering Rooms via Room Picker while in a Room
- [x] There are Release Notes under the 1.0 describing the first version of this feature in CHANGELOG.md
- [x] There is marketing copy describing this unique value proposition on the www.zinc.coop/convene landing page

### Feature Definition

```features/entering-rooms.feature
Feature: Entering Rooms
  In order to interact with people
  I want to enter a room

Scenario: Entering Room via Slug on a Branded Domain
  Given a Workspace with a Branded Domain
  When I visit a Room's using a slug on their Branded Domain
  Then I have entered the Room 

Scenario: Entering Unlocked Room via Room full URL
  When I visit a Room's full URL
  Then I have entered the Room 

Scenario: Entering Room via Room Picker on Workspace Dashboard
  Given I am on a Workspace Dashboard
  When I tap the Room in the Room Picker
  Then I have entered the Room 

Scenario: Entering Room via Room Picker on another Room 
  Given I am in a Room
  When I tap a different Room in the Room Picker
  Then I have entered the different Room 

Scenario: Entering Unlisted Room via Room full URL
  When I visit an Unlisted Room's Full URL
  Then I have entered the Room
```

zspencer:
@user512 - I noticed that even if we know the rooms full URL, we can't access unlisted rooms. I created a scenario illustrating the behavior that should happen:

```
Scenario: Entering Unlisted Room via Room full URL
  When I visit an Unlisted Room's Full URL
  Then I have entered the Room
```

I'll work on getting the feature steps actually wired in this weekend.

user512:
Great, I originally thought it's supposed to not able to access unlisted room.

zspencer:
Yea, I wonder if what I actually want is a ` features/unlisted-rooms.feature` file that describes what interactions people can have with unlisted rooms, similar to how we have [`features/locking-rooms.feature`](https://github.com/zinc-collective/convene/blob/development/features/locking-rooms.feature) that describes what interactions are valid for locked rooms.

zspencer:
Now that we have  tests in place and copy in the changelog, I'm going to say this is good to go! wooo!

PR #64, opened 2020-08-02
Enter room if active_room query parameter is passed in

user512:
Connects: #59 

Example:
When user enter the below url, app will enter room ada
http://convene.zinc.coop/workspaces/zinc?active_room=ada

### DIFF
--- a/convene-web/app/javascript/controllers/video_room_controller.js
+++ b/convene-web/app/javascript/controllers/video_room_controller.js
@@ -12,5 +12,15 @@ export default class extends Controller {
     this.videoRoom.addEventListener('exitedRoom', () => {
       this.wrapperTarget.classList.remove('active-room');
     })
+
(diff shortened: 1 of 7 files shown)


convene-web/app/javascript/controllers/video_room_controller.js
```diff
@@ -12,5 +12,15 @@ export default class extends Controller {
...
     })
+
+    const activeRoomName = this.activeRoomName();
+    if (activeRoomName) {
+      this.videoRoom.enterRoom(activeRoomName)
+    }
+  }
+
+  activeRoomName() {
+    const urlParams = new URLSearchParams(window.location.search);
```
user512, line 23: Doesn't work with IE, need a polyfill?
https://caniuse.com/#feat=urlsearchparams

Issue #61, opened 2020-08-02
Acquisition: Demoing Convene


zspencer:
```features/demoing-convene.feature
Feature: Demoing Convene
  In order to determine if Convene is a product I would like to use
  I want to walk through it's value proposition for my particular context


Scenario: Prospective Client Schedules a Demo
  Given I am on the Convene landing page
  When I schedule a demo
  Then an email is sent to my email address with the demo information and a calendar invite 
  And an email is sent to convene-customer-team@zinc.coop with the demo information and a calendar invite


Scenario: Customer Team Member May Perform a Demo
  Given a fresh Convene deployment
  When I visit the fresh deployment
  Then there is a Convene-Demo Workspace
  And the Convene-Demo Workspace has a Room named "Vivek's Desk"
  And the Convene-Demo Workspace has a Room named "Zee's Desk"
  And the Convene-Demo Workspace has a Room named "Water Cooler"
  And the Convene-Demo Workspace has a Room named "The Ada Lovelace Room"
  And I can access the `convene-demo` workspace directly via `https://convene-demo.{{ deploymentDomain }}`
```

### TODO
- [x] Put the feature tests in the repo
- [x] Create a `Convene Demo` workspace
- [x] Smoke test! Ensure convene-demo.zinc.coop lands us in the `convene-demo` workspace on production
- [ ] Wire in the step definitions so the feature tests work
- [ ] Integrate the Convene Demo calendly into the Convene landing page somehow

PR #65, opened 2020-08-02
Sprout a Convene Demo Workspace

zspencer:
See https://github.com/zinc-collective/convene/issues/61

### DIFF
--- a/convene-web/app/models/demo_workspace.rb
+++ b/convene-web/app/models/demo_workspace.rb
@@ -1,0 +1,47 @@
+class DemoWorkspace
+  # Creates the demo workspace, but only on environments which are
+  # configured to include the demo workspace.
+  def self.prepare
+    return unless Feature.enabled?(:demo)
+
+    Factory.new(Client).find_or_create_workspace!
+  end
+
+  # Assembles the DemoWorkspace using the provided Client and Workspace
+  # classes.
+  class Factory
+    attr_accessor :client_repository
+
+    # These are the Rooms we expect the DemoWorkspace to have by default.
+    # Our customer team leverages them for their demonstrations of Convene's
+    # feature set.
+    DEMO_ROOMS = ["Zee's Desk", "Vivek's Desk", 'Water Cooler', 'The Ada Lovelace Room'].freeze
+
+    # @param [ActiveRecord::Relation<Client>] client_repository Where to ensure there
+    #  is a Zinc Client with the Convene Demo workspace
+    def initialize(client_repository)
+      self.client_repository = client_repository
+    end
+
+    # Creates the Convene Demo Workspace and Zinc Client if necessary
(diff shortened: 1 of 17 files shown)


convene-web/app/models/feature.rb
```diff
@@ -0,0 +1,10 @@
+# An internal Microlibrary for feature flagging to help avoid
+# sprinkling `ENV` calls throughout the codebase.
+class Feature
+  # @param [Symbol, #to_sym] feature The name of the feature we are checking
+  # @return [TrueClass, FalseClass]
+  def self.enabled?(feature)
+    feature = feature.to_sym if feature.respond_to?(:to_sym)
+    return true if feature == :demo && ENV['DEMO_ENABLED']
```
user512, line 8: I think this never returns `false`?

zspencer: Good point! https://github.com/zinc-collective/convene/pull/66

convene-web/app/models/demo_workspace.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    # These are the Rooms we expect the DemoWorkspace to have by default.
+    # Our customer team leverages them for their demonstrations of Convene's
+    # feature set.
+    DEMO_ROOMS = ["Zee's Desk", "Vivek's Desk", 'Water Cooler', 'The Ada Lovelace Room'].freeze
+
+    # @param [ActiveRecord::Relation<Client>] client_repository Where to ensure there
+    #  is a Zinc Client with the Convene Demo workspace
+    def initialize(client_repository)
+      self.client_repository = client_repository
```
user512, line 23: I missed this during review, I don't see any code referencing `client_repository`, did I miss anything?

zspencer: Woops, it should be in 44.

user512: Got it, makes sense.

user512: Erm, I am thinking why do we want to leave this interface open since code in line 44 needs to be `Client` anyway?

zspencer: So, I have a habit (which may or may not be useful) of not referencing model classes whenever I can help it. Instead, I treat the Model class like a [Repository](https://8thlight.com/blog/mike-ebert/2013/03/23/the-repository-pattern.html). 

It's not particularly "necessary" by any means, but I have grown to appreciate explicit seams like this.  You can see the complete implementation here: https://github.com/zinc-collective/convene/pull/67/files#diff-9740bfa8740b2c2a549405d21a90605a

zspencer: Moved to it's own PR so that 67 can stay focused - https://github.com/zinc-collective/convene/pull/68

PR #67, opened 2020-08-03
Workspaces may be entered via Branded Domain

zspencer:
See: https://github.com/zinc-collective/convene/issues/61

Branded Domains allow Visitors to bypass hitting the `/workspaces/:slug`
path and drop them right in the Workspaces Lobby.

For example, `convene-demo.zinc.coop` is a branded domain for the

`Convene Demo` workspace. `meet.zinc.coop` is a branded domain for the
Zinc workspace.

Screen Shot 2020-08-02 at 4 48 46 PM


convene-web/db/seeds.rb
```diff
@@ -6,24 +6,26 @@
...
+                        jitsi_meet_domain: 'meet.zinc.coop')
 
 zincs_workspace.members << zee
 zincs_workspace.members << tom
 zincs_workspace.members << bene
 
-
-ada = zincs_workspace.rooms.create(name: 'Ada', access_level: :unlocked, publicity_level: :listed)
-ttz = zincs_workspace.rooms.create(name: 'Talk to Zee', access_level: :unlocked, publicity_level: :unlisted)
+ada = zincs_workspace.rooms.find_or_create_by!(name: 'Ada'
```
user512, line 25:
```suggestion
ada = zincs_workspace.rooms.find_or_create_by!(name: 'Ada')
```

convene-web/app/models/workspace.rb
```diff
@@ -4,6 +4,9 @@ class Workspace < ApplicationRecord
   # Which client owns the workspace
   belongs_to :client
 
+  # The fully-qualified domain to enter the workspace
+  attribute :branded_domain, :string
```
user512, line 8: Do we need to add some unique constraint/ validation on `branded_domain`?

zspencer: Good point! I've added a uniquness validation.

PR #70, opened 2020-08-03
Automate creating DNS entries for Convene-Web on Heroku

zspencer:
I noticed that we had set some cloudflare records for convene that were not being managed.

I'm not 100% sure that this is _quite_ how I want to do this; for instance it's not a _client_ that we are setting the DNS for right now; but I'm not sure it's too important.


zspencer:
For example, it may be worth digging into using this: https://registry.terraform.io/providers/heroku/heroku/latest/docs/resources/domain

So that when we add branded domains on a client-by-client basis it will configure them in Heroku for us.

Issue #23, opened 2020-05-29
Spaces: Customization


zspencer:
```features/branded-spaces.feature
Feature: Branded Spaces
  In order to have a place that is truly mine
  As a Space Owner
  I want to be able to brand my space with communicative and cosmetic details
```

### TODO
- [ ] Design
  - [ ] Write User Journey for Branded Spaces
  - [ ] Mockups for where to Configure Branded Spaces
- [ ] Space Names
  - [ ] Space Owner may set their Space Name
- [ ] https://github.com/zinc-collective/convene/issues/74
- [ ] Themes

### Done
- [x] Remove Jitsi branding
  -  Google Slides laying out where in the UI these exist https://docs.google.com/presentation/d/15KlJoJRFng3xadSw1o0E0xSunGWpECSEcdBRiZBgDBA/edit#slide=id.g860bdd4a39_0_3
  - Logo: https://drive.google.com/drive/folders/1Rmj17dQweGt5l9EFFBrp_9DHIY1WHmG2 
  - [x] In Rooms
    - [x] People are called "Participants" instead of "Fellow Jitsers"
    - [x] The  `jitsi.org` + logo watermark is removed
     - [x] ~~And replaced with the [Convene Logo](https://drive.google.com/drive/folders/1Rmj17dQweGt5l9EFFBrp_9DHIY1WHmG2)~~
    - [x] The `Jitsi` name is replaced with `Convene` when requesting permissions for video and audio
      - [x] in the Browser prompt (This is correct on convene.zinc.coop, but not on meet.zinc.coop)
      - [x] In the web page "Jitsi Meet Needs to..."
  - [x] ~~When Links are "unfurled" into Facebook/Slack/Etc it says "Convene" and not "Jitsi"~~

maximegalon5:
I've created a [G Slide](https://docs.google.com/presentation/d/15KlJoJRFng3xadSw1o0E0xSunGWpECSEcdBRiZBgDBA/edit?usp=sharing) where I've pointed to some considerations.

They revolve around:

- replacing with the Convene Logo <- which means we have to decide and pick something

- Consider changing the "meet.zinc.coop" URL if we believe the branding is inconsistent with Convene (would love to know what @colombene @cricketlova think about this).

- Consider changing the text on the landing page. I've made suggested changes in the Notes section of the slides, but defer to @colombene for a final go ahead.

cricketlova:
[...]

PR #73, opened 2020-08-03
Copyable link to enter specific rooms

user512:
Part of https://github.com/zinc-collective/convene/issues/39 & https://github.com/zinc-collective/convene/issues/59:
This pull request intended to allow copyable link to enter specific rooms by initialize Jitsi connection when user go to `/workspaces/workspace-slug/rooms/room-slug`.
Since branded-domain link looks nicer, all rooms/ workspaces links will use custom domain and use conventional RESTful path as a fallback.

Since we migrated to server side path, previous JS solution is cleaned up.

Part of https://github.com/zinc-collective/convene/issues/23, remove `Fellow Jitser` display name.

TODO:
- [x] Switching rooms connect to another Jitsi rooms
- [x] Gracefully handle hang up call
- [x] Look into https://github.com/zinc-collective/convene/pull/73#discussion_r464606188
- [x] Clean up video room js.
Screen Shot 2020-08-03 at 2 57 55 PM
Screen Shot 2020-08-03 at 3 52 10 PM


convene-web/app/controllers/rooms_controller.rb
```diff
@@ -1,5 +1,9 @@
...
   def show
-    @room = current_workspace.rooms.friendly.find(params[:id])
+    current_room
   end
-end
\ No newline at end of file
+
+  private def current_room
+    @current_room ||= current_workspace.rooms.friendly.accessable_by(current_person).find(params[:id])
+  end
```
zspencer, line 8: I'm not sure what `current_room` means in this case; but it looks kind of like we're starting to define `loader` methods; which makes me want to reach for a tool like [decent_exposure](https://github.com/hashrocket/decent_exposure)

convene-web/config/routes.rb
```diff
@@ -5,6 +5,6 @@
   resources :workspaces
   constraints BrandedDomain.new(Workspace) do
     root 'workspaces#show'
-    get "/:id", to: 'rooms#show'
+    get "/:id", to: 'rooms#show', as: :room
```
zspencer, line 8: I think that we want room names on branded domains to purely be an entrance point, so we don't want to overload the `room_path` and instead we want to continue to use `workspace_room_path(workspace.slug, room.slug)` when creating links.

convene-web/app/javascript/src/video_room.js
```diff
@@ -21,7 +21,6 @@ export default class VideoRoom {
   }
 
   exitRoom() {
-    this.jitsi.dispose();
```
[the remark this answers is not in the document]
user512: No, going to clean up unnecessary js on next commit.

convene-web/config/routes.rb
user512: Just realized we want to keep the `host/workspaces/:workspaces_slug/rooms/:room_slug`, now I understand this, will make change.

convene-web/app/controllers/rooms_controller.rb
user512: I was trying to expose the room show in `rooms/show.html.erb`.
Could be I haven't understand the benefit of [decent_exposure](https://github.com/hashrocket/decent_exposure) but don't feel like pull in a library to expose a room just yet.

convene-web/app/javascript/src/video_room.js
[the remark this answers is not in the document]
user512: I believe what `dispose()` does is to remove all listeners and remove the iframe.
https://github.com/jitsi/jitsi-meet/blob/342a00a6af323d98eef33fc60f4e2a515c8e254d/modules/API/external/external_api.js#L521-L527
Since we are reloading the page anyway, so I decided to remove all logic using `dispose()`.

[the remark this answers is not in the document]
zspencer: I am slightly worried it leaves the server thinking the connection is not closed; but we can validate that by seeing that when user A leaves by clicking a button, user B sees them leave immediately.

convene-web/app/controllers/rooms_controller.rb
zspencer: Yea, the value in decent_exposure is it creates a standard mechanism for exposing the data that a controller wants to expose in a way that is consistent across all controllers. I think it’s something to look into as we start wiring in pundit; or if we start defining helper methods to expose data to the UI instead of a single instance variable.

convene-web/app/helpers/rooms_helper.rb
```diff
@@ -0,0 +1,17 @@
+module RoomsHelper
+  def end_call_path(workspace)
+    if workspace.branded_domain.present?
+      "https://#{workspace.branded_domain}/"
+    else
+      workspace_path(workspace)
+    end
+  end
```
zspencer, line 8: These custom paths methods indicate to me that we’re “going off the Rails.” I’d rather we rely on `workspace_path` and don’t create custom path methods unless necessary

user512: I feel the same way but also saw `doorkeeper` was doing something similar so I thought it's ok.
https://github.com/doorkeeper-gem/doorkeeper/blob/2fd1556ade7572d01e507c0fbc53eaf6c5bbfd19/app/helpers/doorkeeper/dashboard_helper.rb#L17-L19

convene-web/app/javascript/src/video_room.js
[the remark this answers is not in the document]
user512: Good concern, I have confirmed once user A leaves, user B sees them leave.
But I wasn't sure how the server was able to detect that.

convene-web/app/helpers/rooms_helper.rb
user512: One thing I don't like about this `branded_domain` route is it's making smoke test locally more difficult, I am more concern on how we write e2e test on this.

zspencer: Yea, that's why I'm trying to make sure it's something that _only_ exists as an entrance point into the system; and _not_ as something that the system tries to maintain consistency on. Essentially it's something you put on the business card; but it's not something we project back into the world.

PR #75, opened 2020-08-04
Fix bug where invalid room name in pass onto Jitsi

user512:
Fix this:
Screen Shot 2020-08-03 at 8 17 21 PM
Because the JS was try to GET `https://meet.zinc.coop/Vivek's%20Desk` and it errors out.


user512:
Since this is a small change and will avoid error on demo, going to merge.

zspencer:
Thank you!

PR #78, opened 2020-08-11
Document how Entering Rooms works

zspencer:
See: https://github.com/zinc-collective/convene/issues/59

I've added the `@built` tag to indicate which features have been
_built_, so that it's a bit easier for someone who wants to start doing
some test automation to figure out what feature files need more work.

### DIFF
--- a/features/entering-rooms.feature
+++ b/features/entering-rooms.feature
@@ -1,0 +1,34 @@
+Feature: Entering Rooms
+  In order to interact with other People
+  I want to enter a Room
+
+  Rooms are where People gather, and can be entered in a number of ways.
+
+  Most frequently, we expect people to enter a Room from their Workspace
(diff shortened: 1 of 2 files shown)


features/entering-rooms.feature
```diff
@@ -0,0 +1,34 @@
...
+  @built @unimplemented-steps
+  Scenario: Entering Room via Slug on a Branded Domain
+    Given a Workspace with a Branded Domain
+    When I visit a Room's using a slug on their Branded Domain
+    Then I am in the Room
+
+  @built @unimplemented-steps
+  Scenario: Entering Room via Room full URL
+    When I visit a Room's full URL
+    Then I am in the Room
```
user512, line 34: Do you think we need to test both visiting unlisted room and listed room via Room full URL?

PR #80, opened 2020-08-11
Move end call button back to jitsi iframe

user512:
screencast 2020-08-11 11-13-21

### DIFF
--- a/convene-web/app/models/workspace.rb
+++ b/convene-web/app/models/workspace.rb
@@ -6,7 +6,7 @@ class Workspace < ApplicationRecord
 
   # The fully-qualified domain to enter the workspace
   attribute :branded_domain, :string
-  validates :branded_domain, uniqueness: true
+  validates :branded_domain, uniqueness: true, allow_nil: true
 
   # The domain we expect jitsi meet to be running on
   attribute :jitsi_meet_domain, :string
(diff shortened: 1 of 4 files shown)


convene-web/app/models/workspace.rb
```diff
@@ -6,7 +6,7 @@ class Workspace < ApplicationRecord
 
   # The fully-qualified domain to enter the workspace
   attribute :branded_domain, :string
-  validates :branded_domain, uniqueness: true
+  validates :branded_domain, uniqueness: true, allow_nil: true
```
zspencer, line 9: Oh. That makes sense! Thank you!

Something I'm trying to do in my work in Zinc is to attempt to communicate intent in ~4 places:

1. The code itself 
2. The documentation describing the lines of code 
3. The tests describing the developer-facing workings of the code
4. The feature tests describing the human-facing value proposition

This is significantly more work, so feel free to ignore me; but here's some examples of how I'd do that in this case:

1. You did this already with `allow_nil: true` 
2. I would possibly update line 7 from?
```
# The fully-qualified domain to enter the workspace
```
to
```
# The fully-qualified domain to enter the workspace. Workspaces without a branded_domain are still 
# accessible via their slug. The branded_domain must be unique to ensure we don't accidentally place
# a visitor into the wrong workspace.
```
3. I would add a rspec test for allowing nil for this field in `app/models/workspace_spec.rb`
4. I would probably _not_ add a feature test for this.

PR #83, opened 2020-08-12
Fix switch room doesn't work

user512:
Before this pull request, switching room doesn't work smoothly, when users click another room when in different room, the http request to a new room got cancelled. 

I am not 100% sure why disabling turbolink doesn't work but then I know moving the jitsi external_api.js to head and re-enabling turbolink works.

I think it's a sign I need to start implementing system test.


zspencer:
Another option may be to look into using Heroku Pipelines and Review Apps instead of using Github Actions to deploy. This would allow us to smoke-test before merging.

I think we could even point the system tests _at_ the Heroku Review app if we wanted.

PR #90, opened 2020-08-14
System test `Entering Room via Room Picker from Workspace Dashboard`

user512:
Would love to get some early feedback out the first system test to see if this is the correct direction.

### DIFF
--- a/convene-web/.env.example
+++ b/convene-web/.env.example
@@ -1,0 +1,2 @@
+DEMO_ENABLED=false
+SYSTEM_TEST=false
--- a/convene-web/.env.example
+++ b/convene-web/.env.example
@@ -1,2 +1,2 @@
 DEMO_ENABLED=false
-SYSTEM_TEST=false
+SYSTEM_TEST=true
--- a/convene-web/app/models/system_test_workspace.rb
+++ b/convene-web/app/models/system_test_workspace.rb
@@ -38,7 +38,7 @@ class SystemTestWorkspace
     def find_or_create_workspace!
       workspace = client.workspaces.find_or_create_by!(name: 'System Test Branded Domain')
       workspace.update!(jitsi_meet_domain: 'convene-videobridge-zinc.zinc.coop',
-                        branded_domain: "#{ENV.fetch('HEROKU_APP_NAME', 'convene-pr-xxx')}.herokuapp.com",
+                        branded_domain: heroku_app_convene_domain,
                         access_level: :unlocked)
       add_demo_rooms(workspace)
 
@@ -58,6 +58,10 @@ class SystemTestWorkspace
 
     private def client
       @_client ||= client_repository.find_or_create_by!(name: 'Zinc')
+    end
+
+    private def heroku_app_convene_domain
+      "#{ENV.fetch('HEROKU_APP_NAME', 'convene-pr-xxx')}.herokuapp.com"
     end
   end
 end
--- a/convene-web/app/models/system_test_workspace.rb
+++ b/convene-web/app/models/system_test_workspace.rb
@@ -36,12 +36,17 @@ class SystemTestWorkspace
     end
 
     def find_or_create_workspace!
-      workspace = client.workspaces.find_or_create_by!(name: 'System Test')
+      workspace = client.workspaces.find_or_create_by!(name: 'System Test Branded Domain')
       workspace.update!(jitsi_meet_domain: 'convene-videobridge-zinc.zinc.coop',
-                        branded_domain: "#{ENV.fetch('HEROKU_APP_NAME')}.herokuapp.com",
+                        branded_domain: "#{ENV.fetch('HEROKU_APP_NAME', 'convene-pr-xxx')}.herokuapp.com",
                         access_level: :unlocked)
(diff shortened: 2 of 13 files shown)


convene-web/.env.example
```diff
@@ -0,0 +1,2 @@
+DEMO_ENABLED=false
+SYSTEM_TEST=false
```
zspencer, line 2: Would we want this to be true on local dev environments so that the system tests can run against the local dev environment?

convene-web/app/models/system_test_workspace.rb
```diff
@@ -36,12 +36,17 @@ def initialize(client_repository)
     end
 
     def find_or_create_workspace!
+      workspace = client.workspaces.find_or_create_by!(name: 'System Test Branded Domain')
+      workspace.update!(jitsi_meet_domain: 'convene-videobridge-zinc.zinc.coop',
+                        branded_domain: "#{ENV.fetch('HEROKU_APP_NAME', 'convene-pr-xxx')}.herokuapp.com",
```
zspencer, line 41: the `default_branded_domain` logic is getting complicated enough that we may want to pull it into it’s own function

features/entering-rooms.feature
```diff
@@ -7,20 +7,20 @@ Feature: Entering Rooms
...
   @built @unimplemented-steps
   Scenario: Entering Room via Room Picker from another Room
     Given I am in a Room
     When I tap a different Room in the Room Picker
     Then I am in the Room
 
-  However, we also want to support Reople entering a Room via
-  the Room's full URI or from a short URI when a Space has a Branded Domain.
+  # However, we also want to support Reople entering a Room via
+  # the Room's full URI or from a short URI when a Space has a Branded Domain.
```
zspencer, line 23: Does cucumber-js barf on this? :(. I was really hoping that we could Interleave markdown with the scenarios to make the feature level documentation more communicative.

convene-web/app/views/workspaces/_room_card.html.erb
```diff
@@ -1,4 +1,4 @@
-<li class="col-span-1 bg-white rounded-lg shadow">
+<li class="col-span-1 bg-white rounded-lg shadow" name="<%= room.name %>" >
```
zspencer, line 1: I would recommend using [`id`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) over [`name`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Attribute_list) as the `name` attribute is intended to communicate form field attributes.

features/steps/room_steps.js
```diff
@@ -15,9 +18,16 @@ When("a {actor} provides the correct {room} Key", function (actor, room) {
   return "pending";
 });
 
-Then("the {actor} is placed in the {room}", function (actor, room) {
-  // Write code here that turns the phrase above into concrete actions
-  return "pending";
+When('the {actor} taps the {room} in the Room Picker', async function (actor, room) {
+  const workspacePage = new WorkspacePage(this.driver)
+  const roomCard = workspacePage.findRoomCard(room);
+  await workspacePage.enterRoom(roomCard);
```
zspencer, line 24: From a design perspective, I would probably remove a couple of the function calls here, maybe going with something like: 

```
When(‘the {actor} taps the {room} in the Room Picker’, async function (actor, roomName) {  
  this.workspace.enterRoom({ roomName })
})
```

This would require us to set `this.workspace` earlier on, perhaps in the `Given` .

features/steps/room_steps.js
```diff
@@ -15,9 +18,16 @@ When("a {actor} provides the correct {room} Key", function (actor, room) {
...
+});
+
+Then("the {actor} is placed in the {room}", async function (actor, room) {
+  await this.driver.wait(until.elementLocated(By.css("[name='jitsiConferenceFrame0']")));
+  const jitsiIframe = await this.driver.findElements(By.css("[name='jitsiConferenceFrame0']"))
+  assert.equal(jitsiIframe.length, 1, 'Jitsi iframe not found');
```
zspencer, line 30: This feels like something belongs in the WorkspacePage object, i.e. `assert.equal(await this.workspace.videoPanel().isPresent())` to keep calls to the driver directly out of step definitions.

features/steps/workspace_steps.js
```diff
@@ -1,4 +1,9 @@
 const { Given, When, Then } = require("cucumber");
+const WorkspacePage = require("../page-objects/WorkspacePage");
+
+Given('the {actor} is on the {workspace} Dashboard', async function (actor, workspace) {
+  await new WorkspacePage(this.driver).enter(workspace);
```
zspencer, line 5: Here is where I would do `this.workspace = ...` to provide the workspace to steps later on.

features/entering-rooms.feature
user512: Yes, unfortunately that's the case and I haven't been able to find a way to solve this.
I think moving it up closed to the `Feature:` block would make sense.

zspencer: That makes sense. I’m a bit bummed but I’ll deal :).

PR #92, opened 2020-08-17
test entering room features

user512:


convene-web/app/models/room.rb
```diff
@@ -12,7 +12,7 @@ class Room < ApplicationRecord
 
   # FriendlyId's does the legwork to make the slug uri-friendly
   extend FriendlyId
-  friendly_id :name, use: :slugged
+  friendly_id :name, use: :scoped, scope: :workspace
```
user512, line 15: Ref: http://norman.github.io/friendly_id/file.Guide.html#Unique_Slugs_by_Scope

features/page-objects/WorkspacePage.js
```diff
@@ -10,18 +10,23 @@ class WorkspacePage extends Page {
...
+  async findRoomCard(room) {
+    await this.driver.wait(until.elementLocated(By.id(room.name)));
+    return await this.driver.findElement(By.id(room.name));
   }
 
   async videoPanel() {
-    await this.driver.wait(until.elementLocated(By.css("[name='jitsiConferenceFrame0']")));
-    return await this.driver.findElement(By.css("[name='jitsiConferenceFrame0']"));
+    await this.driver.wait(until.elementLocated(By.css("[name*='jitsiConferenceFrame']")));
+    return await this.driver.findElement(By.css("[name*='jitsiConferenceFrame']"));
```
user512, line 29: Relax the selector a bit because when entering a second room, we will have jitsiConferenceFrame1 instead.

convene-web/app/models/room.rb
zspencer: Nice call. Perhaps adding a comment near the implementation or test linking to the docs would provide a breadcrumb for folks to follow in the future without having to immediately go through search.

convene-web/app/helpers/rooms_helper.rb
```diff
@@ -14,4 +14,8 @@ def video_room_path(workspace, room)
       workspace_room_path(room.workspace, room)
     end
   end
+
+  def video_room_name(workspace, room)
+    "#{workspace.slug}--#{room.slug}"
+  end
```
zspencer, line 20: Could we put this on the `Room` class, perhaps with a method name like `cross_system_unique_identifer` or something to indicate that it's used to figure out how to reference the particular room across Jitsi/DOM/etc? It's almost like a URI but not quite...

Also, are the two dashes semantically meaningful? My understanding of the direction that semantic HTML/CSS are trending towards is that: ` ` or `_` seperates ideas, `-` seperates words, `--` gives you a state.

features/entering-rooms.feature
```diff
@@ -16,19 +16,19 @@ Feature: Entering Rooms
     When the Workspace Member taps the "Listed Room 1" Room in the Room Picker
     Then the Workspace Member is placed in the "Listed Room 1" Room
 
-  @built @unimplemented-steps
+  @built
   Scenario: Entering Room via Room Picker from another Room
-    Given I am in a Room
-    When I tap a different Room in the Room Picker
-    Then I am in the Room
+    Given the Workspace Member is in the "System Test" Workspace and in the "Listed Room 1" Room
```
zspencer, line 21: Good call on moving this to the Workspace Member instead of I to make it more clear who the actor is.

features/page-objects/WorkspacePage.js
zspencer: Hmm, that's concerning. Does that mean we're inserting a new iFrame every time we switch rooms? Is there a possibility for leaking Video/Audio to a room unintentionally then?

convene-web/app/helpers/rooms_helper.rb
zspencer: Apparently this is called "block element modifier" http://getbem.com/

user512: Yes and great name suggestion.

features/page-objects/WorkspacePage.js
user512: Legit concern, will fix in the next pr.

PR #96, opened 2020-08-20
Fix a bug where hitting brower back button would show two jitsi iframe

user512:
Before fix:
screencast 2020-08-20 11-24-30


user512:
No, that's intended because I set the heroku app url as the custom domain but I will move them into its own subdomain of zinc.coop soon in https://github.com/zinc-collective/convene/pull/94.

PR #105, opened 2020-08-23
Setup test on GH action with the ability to test branded domain

user512:
* Run RSpec and e2e test on Github actions.
* Run rails server on port 80 to test branded domain.
If we run e2e test when server is running on port 80 and edit `/etc/host` with 127.0.0.1 system-test.zinc.local,
then the driver can simply navigate to `http://system-test.zinc.local`, and user can click enter room and reach `http://system-test.zinc.local/room-slug`.

But, I wasn't sure how to make branded domain pass locally.


.github/workflows/test-convene-web.yml
```diff
@@ -0,0 +1,89 @@
...
+
+    - name: Set up Ruby 2.7
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 2.7
+
+    - name: Increase the amount of inotify watchers
+      run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
+    - name: Install yarn & node packages
```
zspencer, line 34: Could we run `bin/setup` and install Yarn, bundle, database, etc. all at once?

convene-web/spec/models/demo_workspace_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 RSpec.describe DemoWorkspace, type: :model do
   describe ".prepare" do
+    before { allow(ENV).to receive(:[]).with('DEMO_ENABLED').and_return('true') }
```
zspencer, line 5: Oh nice catch. Maybe we want this in it's own context for `When the environment is set to enable demo` and then there could be the inverse with an it.

user512:
I was exploring to run the server at port 80 to test branded domain. But still facing some problem.
I think I’m very close to getting it working.

user512:
@zspencer, I think configuring our CI to spin up server on port 80 would be an easy path to system test branded domain. But I wonder if you have other thoughts. 
I might switch gear back to development instead if I can't get branded domain testing working in a reasonable amount of time.

PR #111, opened 2020-09-06
Begin to wire in feature test for Locking Rooms

zspencer:
Paired with @cjoulain to flesh out the Feature definitions for Locking Rooms.

First, we noticed that there were a lot of scenarios that maybe didn't need to exist; and we collapsed them into a single scenario that lays out which people have which access rights.

Then, we noticed that there were some missing scenarios, like locking and unlocking the room. So we put those in as well.

Finally, we began to wire those scenarios into parameterTypes and step definitions so we can write user-level integration tests.

(P.s, I'm going to let @cjoulain merge after approval, so they get the `Co-Authored By` credit)

### DIFF
--- a/features/locking-rooms.feature
+++ b/features/locking-rooms.feature
@@ -22,14 +22,14 @@ Feature: Locking Rooms
   @wip
   Scenario: Entering a Locked Room
     Given a Workspace with a Locked Room
-    Then a Workspace Member may enter the room after providing the correct Room Key
-    And a Workspace Member may not enter the room after providing the wrong Room Key
-    And a Guest may enter the room after providing the correct Room Key
-    And a Guest may not enter the room after providing the wrong Room Key
+    Then a Workspace Member may enter the Room after providing the correct Room Key
+    And a Workspace Member may not enter the Room after providing the wrong Room Key
+    And a Guest may enter the Room after providing the correct Room Key
+    And a Guest may not enter the Room after providing the wrong Room Key
     # We're not sure if this is actually a good idea, we should check with
     # design / product to make sure that it makes sense for admins to be able
     # to barge in like a grumpy parent on prom night.
-    And a Workspace Admin may enter the room without providing a Room Key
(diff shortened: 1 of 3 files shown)


features/locking-rooms.feature
```diff
@@ -18,32 +17,52 @@ Feature: Locking Rooms
...
-    When a Workspace Admin provides the correct Room Key
-    Then the Workspace Admin is placed in the Room
+    Then a Workspace Member may enter the Room after providing the correct Room Key
+    And a Workspace Member may not enter the Room after providing the wrong Room Key
+    And a Guest may enter the Room after providing the correct Room Key
+    And a Guest may not enter the Room after providing the wrong Room Key
+    # We're not sure if this is actually a good idea, we should check with
+    # design / product to make sure that it makes sense for admins to be able
+    # to barge in like a grumpy parent on prom night.
+    And a Workspace Admin may enter the Room without providing a Room Key
```
zspencer, line 32: @maximegalon5 / @colombene  - We were thinking we weren’t sure if a locked room should block out admins or not. What do you think?

features/parameter-types/rooms.js
```diff
@@ -20,7 +33,7 @@ class Room {
 // See: https://github.com/zinc-collective/convene/issues/41
 defineParameterType({
   name: "accessLevel",
-  regexp: /(Unlocked|Internal|Locked)/,
+  regexp: /(an |a )?(Unlocked|Internal|Locked)/,
```
zspencer, line 36: We will probably want to update the arguments we’re passing to the transformer here, since it seems like there will be _two_ arguments in cases where an article adjective is provided, and _one_ argument in cases where only `(Unlocked|Internal|Locked)` are provided.

We may want to think a bit more about how to use parameter types so that we don’t run into this rough edge in the future (for instance, we may want to _require_ the article adjectives, so that we don’t have some cases where a single argument is passed into the classes inflated by the transformer.)

user512 (approved):

PR #120, opened 2020-09-11
Revert locking rooms features definitions

user512:
* We want to keep the features defintions untouched, it was intentionally grouped this way.

### DIFF
--- a/features/locking-rooms.feature
+++ b/features/locking-rooms.feature
@@ -19,25 +19,17 @@ Feature: Locking Rooms
 
   # Wireframe:
   # https://xd.adobe.com/view/fd425dbe-5384-44c9-997a-eeee6e886a86-a811/screen/04ee266e-931b-4bde-bcf9-af94c7ac444e
-  @built @unimplemented-steps
-  Scenario: Workspace Member entering a Locked Room
-    Given the Workspace Member is on the "System Test" Workspace Dashboard
-    Then a Workspace Member may enter the "Listed Locked Room 1" Room after providing the correct Room Key
-    And a Workspace Member may not enter the "Listed Locked Room 1" Room after providing the wrong Room Key
-
   @wip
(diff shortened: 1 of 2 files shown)


features/locking-rooms.feature
```diff
@@ -19,25 +19,17 @@ Feature: Locking Rooms
...
-
-  @wip
-  Scenario: Guest entering a Locked Room
-    Given the Guest is on the "System Test" Workspace Dashboard
-    And a Guest may enter the "Listed Locked Room 1" after providing the correct Room Key
-    And a Guest may not enter the "Listed Locked Room 1" after providing the wrong Room Key
-
   @wip
-  Scenario: Workspace Admin entering a Locked Room
+  Scenario: Entering a Locked Room
```
zspencer, line 23: Is this now `built`? Or do we still have unimplemented steps or features?

zspencer (changes requested):
Want to make sure we’re running it if it’s actually been built.

features/locking-rooms.feature
user512: I think this is not `built` because we haven't implement `Workspace Admin`.

If this test doesn't include `Workspace Admin` behavior, then I think it can be considered built because `Guest` and `Workspace Member` have same behavior where we can still write test without implementing those two roles.

zspencer: Ah, excellent. So @maximegalon5 and @colombene and I talked through this a bit here: https://soloish.slack.com/archives/C013U0E797Y/p1599436256003600

Right now, we think that Space Admins / Owners _should not_ be able to enter the Room without the key for now. However, we may want _Operators_ to be able to enter rooms without a key. I think it’s probably safe to ditch the comment and the “And a Worksapace Admin...” rules completely for now; and when we start making Space Owners a “thing” we can add them back in.

Issue #115, opened 2020-09-08
Refreshing the page drops you in the hallway


zspencer:
Noticed when I refreshed the page it took me back to the Space hallway; would rather it stayed in the current page.

zspencer:
I can confirm this has been fixed! Thank you @user512!

PR #121, opened 2020-09-11
Stay in video room when user refresh

user512:
Connects #115 
* Previously the `videoConferenceLeft` event will get triggered when user reload.
Use `readyToClose` event instead because it only got triggered when user intended to hangup.

### DIFF
--- a/convene-web/app/javascript/src/video_room.js
+++ b/convene-web/app/javascript/src/video_room.js
@@ -32,7 +32,9 @@ export default class VideoRoom {
     if (!this.roomName) return;
     this.jitsi = new JitsiMeetExternalAPI(this.domain, this.jitsiApiOption());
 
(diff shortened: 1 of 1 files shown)


zspencer:
Hmm, I think what I would prefer is that when the meet closes, it doesn’t actually send the user to a different location. I’d rather it terminated the Jitsi connection and removed the video panel from the UI.

This is because as we build out additional pieces to a room (for instance, if we put in chat, or a ‘message of the day’ or what not) we can’t really depend on “closing video” to be the same as “leaving the room.”

Thoughts?

user512:
I think what you said makes sense if we currently have some other features other than video room.
But for now, we only have video room when user enter room so it makes sense to "redirect" user back to workspace once user hang up/ end call, or else, by just collapsing the video room div and not changing the url to `/workspace/workspace_slug`, it puts the user in a weird state where the url tells the users the users is in the room but the UI tells otherwise.

PR #126, opened 2020-09-20
Wire in steps for entering a locked room

user512:
* [x] Fix AccessLevels getting undefined values from the accessLevel parameterType
* [X] Implement step definition for entering a workspace with a Locked Room
* [X] Implement step definition for entering a Locked Room
* [X] Implement step definition for being prevented from entering a Locked Room
* [X] Expose the Room's Access Level in the the room card's DOM
* [X] Add helper method to Workspace PageObject for selecting RoomCards from the DOM
* [x] Verify that an error message is provided when entering the wrong room and that people don't enter the room
* [ ] (Maybe?) Clarify parameter type and regex in README (@KellyAH, @zspencer)
* [ ] (Maybe?) Align SystemTest Workspace locked room access code with Demo Workspace

### DIFF
--- a/convene-web/app/models/room.rb
+++ b/convene-web/app/models/room.rb
@@ -19,6 +19,7 @@ class Room < ApplicationRecord
   # `locked` indicates that only participants who know the rrooms `access_code` may access the room.
   # `internal` indicates that only participants who are Workspace Members _or_ know the Workspaces `access_code` may
   # access the room.
+  # TODO Return :unlocked by default, add database non null
   attribute :access_level, :string
 
   # A room's Access Code is a "secret" that, when known, grants access to the room.
--- a/convene-web/app/views/workspaces/_room_card.html.erb
+++ b/convene-web/app/views/workspaces/_room_card.html.erb
@@ -1,4 +1,4 @@
-<li class="col-span-1 bg-white rounded-lg shadow mb-4" id="<%= room.name %>" >
+<li class="col-span-1 bg-white rounded-lg shadow mb-4 <%= "--#{room.access_level}" %>" id="<%= room.name %>">
   <div class="w-full flex items-center justify-between p-6 space-x-6">
     <div class="flex-1 truncate">
       <div class="flex items-center space-x-3">
--- a/features/page-objects/WorkspacePage.js
+++ b/features/page-objects/WorkspacePage.js
@@ -48,6 +48,11 @@ class WorkspacePage extends Page {
     assert.equal(videoPanels.length, 1, `${videoPanels.length} was found.`)
     return await this.driver.findElement(jitsiConferenceFrame);
   }
+
+  roomCardsWhere({ accessLevel }) {
+    const selector = `.--${accessLevel.level.toLowerCase()}`
+    return this.driver.findElements(By.css(`.--${accessLevel.level.toLowerCase()}`));
+  }
 }
 
 module.exports = WorkspacePage;
--- a/features/page-objects/WorkspacePage.js
+++ b/features/page-objects/WorkspacePage.js
@@ -22,14 +22,16 @@ class WorkspacePage extends Page {
   }
 
   async enterRoomWithAccessCode(room, accessCode) {
+    this.driver.manage().deleteAllCookies();
+    this.enter();
     await this.enterRoom(room);
 
-    const placeholderSelector = By.css("[placeholder='Access Code']");
-    await this.driver.wait(until.elementLocated(placeholderSelector));
-    const accessCodeInput = await this.driver.findElement(placeholderSelector);
+    const inputSelector = By.css("[id='waiting_room_access_code']");
+    await this.driver.wait(until.elementLocated(inputSelector));
+    const accessCodeInput = await this.driver.findElement(inputSelector);
     accessCodeInput.sendKeys(accessCode);
 
-    const submitInput = await this.driver.findElement(By.css("[value='Submit']"));
+    const submitInput = await this.driver.findElement(By.css("[type='submit']"));
     submitInput.click();
   }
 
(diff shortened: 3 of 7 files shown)


zspencer:
Would you be able to amend the commit to include CJ, Egbet, Kelly and my Github Emails?

```
Co-authored-by: CJ Joulain <cjoulain@users.noreply.github.com>
Co-authored-by: KellyAH <KellyAH@users.noreply.github.com>
Co-authored-by: Egbet <eg-bet@users.noreply.github.com>
```

convene-web/app/views/workspaces/_room_card.html.erb
```diff
@@ -1,4 +1,4 @@
-<li class="col-span-1 bg-white rounded-lg shadow mb-4" id="<%= room.name %>" >
+<li class="col-span-1 bg-white rounded-lg shadow mb-4 <%= "--#{room.access_level}" %>" id="<%= room.name %>">
```
zspencer, line 1: Not-blocking: Something that may be nice would be to also include a `room-card` class to the `li` so we can key off of that as well.

convene-web/app/models/room.rb
```diff
@@ -19,6 +19,7 @@ class Room < ApplicationRecord
   # `locked` indicates that only participants who know the rrooms `access_code` may access the room.
   # `internal` indicates that only participants who are Workspace Members _or_ know the Workspaces `access_code` may
   # access the room.
+  # TODO Return :unlocked by default, add database non null
```
zspencer, line 22: Let's do this in it's own PR

features/page-objects/WorkspacePage.js
```diff
@@ -48,6 +48,11 @@ class WorkspacePage extends Page {
     assert.equal(videoPanels.length, 1, `${videoPanels.length} was found.`)
     return await this.driver.findElement(jitsiConferenceFrame);
   }
+
+  roomCardsWhere({ accessLevel }) {
+    const selector = `.--${accessLevel.level.toLowerCase()}`
+    return this.driver.findElements(By.css(`.--${accessLevel.level.toLowerCase()}`));
```
zspencer, line 54: Not blocking, but one way we could start to distribute responsibility more broadly would be to move the responsibility for knowing how to find the accessLevel element into the other parameterTypes. I think `roomCardsWhere` still makes sense on the Workspace object, and doing the actual `driver.findElements()` within it feels good to me.

```
// in parameterTypes/workspacePage.js
class WorkspacePage {
  // ...
  roomCardsWhere({ accessLevel }) {
    return this.driver.findElements(accessLevel.selector))
  }
  // ...
}

// in parameterTypes/accessLevel.js
class AccessLevel {
  // See the MDN Getter docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get
  get selector() {
    return By.css(`.--${this.level.toLowerCase()}`)
  }
}
```

features/parameter-types/rooms.js
```diff
@@ -33,8 +33,8 @@ class Room {
 // See: https://github.com/zinc-collective/convene/issues/41
 defineParameterType({
   name: "accessLevel",
-  regexp: /(an |a )?(Unlocked|Internal|Locked)/,
-  transformer: (level) => new AccessLevel(level),
+  regexp: /(an |a |the )?(Unlocked|Internal|Locked)/,
```
zspencer, line 36: Not blocking: We may want to remove the `?` from this regex, so that we don't have to deal with different numbers of arguments being passed into the Transformer.

features/steps/room_steps.js
```diff
@@ -1,9 +1,15 @@
...
+const Workspace = require("../parameter-types/workspaces");
+const Room = require("../parameter-types/rooms");
 const assert = require('assert').strict;
 
-Given("a Workspace with {accessLevel} {room}", function (accessLevel, room) {
-  // Write code here that turns the phrase above into concrete actions
-  return "pending";
+Given("a Workspace with {accessLevel} {room}", async function (accessLevel, room) {
+  const workspace = new Workspace("System Test");
+  this.workspace = new WorkspacePage(this.driver, workspace);
```
zspencer, line 9: Not blocking, something to think about is we have a WorkspacePage and a Workspace object and I'm not sure where the boundaries between these two things are. Does it make sense for there to be two different classes? Or could we do something to add the `driver` to the `Workspace` class? I'm not really sure. Something to noodle on over the next few months.

features/steps/room_steps.js
```diff
@@ -1,9 +1,15 @@
...
+  await this.workspace.enter();
+  const lockedRooms = await this.workspace.roomCardsWhere({ accessLevel });
```
zspencer, line 11: Not blocking: I almost wish we had a `fluent interface` for this. I.e. `await this.workspace.enter().roomCardsWhere({ accessLevel })` but I think that may be hard to pull off.

zspencer (commented):
I think this is a really incredible first step! I think there's some efforts that may (or may not) be worth making to place responsibility more squarely where it belongs; but in the meantime I think once we get the assertion in about the error message and (maybe) figure out a way to verify the video panel is not visible we're gonna be good-to-go.

zspencer:
Also, could you add a reference to the GithubIssue we are working on:

```
See: https://github.com/zinc-collective/convene/issues/40
```

convene-web/app/views/workspaces/_room_card.html.erb
user512: 👍 Will be working on this on a separate PR.

features/page-objects/WorkspacePage.js
```diff
@@ -48,6 +50,11 @@ class WorkspacePage extends Page {
     assert.equal(videoPanels.length, 1, `${videoPanels.length} was found.`)
     return await this.driver.findElement(jitsiConferenceFrame);
   }
+
+  roomCardsWhere({ accessLevel }) {
+    const selector = `.--${accessLevel.level.toLowerCase()}`
```
user512, line 55: This line is dead code.

features/page-objects/WorkspacePage.js
user512: I like this.

PR #127, opened 2020-09-23
Add running db:seed to bin/setup, and clarify a couple of bits in the README that confused me

anaulin:
I propose adding `bin/rake db:seed` to the steps in `bin/setup`, because it seems needed for the System Test and other test/demo workspaces to be created.

I'm not sure if there is a downside to having this step always in `bin/setup`, but at first glance `seeds.rb` seems idempotent. An alternative would be to simply add a step in the README for the developer to separately run `bin/rake db:seed`.

I also attempted to clarify a couple of things in the README that I found unclear.

### DIFF
--- a/convene-web/README.md
+++ b/convene-web/README.md
@@ -18,8 +18,8 @@ First, ensure your development environment has:
 1. Node (See [.nvmrc](./.nvmrc) for version)
 1. [PostgreSQL 12]. (Note: For people using [Docker], a [docker-compose.yml]
    file has been included for convenience.)
-1. Copy `[.env.example]` to `.env` and make any changes: `cp .env.example
-   .env`
+1. Copy `convene-web/.env.example` to `convene-web/.env` and make any changes: `cp convene-web/.env.example
+   convene-web/.env`.
 
 Then, run `bin/setup` to install Ruby and Node dependencies and set up the
 database.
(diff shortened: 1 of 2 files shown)


convene-web/README.md
```diff
@@ -18,17 +18,17 @@ First, ensure your development environment has:
 1. Node (See [.nvmrc](./.nvmrc) for version)
 1. [PostgreSQL 12]. (Note: For people using [Docker], a [docker-compose.yml]
    file has been included for convenience.)
-1. Copy `[.env.example]` to `.env` and make any changes: `cp .env.example
-   .env`
+1. Copy `convene-web/.env.example` to `convene-web/.env` and make any changes: `cp convene-web/.env.example
+   convene-web/.env`.
```
anaulin, line 22: @user512 I noticed that there is also a different `.env.example` file in the root of the repo. Not sure if that should also be mentioned here in the README. Please let me know. 🙏

user512 (approved):
Looks great!

convene-web/README.md
user512: The `.env.example` isn't related to `convene-web` so we can leave that.

zspencer:
Thank you! I've issued you 40 patronage points: https://airtable.com/tblndyi9VR3shxtmS/viwxgldsOPhzLxfpV/recyh1og1X7eZD5vj?blocks=hide

Issue #117, opened 2020-09-11
🥔✨ `Space`: `Members`


zspencer:
- https://github.com/zinc-collective/convene/issues/1154
- https://github.com/zinc-collective/convene/issues/1514

People are granted additional authority within a Space by granting them Membership within that Space. 

## Use Cases
- [x] https://github.com/zinc-collective/convene/issues/1514
- [x] 🥔✨ Viewing `Members` of a `Space`
- [x] 🥔✨ Removing `Members` from a `Space`

## Bugs
- [x] https://github.com/zinc-collective/convene/issues/1607
- [x] https://github.com/zinc-collective/convene/issues/1659

anaulin:
Invitation UI is missing the name of the inviter:
![Screen Shot 2021-11-06 at 3 45 05 PM](https://user-images.githubusercontent.com/6729309/140625908-eb70fa9c-e804-43f4-a4f0-7d439c66f1a1.png)

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #132, opened 2020-10-01
Sketch out Feature Definition for Space Membership

zspencer:


user512 (approved):

user512:
We might want to have another scenario on member/ owner invitation timeout and the person being invited cannot join.

PR #133, opened 2020-10-02
Adding Email Invitation Code Times out Scenario

user512:
I wasn't sure if we would like this scenario but this is pretty common on
a lot of applications.


zspencer (approved):
I think it makes sense, what does @colombene or @maximegalon5 think?

PR #134, opened 2020-10-05
40/indicate rooms are locked

user512:
Add icon to indicate rooms are locked or user has already entered correct access code.

I was trying to setup CSS Pseudo-elements so we can relies on classes instead of <i> tag to show icons but can't get it setup.
I was following this guide: https://fontawesome.com/how-to-use/on-the-web/using-with/sass

### DIFF
--- a/convene-web/app/views/workspaces/_room_card.html.erb
+++ b/convene-web/app/views/workspaces/_room_card.html.erb
@@ -2,17 +2,16 @@
   <header>
     <h3><%= room.name %></h3>
     <%- if room.locked? %>
-      <i class="fas fa-lock"></i>
+      <span class="fas fa-lock"></span>
     <%- end %>
     <%- if room.locked? && room.enterable?(current_access_code) %>
-      <i class="ml-2 fas fa-key"></i>
+      <span class="ml-2 fas fa-key"></span>
     <%- end %>
   </header>
   <footer>
     <div class="room-door_enter">
       <%= link_to workspace_room_path(room.workspace, room) do %>
-        <i class="fas fa-phone-alt"></i>
-        <span>Enter Room</span>
+        <span class="icon-solid phone-alt-icon">Enter Room</span>
       <% end %>
     </div>
   </footer>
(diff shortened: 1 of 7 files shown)


zspencer:
@user512  I think this is good enough. We pushed hard to get to a situation where we didn't need extra elements... And I don't think we can get there without someone who is a bit more of a CSS wizard than us.

That said, do you think you could maybe pull the `fas-locked` and similar into the SASS so that the HTML looks like:




```
<header>
    <h3><%= room.name %></h3>
    <%- if room.locked? %>
      <i class="locked-indicator"></i>
    <%- end %>
    <%- if room.locked? && room.enterable?(current_access_code) %>
      <i class="has-key-indicator"></i>
    <%- end %>
  </header>
  <footer>
    <div class="room-door_enter">
      <%= link_to workspace_room_path(room.workspace, room) do %>
        <i></i> <!-- targetable by the css selector `.room-door_enter i` -->
        <span>Enter Room</span>
      <% end %>
    </div>
  </footer>
```

This leaves us affordances for people being able to change the icons for their spaces pure CSS.

zspencer (approved):
Once CI passes, I think this is good to merge!

convene-web/app/views/workspaces/_room_card.html.erb
```diff
@@ -2,17 +2,16 @@
...
-      <i class="ml-2 fas fa-key"></i>
+      <span class="ml-2 fas fa-key"></span>
     <%- end %>
   </header>
   <footer>
     <div class="room-door_enter">
       <%= link_to workspace_room_path(room.workspace, room) do %>
-        <i class="fas fa-phone-alt"></i>
-        <span>Enter Room</span>
+        <span class="icon-solid phone-alt-icon">Enter Room</span>
```
user512, line 14: Follow [naming convention from Font Awesome](https://fontawesome.com/icons/phone-alt?style=solid), so it's named as `phone-alt`.

convene-web/app/views/workspaces/_room_card.html.erb
```diff
@@ -2,17 +2,16 @@
   <header>
     <h3><%= room.name %></h3>
     <%- if room.locked? %>
-      <i class="fas fa-lock"></i>
+      <span class="fas fa-lock"></span>
     <%- end %>
     <%- if room.locked? && room.enterable?(current_access_code) %>
-      <i class="ml-2 fas fa-key"></i>
+      <span class="ml-2 fas fa-key"></span>
```
user512, line 8: Haven't figured out how to move this two icons into pseudo selector.

PR #137, opened 2020-10-08
Checkin missing icons scss file

user512:
Forgot to checkin `icons.scss`.

### DIFF
--- a/convene-web/app/javascript/src/custom_components/icons.scss
+++ b/convene-web/app/javascript/src/custom_components/icons.scss
@@ -1,0 +1,7 @@
+.icon-solid::before {
+  font-family: "Font Awesome 5 Free"; font-weight: 900;
+  @apply mr-2;
+}
+.phone-alt-icon::before {
+  content: "\f879";
+}


user512:
Merging because prod build failed without this file and SonarCloud failure seems to be not important because having a fallback font-family won't help rendering icons.

zspencer:
Thanks Tom! I think the thing that I am trying to get at here is I'd prefer if our applications HTML not be coupled to the Font-Awesome or Tailwind specific CSS classes where possible; instead pushing for a [Semantic](https://html.com/semantic-markup/), [Block/Element/Modifier](http://getbem.com/introduction/) approach. 

So, in this case; what feels "good" to me would be:

`<i class="icon --enter-meeting"></i>`

In this case, the `icon` element is modified by the `--enter-meeting` modifier. This gives us an HTML structure that indicates _what the purpose_ of the `i` tag is, without indicating how that purpose is represented to users. 

Admittedly, this is not "Pure" BEM, which wants something like `icon icon--enter-meeting`. I am not sure what the tradeoffs are between pure BEM and this composable BEM structure; but I'm curious about exploring it from a "does this make the presentation of our product easier for folks to customize?" perspective.

PR #142, opened 2020-10-18
Feature test `Locking an Unlocked Room`

user512:
Wire in steps to test `locking an unlocked room`.

During testing, CI keep failing because `CONFIGURE_ROOM` wasn't enabled, I ended up writing some half-baked code to capture screenshot when test is failing and upload as `test_case.png` as artifact.
I kept those code as comment so it's easier for us to debug on CI until I can figure out how to properly config `cucumber-html-report` and always generate a test report in html and upload as artifact. (Ref: https://github.com/zinc-collective/convene/pull/147)


features/support/env.js
```diff
@@ -1,20 +1,33 @@
...
 
 setWorldConstructor(CustomWorld);
 
-After(function() {
+After(function(testCase) {
+  if (testCase.result.status == Status.FAILED) {
+    return this.driver.takeScreenshot().then( screenShot => {
+      console.log('screenshot', screenShot)
+      this.attach(screenShot, 'image/png');
+      this.driver.quit()
```
KellyAH, line 28: I assume this is making the driver take a screenshot when the test fails, so we have helpful test results when debugging test failures. Where is the screenshot being saved to? To a local directory ignored by git?

features/support/env.js
```diff
@@ -1,20 +1,33 @@
...
-      ).build();
+      .setFirefoxOptions(this.firefoxOption())
+      .build();
+  }
+
+  firefoxOption() {
+    // GitHub Actions set CI to true
+    // Ref: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
+    return process.env.CI ? new firefox.Options().headless() : new firefox.Options()
   }
```
KellyAH, line 18: I don't remember these changes being made in our mob testing session. Aren't these from https://github.com/zinc-collective/convene/pull/146/files ? Should those changes be included in this PR?

features/support/env.js
user512: Still exploring, but the intention is to generate report using `cucumber-html-reporter`.
And it will be a separate PR.

features/support/env.js
user512: Yes, I split out https://github.com/zinc-collective/convene/pull/146/ intentionally so those code will be in its own PR but I need these code to drive today session so I don't need to manually toggle headless mode.
And yes, it will be removed from this PR.

features/support/env.js
user512: I was trying to avoid making test report but somehow the test is failing on CI only and I need a way to view the screenshot to debug.

features/page-objects/WorkspacePage.js
```diff
@@ -54,6 +55,14 @@ class WorkspacePage extends Page {
   roomCardsWhere({ accessLevel }) {
     return this.driver.findElements(By.css(`.--${accessLevel.level.toLowerCase()}`));
   }
+
+  async enterConfigureRoom(room) {
+    const roomCard = await this.findRoomCard(room);
+    console.log('roomCard', roomCard)
+    const linkText = await roomCard.findElement(By.partialLinkText("Configure Room"));
```
user512, line 62: Wasn't sure why Configure Room link text can't be found on CI.

features/parameter-types/rooms.js
```diff
@@ -33,7 +33,7 @@ class Room {
 // See: https://github.com/zinc-collective/convene/issues/41
 defineParameterType({
   name: "accessLevel",
-  regexp: /(an |a |the )?(Unlocked|Internal|Locked)/,
+  regexp: /(an |a |the |is )?(Unlocked|Internal|Locked)/,
```
user512, line 36: Think of a better way to solve this.

user512:
still need to figure out a way to make test fail and be able to upload artifact. Now I am doing an ugly `exit(0)` after writing png file which is bad because the GH action now consider the test passed.

features/page-objects/WorkspacePage.js
user512: Because feature flag was off.

features/page-objects/RoomSettingPage.js
```diff
@@ -0,0 +1,34 @@
...
+
+  async lock(accessCode) {
+    await this.setAccessLevel('locked');
+    await this.setAccessCode(accessCode);
+    await this.clickUpdateRoom();
+  }
+
+  async setAccessLevel(accessLevel) {
+    const accessLevelSelect = await this.findByCss("[id='room_access_level']");
+    accessLevelSelect.click();
```
zspencer, line 17: Not Blocking: Something that may be (but may not be) worth doing is leveraging [JavaScripts `getter` syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) so that we could do something like this:

```javascript
class RoomSettingsPage < Page {
  get accessLevelSelector() {
    return this.findByCss("[id='room_access_level']")
  }

  async setAccessLevel(accessLevel) {
    (await this.accessLevelSelect).click();
    (await this.findByCss(`[value=${accessLevel}]`)).click();
  }
}
```

We could also sprout a "SelectBox" class that takes responsibility for the asyncronicity of clicking/selecting options; but that increases the kinds of test-driver-y objects in the system to Drivers + Page Objects + Component Objects. 

That would look something like this:

```javascript
class SelectTag < Component {
  // this constructor and the component method probably wants to live in Component
  constructor(selector, driver) {
  }
  get component() {
    return this.driver.findByCss(this.selector)
  }
  select(value) {
    return this.component
     .then((el) => el.click())
     .then(this.findByCss(`${selector} [value=${accessLevel}]`))
  }
}
```

I think we'd want to explore this as a group at our next session. Thoughts @KellyAH / @cjoulain / @user512  and @eg-bet?

features/parameter-types/rooms.js
zspencer: I think we may want a set of article adjectives that are shared across the parameter types; perhaps there's a way to interpolate a constant into a regex? (Dunno if this syntax is correct; but this is the rough shape)

```javascript
const FLEXIBLE_ARTICLE_ADJECTIVES="( |an |the |is |a )"

// ...
defineParameterType({
  //...
  regexp: /${FLEXIBLE_ARTICLE_ADJECTIVES}(Unlocked|Internal|Locked)/
})
```

zspencer (approved):

zspencer:
Wow! This is amazing! Great work @KellyAH, @eg-bet and @user512! Looking forward to next session!!!

features/support/env.js
```diff
@@ -19,7 +20,25 @@ class CustomWorld {
...
+  //     fs.writeFile(filePath, screenShot, { encoding: 'base64' }, err => {
+  //       if (err) console.log(err)
+  //       console.log("Screenshot created: ", filePath)
+  //       process.exit(0);
+  //     })
+  //     // TODO: Uncomment this once figure out how to build test report
+  //     // this.attach(screenShot, 'image/png');
+  //     this.driver.quit();
+  //   });
+  // }
```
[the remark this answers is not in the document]
user512: Extracted in #148

features/parameter-types/rooms.js
```diff
@@ -1,4 +1,6 @@
 const { defineParameterType } = require("cucumber");
+const concatRegExp = require('../utilities/concatRegExp');
+const FLEXIBLE_ARTICLE_ADJECTIVES = /(an |the |is |a )/;
```
zspencer, line 3: Not blocking: The union of these two pieces feels like a meaningful/useful part of the program; as they are used together.

Perhaps it's `withArticleAdjectives` and it's used like:

`regexp: withArticleAdjectives(/(Unlocked|Internal|Locked)/)` or `regexp: withArticleAdjectives(/( |correct |wrong )?Room Key/)` etc?

PR #146, opened 2020-11-01
Run feature test in headless mode on CI and non headless mode during development

user512:
Most of the time we need non headless mode during feature test development locally but headless mode on CI.


zspencer (approved):
Nicely done.

PR #149, opened 2020-11-12
Convene may Send Email

zspencer:
TODO:
- [x] How do we do mailer previews?
  * Developer can preview mailer by visiting http://localhost:3000/rails/mailers/
- [x] Probably should include useful documentation for people
  * Wrote mailer development documentation in convene-web/README.md
- [x] TEST? Does this even work?


anaulin:
For the record, when testing this locally on @user512 's machine we had to change the port from 587 to 465 to make it work. It is possible that in his local net that port is blocked. For future reference, here is Mailgun's guidance on which ports to use: https://www.mailgun.com/blog/which-smtp-port-understanding-ports-25-465-587/

user512:
Just tested on Heroku review app and it works.
But then we set those mailgun creds on review app manually.
We should eventually figure out a way to pre-populate these creds for easier testing.

anaulin (approved):
I believe this is ready to be merged. @zspencer wdyt?

zspencer:
Absolutely lovely! Thank you!

PR #150, opened 2020-11-12
Setup Passwordless Login

user512:
Setup passwordless login but we are not using it for now.
People are created when identification happens if we don't already have a person matching the email.
* We allow people to be created when identification happens because we might allow guest to enter room and provide id information.
* Route for the signing form: `/people/sign_in`

### DIFF
--- a/convene-web/Gemfile
+++ b/convene-web/Gemfile
@@ -62,3 +62,4 @@ end
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
 gem 'active_record_extended'
+gem 'passwordless'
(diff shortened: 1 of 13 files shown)


anaulin:
For future reference, here is the issue I opened in the passwordless repo to get guidance on the `uuid` issue: https://github.com/mikker/passwordless/issues/91

anaulin:
FYI, now the recommended approach in the Passwordless README to use `uuid` is exactly what we did here (https://github.com/mikker/passwordless#supporting-uuid-primary-keys), so I think we can continue with this spike if/when we're ready without worrying about the `uuid` thing.

user512:
That’s great! Thank you.

Issue #154, opened 2020-11-26
Furniture: Breakouts


zspencer:
### TODO
- [ ] Write a feature definition
- [ ] Spike an implementation (with or without "furniture" concept)
- [ ] Spike how to have events like `workspace.room.breakout-rooms.created` and `workspace.room.breakout-rooms.expiring-soon` etc. distributed to each room occupant.
- [ ] Migrate the feature definition to the code-base
- [ ] Creating Breakout Rooms
  - [ ] Sketch Feature Definition
  - [ ] Make a UI Mockup
  - [ ] Wire in step definitions
  - [ ] Implement it!
- [ ] Entering Breakout Rooms
  - [ ] Sketch Feature Definition
  - [ ] Make a UI Mockup
  - [ ] Wire in step definitions
  - [ ] Implement it!
- [ ] Enable feature globally!

```feature/breakout-rooms.feature
Feature: Breakout Rooms
  In order for people to organize themselves by topic
  Members want to be able to create named breakout rooms within a room
  And Guests want to move between those rooms leisurely

Scenario: Creating Breakout Rooms
  Given a Room with Breakout Rooms enabled
   When a Member adds the following Breakout Rooms
    | name                           | expires        | access   | 
    | Business Model Canvas Workshop | in 1 hour.     | public.  | 
    | Engineering.                   | in 3 hours     | internal | 
    | Standup.                       | in 15 minutes. | public.  | 
   Then the "Engineering" Breakout Room is visible to:
    | Members | 
    | Owners  | 
    And the "Engineering" Breakout Room is visible to:
    | Guests | 
   Then the "Standup" Breakout Room is visible to:
    | Members | 
    | Owners  | 
    | Guests  | 
   Then the "Business Model Canvas" Breakout Room is visible to:
    | Members | 
    | Owners  | 
    | Guests  | 

Scenario: Entering Breakout Rooms

Scenario: Breakout Room Expiration
```

PR #155, opened 2020-11-26
Room Occupant may take a Seat at a Breakout Table

zspencer:
https://github.com/zinc-collective/convene/issues/154

### DIFF
--- a/convene-web/app/models/furniture_placement.rb
+++ b/convene-web/app/models/furniture_placement.rb
@@ -1,0 +1,21 @@
+# {Furniture} placed in a {Room} allows it to be used
+class FurniturePlacement < ApplicationRecord
+  belongs_to :room
+  delegate :workspace, to: :room
+
+  attribute :name, :string
+  validates :name, uniqueness: { scope: :room_id }
+
+  attribute :settings, :json
+
+  # TODO: Ordering of records is likely handled by a well-maintained gem already
+  # The order in which {Furniture} is rendered in a Room. Lower is higher.
+  attribute :slot, :integer
+  validates :slot, presence: true, uniqueness: { scope: :room_id }
+
+  # TODO: Furniture probably wants to have a Registry that finds and
+  # instantiates the appropriate furniture class for the placement.
+  def furniture
+    @furniture ||= Furniture::BreakoutTables.new(self)
+  end
+end
--- a/convene-web/app/views/rooms/show.html.erb
+++ b/convene-web/app/views/rooms/show.html.erb
@@ -6,12 +6,20 @@
   id="meet"
   data-controller="video-room"
   data-target="video-room.wrapper"
-  data-video-room-video-host="<%= current_workspace.jitsi_meet_domain %>"
-  data-video-room-name="<%= video_room_name(current_workspace, current_room) %>"
-  data-video-room-workspace-path="<%= end_call_path(current_workspace) %>"
+  data-video-room-video-host="<%= room.video_host %>"
+  data-video-room-name="<%= room.full_slug %>"
+  data-video-room-workspace-path="<%= workspace_path(room.workspace) %>"
   class="bg-gray-800"
 >
 </div>
+
+<%- room.furniture_placements.each do |furniture_placement| %>
+  <%= render partial: furniture_placement.furniture.in_room_template,
+             locals: { furniture: furniture_placement.furniture, room: room,
+                       person: current_person } %>
+<% end %>
+
+
 
 <div class="px-6 pt-3">
   <ul class="mt-3 grid grid-cols-1 gap-5 sm:gap-6 sm:grid-cols-2 lg:grid-cols-4">
--- a/convene-web/config/environments/test.rb
+++ b/convene-web/config/environments/test.rb
@@ -12,7 +12,7 @@ Rails.application.configure do
   # Do not eager load code on boot. This avoids loading your whole application
   # just for the purpose of running a single test. If you are using a tool that
   # preloads Rails for running tests, you may have to set it to true.
-  config.eager_load = false
+  config.eager_load = true
 
   # Configure public file server for tests with Cache-Control for performance.
   config.public_file_server.enabled = true
--- a/convene-web/config/routes.rb
+++ b/convene-web/config/routes.rb
@@ -1,26 +1,29 @@
-
-
 Rails.application.routes.draw do
   passwordless_for :people
 
   namespace :admin do
-      resources :workspaces
-      resources :clients
-      resources :people
-      resources :workspace_memberships
-      resources :room_ownerships
-      resources :rooms
+    resources :workspaces
+    resources :clients
+    resources :people
+    resources :workspace_memberships
+    resources :room_ownerships
+    resources :rooms
 
-      root to: "workspaces#index"
-    end
+    root to: 'workspaces#index'
+  end
+
   # For details on the DSL available within this file, see https://guides.rubyonrails.org/routing.html
   resources :workspaces do
-    resources :rooms, only: [:show, :edit, :update] do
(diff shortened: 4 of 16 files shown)


anaulin:
The direction this is going in makes sense to me, but it feels like it needs more to flesh out the fine detail of the interactions / data flow.

Quick thoughts after looking through the diff:
* What's the connection between `BreakoutRoom` and `Furniture`? I don't see any connections between the models in this PR, but I would expect `BreakoutRoom` to be a type of `Furniture` in some way?
* Related to the point above, I don't see changes to `schema.rb` -- seems like that is missing from the PR.
* Having a different partial for each type of `Furniture` to render it in the UI makes sense to me at first glance.
* How will we handle e.g. creating a BreakoutRoom? Is that a different furniture-type-specific partial to create/edit?

convene-web/app/views/rooms/show.html.erb
```diff
@@ -13,6 +13,20 @@
 >
 </div>
 
+<%- room.furniture_placements.each do |furniture_placement| %>
+  <div
+    class="<%=furniture_placement.furniture_slug%>"
+    data-controller="<%=furniture_placement.furniture_slug %>"
+    <%- furniture_placement.data_tuples.each do |name, value| %>
```
anaulin, line 20: What are these `data_tuples`? Is that an intended field on `FurniturePlacement` that is currently missing from the model, or something that Rails magically provides? (Quick search for this didn't shed light, sadly.)

zspencer: My intent was for this to be a way to expose data to stimulus: https://stimulusjs.org/handbook/managing-state#the-data-api-explained. It's a bit of "coding by wishful thinking" in that I never actually defined them; but essentially they are a mechanism for a piece of furniture to expose data to the front-end.

zspencer:
Re: What's the connection between BreakoutRoom and Furniture? I don't see any connections between the models in this PR, but I would expect BreakoutRoom to be a type of Furniture in some way?

I'm not sure yet;  my gut agrees with you that a BreakoutRoom is likely a _kind of_ furniture; and whether we go with Furniture as a Mixin type inheritance or Parent Child will depend on what makes sense when we have several types of furniture.

Re: Related to the point above, I don't see changes to schema.rb -- seems like that is missing from the PR.

Yep, I'm "coding by wishful thinking" here and sketching very lightly and not including some of the details that would make this code "work"; more trying to play with how the different objects interact.

Re: Having a different partial for each type of Furniture to render it in the UI makes sense to me at first glance.

:+1:

Re: How will we handle e.g. creating a BreakoutRoom? Is that a different furniture-type-specific partial to create/edit?

Hmmm, my preference would be for the piece of furniture to fully encapsulate it's data-model. I.e. the BreakoutRooms furniture would hold responsibility for deciding who is allowed to perform what CRUD operations on it's internal state based upon the rights the person has within the room. I need to noodle a bit more on how the interface for that would work; and a 0.1 version of BreakoutRooms may not have that degree of control.

convene-web/app/views/rooms/show.html.erb
zspencer: I've ditched this in favor of each piece of Furniture being able to define it's own partial, so that we're not coupling implementation details of how we pass data to the UI too early..

zspencer:
@anaulin / @user512 - Made a bit more progress on this, some of which probably wants to be it's own PR to reduce noise (Grouping the scopes with the fields they're querying in the room model; for instance...) 

There's also some work still to be done before I feel confident that it's worth trying to build a second kind of furniture with; but it feels like it's getting into a bit better shape...

Thoughts?

anaulin:
@zspencer this direction makes sense to me, and it is nice to see it start to take shape!

To me it makes sense to get something like this checked in, and use it to get ourselves to play with "Tables" and iterating on them as needed. Later we can refine things to support more than one type of architecture, incorporating whatever we learn from "Tables".

(From a product standpoint, users are used to "Breakout Rooms" as terminology, so maybe we can call ours "Breakout Tables" or something like that? Otherwise I fear that if we call them externally "Tables" that would just be confusing to folks.)

zspencer:
I'm a bit torn. I feel like "Breakout Rooms" breaks the [skeuomorphic mental model](https://blueprintdigital.com/skeuomorphic-design-the-pros-and-cons/) of the "Rent a Space, Lay it out with Rooms, add Furniture" framing that I'm using in the customer/contributor-development cycle.

That said, that attachment may be unproductive at the moment.

user512:
I think table is fine and potentially is a good term if we foresee other furniture will be added in the future and I like the idea of getting this checked in and change if we have stronger feeling to change later.

> Realtime! When someone creates a breakout room; we want Guests to see that become available right away.

We can look into using [Turbo](https://turbo.hotwire.dev/), where someone created a table, the app would stream partial page updates over websocket. I haven't try using this library but by watching this [demo](https://gorails.com/episodes/hotwire-rails), it seems promising. 

> How do we kick people out of the breakout room when they expire?

Implement some JS to check if the room is expired every few minutes and redirect people if room is expired.

anaulin:
I don't have super strong feelings on naming, but to clarify that above I was proposing something like "Breakout Tables", kinda splitting the difference (not "Breakout Rooms", as "Rooms" would be confusing in various ways in our model).

convene-web/app/lib/furniture/tables.rb
```diff
@@ -0,0 +1,67 @@
+module Furniture
+  class Tables
```
user512, line 2: How do you feel about using plural for namespace and singular for model?
```ruby
module Furnitures
  class Tables
    ...
  end
end
```

zspencer:
That makes sense, had a bit of a brain-fart! Updated to `BreakoutTables` and added in specs for the FurniturePlacement class; as well as fixed the breakages in the feature tests.

convene-web/app/models/furniture_placement.rb
```diff
@@ -0,0 +1,21 @@
...
+class FurniturePlacement < ApplicationRecord
+  belongs_to :room
+  delegate :workspace, to: :room
+
+  attribute :name, :string
+  validates :name, uniqueness: { scope: :room_id }
+
+  attribute :settings, :json
+
+  # TODO: Ordering of records is likely handled by a well-maintained gem already
```
user512, line 11: Which gem were you referring to here?

convene-web/config/routes.rb
```diff
@@ -1,26 +1,29 @@
...
-      root to: "workspaces#index"
-    end
   # For details on the DSL available within this file, see https://guides.rubyonrails.org/routing.html
   resources :workspaces do
-    resources :rooms, only: [:show, :edit, :update] do
-      resource :waiting_room, only: [:show, :update]
+    resources :rooms, only: %i[show edit update] do
+      resource :waiting_room, only: %i[show update]
+      namespace :furniture do
+        resources :tables, only: [:show], controller: 'breakout_tables/'
```
user512, line 20: [Non-blocking] There is no show action in `BreakoutTables`, I assume this is intentional to be an example on how to add routes to point to this controller?

user512 (approved):
I think how furniture works is still in flux (like the shape of data in breakout table setting) but this seems safe to merge and developer can keep exploring how breakout table works.

convene-web/config/environments/test.rb
```diff
@@ -12,7 +12,7 @@
   # Do not eager load code on boot. This avoids loading your whole application
   # just for the purpose of running a single test. If you are using a tool that
   # preloads Rails for running tests, you may have to set it to true.
-  config.eager_load = false
+  config.eager_load = true
```
user512, line 15: Let me know if the solution in https://github.com/zinc-collective/convene/pull/172 looks good to you.
It can be merge as a separate PR or just merged into this branch.

Issue #40, opened 2020-07-03
Locked Rooms


user512:
### Done
 - [x] Locking a Room
    - [x] [Wireframe](https://xd.adobe.com/view/fd425dbe-5384-44c9-997a-eeee6e886a86-a811/screen/847810bf-5d62-4131-a70d-d9efdfadb334)
    - [x] Define [Scenario](https://github.com/zinc-collective/convene/blob/70b65c82148c20db0c2b580715b0825325ccfbe6/features/locking-rooms.feature#L41-L44)
    - [x] Wire in Steps
    - [x] Implement Feature
  
 - [x] Unlocking a Room
    - [x] [Wireframe](https://xd.adobe.com/view/fd425dbe-5384-44c9-997a-eeee6e886a86-a811/screen/847810bf-5d62-4131-a70d-d9efdfadb334)
    - [x] Define [Scenario](https://github.com/zinc-collective/convene/blob/70b65c82148c20db0c2b580715b0825325ccfbe6/features/locking-rooms.feature#L61-L64)
    - [x] Wire in Steps
    - [x] Implement Feature
 - [x] Failing to enter a locked room
   - [x] Define [Scenario](https://github.com/zinc-collective/convene/blob/70b65c82148c20db0c2b580715b0825325ccfbe6/features/locking-rooms.feature#L28)
   - [x] Wire in Steps
   - [x] Implement Feature
- [x] [Demo workspace has a Locked Room with the password “friend”](https://github.com/zinc-collective/convene/commit/a36e2c2d9e0a7f5e05df127af3e9dfec9e94cf56)
 - [x] Access Code input becomes editable only when room is locked (per @colombene, see: https://github.com/zinc-collective/convene/pull/139#issuecomment-709625969) (https://github.com/zinc-collective/convene/pull/144)
- [x] Document the features intent within [`/features/locking-rooms.feature`](https://github.com/zinc-collective/convene/blob/development/features/locking-rooms.feature) (https://github.com/zinc-collective/convene/pull/54)
 - [x] Entering a Locked Room
    - [x] [Wireframe](https://xd.adobe.com/view/fd425dbe-5384-44c9-997a-eeee6e886a86-a811/screen/04ee266e-931b-4bde-bcf9-af94c7ac444e/)
    - [x] Define [Scenario](https://github.com/zinc-collective/convene/blob/70b65c82148c20db0c2b580715b0825325ccfbe6/features/locking-rooms.feature#L23-L28)
[...]

PR #157, opened 2020-11-29
Improve composability of story tests and step definitions

zspencer:
See: https://github.com/zinc-collective/convene/issues/40

The changes here are intended to make it so our feature files and step definitions don't have to care as much about hardcoded values; thus making them easier to:
- Recompose into new story tests and...
- Parallelize the feature tests 

There's likely some more work to be done here; but this feels like a decent, safe step?

### DIFF
--- a/features/page-objects/RoomSettingPage.js
+++ b/features/page-objects/RoomSettingPage.js
@@ -1,10 +1,9 @@
 const Page = require("./Page");
 
 class RoomSettingPage extends Page {
-  constructor(driver, room, accessCode) {
+  constructor(driver, room) {
     super(driver);
     this.room = room;
-    this.accessCode = accessCode;
   }
 
   async lock(accessCode) {
@@ -13,17 +12,19 @@ class RoomSettingPage extends Page {
     await this.clickUpdateRoom();
   }
 
-  async unlock() {
-    await this.enterAccessCode()
+  async unlock(accessCode) {
+    await this.enterAccessCode(accessCode)
+
     await this.setAccessLevel('unlocked');
     await this.clickUpdateRoom();
   }
 
-  async enterAccessCode() {
+  async enterAccessCode(accessCode) {
     const accessCodeInput = await this.findByCss("[id='waiting_room_access_code']");
-    accessCodeInput.sendKeys(this.accessCode);
+    accessCodeInput.sendKeys(accessCode.value);
+
     const submitInput = await this.findByCss("[type='submit']");
-    submitInput.click();
+    return submitInput.click();
   }
 
   async setAccessLevel(accessLevel) {
@@ -36,12 +37,12 @@ class RoomSettingPage extends Page {
 
   async setAccessCode(accessCode) {
     const accessCodeInput = await this.findByCss("[id='room_access_code']");
-    accessCodeInput.sendKeys(accessCode);
+    accessCodeInput.sendKeys(accessCode.value);
   }
 
   async clickUpdateRoom() {
     const submitInput = await this.findByCss("[type='submit']");
-    submitInput.click();
+    return submitInput.click();
   }
 
   async accessCodeError() {
--- a/features/parameter-types/rooms.js
+++ b/features/parameter-types/rooms.js
@@ -1,4 +1,5 @@
 const { defineParameterType } = require("cucumber");
+const { By } = require('selenium-webdriver');
 const concatRegExp = require('../utilities/concatRegExp');
 const FLEXIBLE_ARTICLE_ADJECTIVES = /(an |the |is |a )/;
 
@@ -28,6 +29,17 @@ class Room {
     this.name = roomName;
     this.slug = slugify(roomName);
   }
+
+  reinitialize({ accessLevel }) {
+    this.accessLevel = accessLevel
+  }
+
+  get cardLocator() {
+    if(this.accessLevel) {
+      return this.accessLevel.locator
+    }
+    return By.id(this.name)
+  }
 }
 
 // This matches steps based on the access control model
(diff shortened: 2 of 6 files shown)


features/page-objects/RoomSettingPage.js
```diff
@@ -13,17 +12,19 @@ class RoomSettingPage extends Page {
...
-    await this.enterAccessCode()
+  async unlock(accessCode) {
+    await this.enterAccessCode(accessCode)
+
     await this.setAccessLevel('unlocked');
     await this.clickUpdateRoom();
   }
 
-  async enterAccessCode() {
+  async enterAccessCode(accessCode) {
```
KellyAH, line 22: `enterAccessCode` and `setAccessCode` are named so similar it's a little confusing. `enterAccessCode` is doing more than just inputting the access code, it looks like it's also clicking submit button with `submitInput.click()`. So maybe rename `enterAccessCode` to `submitAccessCode`?

features/page-objects/RoomSettingPage.js
```diff
@@ -36,12 +37,12 @@ class RoomSettingPage extends Page {
 
   async setAccessCode(accessCode) {
     const accessCodeInput = await this.findByCss("[id='room_access_code']");
```
KellyAH, line 39: I assume all these async functions are controlling the same page: `RoomSettingPage`. So why is the`enterAccessCode` function and `setAccessCode` function setting `accessCodeInput` using 2 different ids?

`id='waiting_room_access_code'` and  `id='room_access_code'`. Are they interacting with the same field and that field's id selector changes based on the current state of the page? Or are these 2 different fields?

features/parameter-types/rooms.js
```diff
@@ -59,8 +75,25 @@ class PublicityLevel {
 }
 
 defineParameterType({
-  name: "roomKey",
-  regexp: concatRegExp(FLEXIBLE_ARTICLE_ADJECTIVES, /(correct |wrong )?Room Key/)
+  name: "accessCode",
+  regexp: concatRegExp(FLEXIBLE_ARTICLE_ADJECTIVES, /(correct |valid |wrong |empty )?Access Code/),
```
KellyAH, line 79: add a comment explaining why both `correct` and `valid` must be handled in this file. Is this just to accommodate variations of cucumber strings and to not break existing tests?

features/steps/room_steps.js
```diff
@@ -8,37 +8,42 @@ const assert = require('assert').strict;
 const { By, until } = require('selenium-webdriver');
 const RoomSettingPage = require("../page-objects/RoomSettingPage");
 
+/**
+ * Merges extracted parameter types together for convenience within step definitions
+ */
```
KellyAH, line 13: This could use more comments explaining why this function makes things convenient and an example of the ideal way to invoke it.

KellyAH (changes requested):

features/parameter-types/rooms.js
zspencer: Good point!

zspencer: Great point! Yes, this is to make it a bit easier to use in steps where either `correct` or `valid` would feel funky

features/page-objects/RoomSettingPage.js
zspencer: Great catch! 

I wonder if what we _really_ want here is to decompose the responsibility for interacting with the WaitingRoom  from the RoomSettingsPage; as that I think is the root cause for the mismatched IDs that you noticed further down.

```
await waitingRoom.submitAccessCode("value");
```

features/page-objects/RoomSettingPage.js
zspencer: It _looks like_ we probably are conflating two distinct pages; the "WaitingRoom" page, which someone who is not allowed into the room without entering the access code; and the "RoomSettingsPage" which is where a Space owner _sets_ the access code for the space.

May be worth splitting those out!

features/steps/room_steps.js
zspencer: Yea, I think In need to up my jsdoc fluency, because I was like "ehhhh, I don't want to figure out how to annotate this function definition!" even though I _knew_ I would want an example and information about what parameters it was expecting and why.

features/page-objects/RoomSettingPage.js
KellyAH: Agreed. The "WaitingRoom" page and "RoomSettingsPage" might display the same elements right now, but if the elements have different ids, then they should have separate page classes. Else it'll get confusing. Also if the pages need to execute the same functions, those functions can be stored in a separate file to be DRY.

The "RoomSettingsPage" may also change in the future when extra settings options are added.

features/steps/room_steps.js
KellyAH: https://jsdoc.app/tags-example.html seems helpful. Tho I personally prefer adding an empty new line in between examples and long sentences for easy scannability and readability.

user512 (approved):

features/parameter-types/rooms.js
user512: Just realized valid and correct is two different concept.

user512:
Merging this and attempt to resolve @KellyAH comment in next PR.

features/parameter-types/rooms.js
zspencer: They are distinct concepts; however in this particular case they are synonyms. I don't know whether exposing the synonyms is clarifying or confusing.

features/page-objects/RoomSettingPage.js
```diff
@@ -36,12 +37,12 @@ class RoomSettingPage extends Page {
...
-    accessCodeInput.sendKeys(accessCode);
+    accessCodeInput.sendKeys(accessCode.value);
```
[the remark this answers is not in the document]
zspencer: Maybe? Does `sendKeys` return a `Promise`? If so, then yes so that `await`s/`then`s chained onto `setAccessCode` don't accidentally execute before the `sendKeys` operation has completed.

PR #163, opened 2020-12-13
Avatar with profile menu dropdown

user512:
This feature is hidden behind the `IDENTIFICATION` feature flag.
Added an avatar and dropdown menu on the header and show log in and log out option depends on if the user have identified themselves.

Pending item to complete this:
- [ ] Get a better avatar from Colombene?
- [ ] Make sign in page looks better
- [ ] Make redirect to workspace page instead of root path


convene-web/app/models/feature.rb
```diff
@@ -8,6 +8,7 @@ def self.enabled?(feature)
     return true if feature == :demo && ENV['DEMO_ENABLED']
     return true if feature == :system_test && ENV['SYSTEM_TEST']
     return true if feature == :configure_room && ENV['CONFIGURE_ROOM']
+    return true if feature == :identification && ENV['IDENTIFICATION']
```
zspencer, line 11: I'm beginning to get a bit worried that we're putting a so much configuration into distinct ENV variables; perhaps it's worth turning the Feature model into a "real model" that stores features in a comma-separated list or something? Not urgent by any means; but may be worth adding a `TODO` to the codebase for someone who is interested in doing some of that kind of work!

convene-web/app/views/layouts/_header.html.erb
```diff
@@ -1,9 +1,38 @@
...
+            <ul class="profile-menu-dropdown hidden" data-target="menu.items">
+              <li>
+                <%= link_to "Login", people.sign_in_path, class: "profile-menu-dropdown-item" %>
+              </li>
+            </ul>
+          <%- end %>
+
+        </div>
+
+        <%= button_tag "", type: "button", class: "icon-solid chevron-icon", "data-action": "click->menu#toggle click@window->menu#hide",  "data-target": "menu.button" %>
```
zspencer, line 29: I'm a bit worried about the accessibility implications of having a button without any text.

convene-web/app/views/layouts/_header.html.erb
```diff
@@ -1,9 +1,38 @@
...
+      <%- end %>
+
+      <div>
+        <h3>Convene</h3>
+        <h3 class="font-hairline">video meeting spaces</h3>
```
zspencer, line 34: This structure implies that Video Meeting Spaces and Convene are two distinct "headings" within the information hierarchy of the interface; as opposed to `video meeting spaces` being a tagline that modifies the h3. See: https://www.w3.org/TR/html51/common-idioms-without-dedicated-elements.html#common-idioms-without-dedicated-elements for more context.

convene-web/app/views/layouts/_header.html.erb
```diff
@@ -1,9 +1,38 @@
...
+
+      <%- if Feature.enabled?(:identification) %>
+        <div class="profile-menu">
+          <figure class="avatar">
+            <%= image_tag current_person.avatar_url, size: 30 %>
+            <figcaption><%= current_person.name %></figcaption>
+          </figure>
+
+          <%- if current_person.is_a? Person %>
+            <ul class="profile-menu-dropdown hidden" data-target="menu.items">
```
zspencer, line 14: This `ul` probably wants to be a `nav`, and since we only have one element we may not need a `ul` as well. See: https://www.w3.org/TR/html51/sections.html#the-nav-element

convene-web/app/views/layouts/_header.html.erb
zspencer, line 29: The `icon-solid` and `chevron-icon` classes don't convey the purpose of the button, and instead convey the underlying design values. We are trying to hold the responsibility for declaring the purpose of elements with the class, and holding the implementation within the CSS. Perhaps `class: "navigation-toggle --closed"` to indicate that the related elements are currently not being shown (this will also let us do cool things like, change the icon when we swap the class to `--open`!)

convene-web/app/views/layouts/_header.html.erb
```diff
@@ -1,9 +1,38 @@
...
+              <li>
+                <%= link_to "Logout", people.sign_out_path, class: "profile-menu-dropdown-item" %>
+              </li>
+            </ul>
+          <%- elsif %>
+            <ul class="profile-menu-dropdown hidden" data-target="menu.items">
```
zspencer, line 20: The `hidden` class is a modifier for the particular state of the element; which we communicate by prefixing the class with `--`, making it `--hidden`. (Note, this a composable variant of [Block/Element/Modifier](http://getbem.com/introduction/)

zspencer (approved):
This looks great! I put in my persnickety Semantic HTML whiney-bits; but they're not blocking; more an attempt to bring our attention to holding it with a bit more intentionality.

convene-web/app/views/layouts/_header.html.erb
zspencer: Looks like we want to set an `aria-describedby` or `aria-labeledby` https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties-3

PR #164, opened 2020-12-13
Separate WaitingRoomPage from RoomSettingPage

user512:
Decompose the responsibility for interacting with the WaitingRoom from the RoomSettingsPage, resolve concern on https://github.com/zinc-collective/convene/pull/157/files#r533904802

### DIFF
--- a/features/page-objects/RoomSettingPage.js
+++ b/features/page-objects/RoomSettingPage.js
@@ -1,4 +1,5 @@
 const Page = require("./Page");
+const WaitingRoom = require ("./WaitingRoom")
 
 class RoomSettingPage extends Page {
   constructor(driver, room) {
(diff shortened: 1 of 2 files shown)


features/page-objects/RoomSettingPage.js
```diff
@@ -20,11 +21,8 @@ class RoomSettingPage extends Page {
...
 
   async enterAccessCode(accessCode) {
-    const accessCodeInput = await this.findByCss("[id='waiting_room_access_code']");
-    accessCodeInput.sendKeys(accessCode.value);
-
-    const submitInput = await this.findByCss("[type='submit']");
-    return submitInput.click();
+    const waitingRoom = new WaitingRoom(this.driver)
+    waitingRoom.submitAccessCode(accessCode)
   }
```
zspencer, line 26: Is there a way we can find the usages of enterAccessCode and replace it with the new `WaitingRoom` class;so that we don't have the responsibility still split between these functions?

zspencer (changes requested):
I like the [extract class](https://refactoring.guru/extract-class) refactoring and I think it's a great step! I would love it if we went a bit further and did an [inline method](https://refactoring.guru/inline-method) refactoring to remove the probability of the `enterAccessCode` function being used in more places?

zspencer (approved):
GLORIOUS!!!!!!

PR #166, opened 2020-12-16
Add link to updated personas document to convene-web README.

roseaboveit:
Looking for comments on language and placement. 
Currently I only added it to the convene-web README as it seems to be contributor facing more than explicitly for the public. 
I can easily add to the overall Convene README if desired or change the language as well.

I have also just sent the legal statement to operations so there shouldn't be any blockers on merging on that front.

### DIFF
--- a/convene-web/README.md
+++ b/convene-web/README.md
@@ -21,6 +21,11 @@ The central piece to Convene is `convene-web`, a Ruby on Rails server that is re
 * serving the Convene UI
 * managing users, workspaces, rooms, permissions, etc
 
+This [living excel sheet](https://docs.google.com/spreadsheets/d/1BOBCT0yrgrbCuQFTx_hIQak0FSQjnjjFZVA3YksEv8s/edit#gid=622652343)
+gives a high level view of the personas and segments we are focusing on initially in our design of
+Convene. It also includes our current vision of the types of workspaces, rooms and participants it
+serves and clarifies the design in privacy permissions.
+
 The Convene UI is based on Rails standard templating system, with heavy use of:
 * [Stimulus JS](https://stimulusjs.org/)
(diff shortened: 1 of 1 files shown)


convene-web/README.md
```diff
@@ -21,6 +21,11 @@ The central piece to Convene is `convene-web`, a Ruby on Rails server that is re
 * serving the Convene UI
 * managing users, workspaces, rooms, permissions, etc
 
+This [living excel sheet](https://docs.google.com/spreadsheets/d/1BOBCT0yrgrbCuQFTx_hIQak0FSQjnjjFZVA3YksEv8s/edit#gid=622652343)
+gives a high level view of the personas and segments we are focusing on initially in our design of
+Convene. It also includes our current vision of the types of workspaces, rooms and participants it
+serves and clarifies the design in privacy permissions.
```
zspencer, line 27: Not Blocking: I think from an information architecture perspective, this is a reasonable place to put this; and I think adding it to the top-level readme as a bullet would also make sense to replace the existing 'Personas' link (Which is more a link to an old scratch pad)

Not Blocking: From a "link design" perspective; Neilson Norman Group has some [great advice on how to label links](https://www.nngroup.com/articles/better-link-labels/); which makes me think that instead of linking the living excel sheet (implementation detail, possibly removable) we could instead link the "A high level view of our design..." section.

zspencer (approved):
Thank you! [I've issued you 20 Patronage Points](https://airtable.com/tblndyi9VR3shxtmS/viwxgldsOPhzLxfpV/recu6TZwWiQSekH7r?blocks=hide) and am merging this now!

Feel free to consider (or disregard) the not-blocking feedback I dropped; and make your own decision about whether it's worth applying!

Issue #160, opened 2020-12-08
Navigate Between Convene and Other Browser Tabs


hangtwenty:
## Original Motivation (by @hangtwenty)
> Subject: The page <title> doesn't have "Convene" in it
>
> The page <title> doesn't have "Convene" in it, so it was hard for me to switch-tab to it just now. (oh tabs)

## TODO
- [x] Define the intent behind the feature.
- [ ] Move the feature definition to the repository (https://github.com/zinc-collective/convene/issues/160)
- [ ] Implement Steps (Note: I don't know if this is possible with webdriver; so we may strike-this out at some point)
- [ ] Implement Feature 
   - [ ] Add a Favicon: [W3C Guidance](https://www.w3.org/2005/10/howto-favicon), [Cheat Sheet](https://github.com/audreyfeldroy/favicon-cheat-sheet)
   - [ ] Include the Brand Name, Space name and (when possible) Room name in the Title tag (https://github.com/zinc-collective/convene/issues/160)

## Design Notes
- Some browsers "bundle" tabs into just their favicon; which means that folks who have lots of tabs open will have a hard time picking Convene out if this is purely a textual change
- It's also not entirely clear to me how wide of text will show up on different browsers; but I _think_ it's "left-to-right" in languages that are read left-to-right; so we want to order the title as Room Name -> Space Name -> Brand name

hangtwenty:
I'll do a quick change

zspencer:
This is a great point! I've issued you [20 patronage points](https://airtable.com/tblndyi9VR3shxtmS/viwxgldsOPhzLxfpV/recxcU0CXO07kzj1S) for raising it to our attention! 

I'm going to do some editing of the top-level issue to reframe it around the problem-to-solve; and add some bullet-points for the things to think through or actions to take in order to resolve it!

zspencer:
Aight, I've updated it! Please keep in mind that we don't expect people to do _all the things_. Starting with just the text makes a ton of sense; but feel free to take a stab at making one if you want! Or check with @colombene if we have a logo that's suitable for rendering as a favicon.

zspencer:
I'm going to close this for now, especially as we need to think more about how Branding works across custom domains; etc.

PR #167, opened 2020-12-17
Signpost the current page as a Convene Space more clearly

zspencer:
See: https://github.com/zinc-collective/convene/issues/160

This builds on @hangtwenty's initial shift; while putting the responsibility into the existing helpers for page-titles; which should allow us to more clearly sign-post which Space & Room a particular Tab is. 

It also drops the Feature into the code-base; so that it's more clear that navigation between external-systems and Convene is something we care about; and may want to more refine with additional intentions.

### DIFF
--- a/convene-web/app/controllers/application_controller.rb
+++ b/convene-web/app/controllers/application_controller.rb
@@ -3,7 +3,11 @@ class ApplicationController < ActionController::Base
   # Override on a per-controller basis to display different title
   # @returns [String]
   helper_method def page_title
-    "ConveneWeb"
+    if current_workspace.present?
+      "Convene - #{current_workspace.name}"
+    else
+      'Convene - Video Meeting Spaces'
+    end
   end
 
   include Passwordless::ControllerHelpers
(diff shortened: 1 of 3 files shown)


zspencer:
Whichever of the @zinc-collective/convene-maintainers merges  this, please take care not to erase @hangtwenty's contribution. I don't quite get how the "Co-authored-by" gets auto-filled on squash merges. I think it pulls each authors email from their commit, but I've noticed it sometimes doesn't work quite the way I expect.

anaulin (approved):

anaulin:
I wasn't sure if during "Squash and merge" I was going to get the option to add co-authorship info, so out of an abundance of caution I made a "manual squash" to end up with one commit with the correct co-authorship info. But it looks like that also added me as a co-author? Which is not really what I intended. Sigh. Sorry.

I'm going to try to merge this now, and see if I can add the co-authorship info during the squash+merge step.

anaulin:
For anyone following along at home: "Squash and merge" did give me the ability to add or change the co-authorship information, and it had correctly already filled it in with what was in my commit (and it did not add me to the authors 😌 ). Here is the resulting merge commit: https://github.com/zinc-collective/convene/commit/8411ab240175f7bd951fabf7eb3a1cb6cf3f27e7

zspencer:
Thank you so much @anaulin!

Issue #118, opened 2020-09-11
🥔 `Space`: `Authentication`


zspencer:
People sign in to a Space by providing an Email Address. Convene confirms they are actually the person with access to that Email Address by sending an email with a link and one-time code that they provide back to us.

Unauthenticated people are _Visitors_, while Authenticated people are either _Guests_, _Neighbors_, or _Residents_.

- Visitors: Unauthenticated People.
- Guests: Authenticated People without a Membership in any Space in the Neighborhood.
- Neighbors: Authenticated People with a Membership in any (other) Space in the Neigbhorhood.
- Residents: Authenticated People with a Membership in the current Space.

In the future, we intend to explore more authentication methods (like, Apple Sign In/web sign in/etc); but not just yet.

## Design Assets
From @colombene - [Wireframes for authentication flow](https://xd.adobe.com/view/fd425dbe-5384-44c9-997a-eeee6e886a86-a811/)

## Use Cases
- [x] 🥔 Signing In by Magic Link
- [x] 🥔 https://github.com/zinc-collective/convene/issues/159
- [x] 🥔 Sign Out Manually
- [x] 🥔 Sign Out from Inactivity

### 1.1
- [ ] 🌫️ Sign in via SMS

## Bugs

- [x] https://github.com/zinc-collective/convene/issues/1809

user512:
Is this feature needed now member can click on the link to identify themselves?
This seems to be a useful feature for identify thru phone number instead.
```features/identification.feature
  Scenario: Identification via entering a Code that is Emailed
    Given I have submitted my email in the sign in form
    When I enter the code emailed to me 
    Then I am identified within the Space for a week
```

Now that we have https://github.com/zinc-collective/convene/pull/150 merged in, I think we just have to implement the avatar on nav bar and expose the logout link to fulfill the `Removing my identity from a device` scenario.

zspencer:
@user512 - I would be comfortable deferring the entering of the code to a different Milestone, if you like. However; I think it's an important Scenario to support; as someone may be logging in on a computer that does not have their email address; such as a shared library computer.

user512:
I see, I was previously confused by if one do not have email access, how can they enter the code from the email the app sent without realizing people can look at their mobile device.

zspencer:
@user512 - I have pulled that Scenario into https://github.com/zinc-collective/convene/issues/159 so that we don't have to worry about it just yet.

zspencer:
[...]

PR #170, opened 2020-12-20
Refactor navbar to semantic html

user512:
Connects: https://github.com/zinc-collective/convene/issues/118
Follow up on https://github.com/zinc-collective/convene/pull/163

Refactor navbar to semantic html and replace screenshot avatar image to svg provided by @colombene.

Screen Shot 2020-12-19 at 4 53 00 PM

### DIFF
--- a/convene-web/app/javascript/controllers/menu_controller.js
+++ b/convene-web/app/javascript/controllers/menu_controller.js
@@ -4,7 +4,7 @@ export default class extends Controller {
   static targets = [ "button", "items" ]
 
   toggle() {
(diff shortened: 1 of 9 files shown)


zspencer (approved):
Thank you! This is much nicer!

I do have one more tiny piece of feedback!

It looks like we have the following hierarchy:

- header
  - space-menu
     - profile-menu
     - brand

I think I was expecting something more like this:
- header
  - space-menu
  - profile-menu
  - brand
  
This feels like it reflects the information hierarchy that the _person_ is distinct from the _space_ they are in.

PR #171, opened 2021-01-03
Styled login form and modified login new & create template wording

user512:
* When implementing the ok button, we were wondering the intent of the button
because we can't redirect people to their own email client,
skipping ok button in the implementation for now.
* UI Mockup:
https://xd.adobe.com/view/fd425dbe-5384-44c9-997a-eeee6e886a86-a811/screen/b1223ba7-c6bf-4e8e-94ee-0ef583a178cb/
https://xd.adobe.com/view/fd425dbe-5384-44c9-997a-eeee6e886a86-a811/screen/456d873b-accd-4f06-89a0-a1dfeadee38c/

Co-authored-by: Ana Ulin <anaulin@users.noreply.github.com>
Co-authored-by: Whitney <roseaboveit@users.noreply.github.com>
Co-authored-by: Zee <zspencer@users.noreply.github.com>

Screen Shot 2021-01-09 at 3 47 25 PM
Screen Shot 2021-01-09 at 3 56 56 PM

### DIFF
--- a/convene-web/app/javascript/src/application.scss
+++ b/convene-web/app/javascript/src/application.scss
@@ -1,5 +1,9 @@
 h2 {
   @apply text-xl font-medium text-gray-900;
 }
+.center-container {
+  @apply flex flex-col items-center justify-center px-4 pt-28;
+}
+
 @import "./custom_components";
 @import "./page_bem_class";
--- a/convene-web/app/javascript/src/custom_components/button.scss
+++ b/convene-web/app/javascript/src/custom_components/button.scss
@@ -8,4 +8,10 @@
   .button-blue:hover {
     @apply bg-blue-700;
   }
+  .button-gray {
+    @apply bg-gray-500 text-white;
+  }
+  .button-gray:hover {
+    @apply bg-gray-700;
+  }
 }
--- a/convene-web/app/javascript/src/custom_components/form.scss
+++ b/convene-web/app/javascript/src/custom_components/form.scss
@@ -76,3 +76,16 @@
 .access-code-form__submit {
   @apply mt-4;
 }
+
+.identification-form {
+  @apply w-full;
+  @screen sm {
+    @apply w-1/2;
+  }
+}
+.identification-form__submit {
+  @apply mt-6;
+  @screen sm {
+    @apply w-2/5 mx-auto
+  }
+}
--- a/convene-web/config/locales/en.yml
+++ b/convene-web/config/locales/en.yml
@@ -48,10 +48,11 @@ en:
     sessions:
       create:
         session_expired: 'Your session has expired, please sign in again.'
-        email_sent_if_record_found: "We've sent you an email."
+        email_sent_if_record_found: "Check your email for a link to log in automatically."
         token_claimed: "This link has already been used, try requesting the link again"
       new:
-        submit: 'Send magic link'
+        email_label: 'Your Email'
+        submit: 'Submit'
     mailer:
       subject: "Your magic link ✨"
       magic_link: "Here's your link: %{link}"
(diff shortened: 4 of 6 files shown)


anaulin:
Recording decision from conversation with @colombene during our weekly meeting: it is fine to _not_ have the "OK" button on the "Check your email for a login link" screen, and instead just try to force the user to go to their email.

And it would be nice for the login link to re-direct the user to their workspace after being logged in. (For bonus points, we could maybe track where they were coming from when they started the login process and redirect to that, but it seems that redirecting to workspace gets us 98% of the way there and is easier.)

convene-web/app/javascript/src/custom_components/form.scss
```diff
@@ -75,4 +75,17 @@
...
   @apply mt-4;
+}
+
+.identification-form {
+  @apply w-full;
+  @screen sm {
+    @apply w-1/2;
+  }
+}
+.identification-form__submit {
```
zspencer, line 86: Not Blocking: I really need to write out what I consider "Semantic, Composable BEM" which is distinct from regular BEM in the following ways:

1. When we _can_ rely on targeting native HTML elements, such as `input[type="submit"]`, we do.
2. When we are targeting sub-components, we use CSS cascading, not `__`. This allows us to have some guarantee that we can override application level styling at the  component or sub-component level.

In this particular case, what this would mean is we would refactor to:

```scss
.identification-form { 
  input[type="submit"] {
  
  }
}
```

convene-web/config/locales/en.yml
```diff
@@ -48,10 +48,11 @@ en:
     sessions:
       create:
         session_expired: 'Your session has expired, please sign in again.'
-        email_sent_if_record_found: "We've sent you an email."
+        email_sent_if_record_found: "Check your email for a link to log in automatically."
         token_claimed: "This link has already been used, try requesting the link again"
       new:
-        submit: 'Send magic link'
+        email_label: 'Your Email'
+        submit: 'Submit'
```
zspencer, line 55: Few things make me happier than using I18n for _every single_ piece of text we show people <3 <3 <3

convene-web/app/javascript/src/custom_components/button.scss
```diff
@@ -8,4 +8,10 @@
   .button-blue:hover {
     @apply bg-blue-700;
   }
+  .button-gray {
+    @apply bg-gray-500 text-white;
+  }
+  .button-gray:hover {
+    @apply bg-gray-700;
+  }
```
zspencer, line 16: Not blocking: I am curious what the semantics of these buttons are; and hope

Generally, there are ~4 different "actions" where pressing a button:

1. Changes nothing, and is purely for acknowledging information; like dismissing an alert.
2. Moves someone forward (or backward) in a workflow
3. Changes something safely (i.e. changes a value that can be easily changed back).
4. Changes something _dangerously_

I would prefer we use the `--modifier` syntax for denoting these things; i.e. `button --acknowledge` or `button --nav`  or `button --commit` or `button --commit-danger` 

This way when we build features, we're consistently signaling to the user what they can expect to happen when they push a button.

convene-web/app/javascript/src/application.scss
```diff
@@ -1,5 +1,9 @@
 h2 {
   @apply text-xl font-medium text-gray-900;
 }
+.center-container {
+  @apply flex flex-col items-center justify-center px-4 pt-28;
```
zspencer, line 5: This feels like a composite utility class; which is probably quite useful; but I would prefer we used the `--modifier` syntax for naming. Perhaps this is `.--centers-contents-vertically`

zspencer (approved):
This looks lovely!! Thank you!

PR #172, opened 2021-01-03
Avoid eager loading in test

user512:
I tried a few things to avoid eager loading:
1. Switch to classic loader instead of the default Rails 6 [Zeitwerk](https://github.com/fxn/zeitwerk#file-structure)
  * It works but classic loader will be deprecated eventually, not recommended.
2. Rename `Admin::ApplicationController` to `Admin::BaseController`
  * I wasn't 100% why it works but my guess is [Zeitwerk only made one autoload look up](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#less-file-lookups) and
    the first one it found was `Admin::ApplicationController`, and raise the superclass mismatch error.
  * There might be a way to config Zeitwerk to do a smarter look up but I think renaming
    `Admin::ApplicationController` to `Admin::BaseController` makes sense, and we can treat
    `ApplicationController` as a Rails keyword instead.

### DIFF
--- a/convene-web/app/controllers/admin/base_controller.rb
+++ b/convene-web/app/controllers/admin/base_controller.rb
@@ -5,7 +5,7 @@
 # If you want to add pagination or other controller-level concerns,
 # you're free to overwrite the RESTful controller actions.
 module Admin
-  class ApplicationController < Administrate::ApplicationController
+  class BaseController < Administrate::ApplicationController
     include Configuration::Configurable
 
     http_basic_authenticate_with configuration.basic_auth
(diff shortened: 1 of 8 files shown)


anaulin (approved):
Ah, nice. Thanks for researching this and cleaning it up, @user512 !

zspencer:
Great sleuthing! I'm a bit worried that we'll run into something similar with the `Furniture::Tables::Controller` name in the future. I don't think we need to deal with it now though; but glad to know where to track down what goes sideways! Thanks Tom!

PR #173, opened 2021-01-10
Global color update

user512:
See: https://github.com/zinc-collective/convene/issues/23

Screen Shot 2021-01-09 at 4 52 59 PM
Screen Shot 2021-01-09 at 4 53 05 PM
Screen Shot 2021-01-09 at 4 53 23 PM

Co-authored-by: Whitney <roseaboveit@users.noreply.github.com>

### DIFF
--- a/convene-web/app/javascript/src/custom_components/header.scss
+++ b/convene-web/app/javascript/src/custom_components/header.scss
@@ -1,7 +1,7 @@
 @layer components {
   .header {
     .space-menu {
-      @apply flex items-center justify-between px-4 py-3 bg-purple-900 text-white;
+      @apply flex items-center justify-between px-4 py-3 bg-primary-900 text-white;
 
       h1 {
         @apply text-2xl font-bold flex-grow;
@@ -13,7 +13,7 @@
           @apply font-hairline text-center text-xs;
         }
         nav {
-          @apply absolute bg-purple-600 px-4 py-1 shadow;
+          @apply absolute bg-primary-600 px-4 py-1 shadow;
         }
       }
 
--- a/convene-web/app/javascript/src/custom_components/header.scss
+++ b/convene-web/app/javascript/src/custom_components/header.scss
@@ -1,7 +1,7 @@
 @layer components {
   .header {
     .space-menu {
-      @apply flex items-center justify-between px-4 py-3 bg-blue-500 text-white;
+      @apply flex items-center justify-between px-4 py-3 bg-purple-900 text-white;
 
       h1 {
         @apply text-2xl font-bold flex-grow;
@@ -13,7 +13,7 @@
           @apply font-hairline text-center text-xs;
         }
         nav {
-          @apply absolute bg-blue-600 px-4 py-1 shadow;
+          @apply absolute bg-purple-600 px-4 py-1 shadow;
         }
       }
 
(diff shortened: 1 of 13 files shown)


convene-web/app/views/waiting_rooms/show.html.erb
```diff
@@ -1,6 +1,6 @@
-<div class="bg-blue-100 min-h-screen flex items-center justify-between">
```
zspencer, line 1: I'm going to create a ticket for refactoring this view to focus the DOM onto the intent and remove extraneous classe (where possible).

In particular, I want the DOM to look something like:

<div class="waiting-room">
  <div class="room-info">
  
  </div>
  <form>
    <div class="access-code-field --with-errors"> 
      <label>Access Code</label>
      <input type="password" name="waiting_room[access_code]" />
      <p class="input-error">Error Message.</p>
      <p class="input-error">Second Error Message.</p>
    </div>
    <input type="submit" />
  </form>
</div>

convene-web/app/javascript/src/custom_components/header.scss
```diff
@@ -1,7 +1,7 @@
 @layer components {
   .header {
     .space-menu {
-      @apply flex items-center justify-between px-4 py-3 bg-blue-500 text-white;
+      @apply flex items-center justify-between px-4 py-3 bg-purple-900 text-white;
```
zspencer, line 4: Not Blocking: We may want to pull the coloring into something more abstract, so that if we do another re-color we don't have to perform shotgun surgery

For example: `bg-primary-900` could be an alias of `bg-purple-900` and if we wanted to switch to cyan, we could change `bg-primary-900` to point to `bg-cyan-900` instead of `bg-purple-900`.

My preference is we identify 3~6 hues and give them formal names that indicate their use; and we choose which color to use in which context. I would probably want @colombene to define these, since she's doing the majority of our UI work.

For example, I could see us using `dominant` or `primary` to describe the hue we use most prominently throughout the design, and `complement` or `accent` to describe the color that gives contrast, and `supplement` to describe a color that we use in cases where we want something that isn't high-contrast but also isn't the same as the dominent hue.

What I'm hoping is that the words in our styling code to be tied relatively closely [to the application of color theory](https://uxplanet.org/color-glossary-for-designers-terms-and-definitions-1a65549804bd), so that it acts as a "signpost" that helps guide folks learning CSS _from_ design (or learning design _from_ css)

zspencer (approved):
I'm really torn, I think it's probably fine to merge this, but I also would love if we invested a bit of time into picking name(s) for our colors that indicate their purpose in our visual language; and used that in place of `gray` and `purple` and `blue` throughout the project.

convene-web/app/javascript/src/custom_components/header.scss
user512: Yes, we can also aggregate all these color in `tailwind.config.js`.
https://tailwindcss.com/docs/customizing-colors#naming-your-colors
@roseaboveit and I talked about extracting these but we were sure the exact color code so we put purple here for now.
We can bring up the need for identifying a few color and name those as `neutral`, `info`, `alert` or `primary` and have all those color scheme store in this config js file for easy global edit.

There might be a need to implement custom theme for different workspace but I think we can worry about that later.

user512:
I don't think there is an urgent need to merge this. We can sync up with @colombene to get a sense of the color scheme and think of a better way to code out our color theme instead of using color name.

Issue #177, opened 2021-01-16
Upgrade Tailwind and TailwindUI


zspencer:
Tailwind just hit 2.0! We probably wanna get in on that for all it's goodness.

user512:
upgraded tailwind to v3 in https://github.com/zinc-collective/convene/pull/458

PR #176, opened 2021-01-16
Sprout guide to locking unlocking rooms

zspencer:
See: https://github.com/zinc-collective/convene/issues/40

- We can now use `.md` and `.markdown` files as erb partials.
- There's now a "breadcrumb" component for allowing folks to go back.
- The Feature model is starting to hold more responsibility.
Screen Shot 2021-01-15 at 4 33 52 PM
Screen Shot 2021-01-15 at 4 32 24 PM

PR #180, opened 2021-01-17
Align our SCSS organization with Tailwinds

zspencer:
Tailwind appears to organize their classes into:
- base (for "application wide" styles)
- utilities (for composable classes)
- components (for classes that are not particularly composable)

If we load Tailwind _first_, then use the @layer directive, it ensures
that order-of-rule-definition plays nicely, so we can override tailwinds
defaults for `h1` with our own.

See: https://github.com/zinc-collective/convene/issues/177


anaulin (approved):
In terms of code organization, this makes sense to me and looks fairly clean. (No idea if this is changing any of the "look", but I'm sure we'll continue iterating on this regardless.)

convene-web/app/javascript/src/components/guide_summary.scss
```diff
@@ -0,0 +1,12 @@
...
+    @apply card;
+    header {
+      @apply card__header;
+    }
+
+    footer {
+      @apply card__footer;
+    }
+  }
+}
```
zspencer, line 12: This is probably extraneous until  #176  gets merged :X.

convene-web/app/javascript/src/utilities/card.scss
```diff
@@ -0,0 +1,12 @@
...
+  .card {
+    @apply grid grid-cols-1 grid-rows-2 grid-flow-col shadow rounded-lg px-3 bg-white;
+  }
+  .card__header {
+    @apply w-full items-center;
+  }
+
+  .card__footer {
+    @apply w-full flex flex-col items-center justify-center border-t border-neutral-200;
+  }
```
zspencer, line 11: For utilities, I'm thinking we'll probably want to go "real BEM" instead of "Composable BEM" because we can't `@apply .card .header` to mix in the utilities.

user512 (approved):
Just one concern on recompile performance and I like the new structure!

convene-web/app/javascript/src/application.scss
```diff
@@ -1,17 +1,9 @@
...
-}
-h2 {
-  @apply text-xl font-medium text-neutral-900;
-}
-.center-container {
-  @apply flex flex-col items-center justify-center px-4 pt-28;
-}
+@import "tailwindcss/base";
+@import "tailwindcss/components";
+@import "tailwindcss/utilities";
```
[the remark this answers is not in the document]
zspencer: I think it should be safe to pull it out. My read of the docs indicate that using the @layer directive is supposed to allow webpacker/tailwind to only rebuild the layer that changes; but I’m not entirely sure how to prove it.

convene-web/postcss.config.js
```diff
@@ -1,8 +1,8 @@
 module.exports = {
   plugins: [
-    require('autoprefixer'),
     require('postcss-import'),
     require('tailwindcss'),
+    require('autoprefixer'),
```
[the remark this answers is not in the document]
zspencer: Not sure, it’s the order they were in the tailwind installation docs.

convene-web/app/javascript/src/utilities/card.scss
user512: I like this!

Issue #182, opened 2021-01-17
bundle install fails at Puma when it tries installing with 4.3.5


user512:
Failed to run bundle install when setting up environment with @churusaa, and I think we came across this issue: https://github.com/puma/puma/issues/2342
This issue is fixed in 4.3.6 and @churusaa and I were able to continue with setup after running `bundle update puma`.

Our Gemfile.lock specify `4.3.5` when this happen. 
Ref: https://github.com/zinc-collective/convene/blob/9df04278b90a15bf360f84aa5d5fb8a73f964cbd/convene-web/Gemfile.lock#L158

PR #183, opened 2021-01-17
Avoid puma 4.3.5 that avoids bundle install

user512:
Closes: https://github.com/zinc-collective/convene/issues/182
Ref: https://github.com/puma/puma/issues/2342


zspencer (approved):
Thanks @user512! This is super helpful!! Do we want to lock to ~> 5.X just so we don’t have to worry about breaking changes on major version bumps? 

My gut says that because we’re not doing anything with puma directly, it’s unlikely to need to be locked down explicitly until we start writing puma specific stuff.

Issue #186, opened 2021-01-23
`Gizmo`: Payment Form


anaulin:
Motivated by potential customer / interested party who have the following need:
1. Have the Investors log in to their bank account using Plaid's Auth Service.
2. Use that to pull their checking/savings details.
3. Ideally, pull that data out as a NACHA file from Plaid, so that they can then upload manually into a banking provider for ACH.

### TODO

- [x] Write Feature Definition (with supplementary links and such)
- [x] Shift the current VideoBridge to a piece of furniture in all Rooms (So we can have Rooms without a Video Bridge) (https://github.com/zinc-collective/convene/pull/191)
- [x] Make Plaid Utility
  - [x] ~Add an API Version drop-down to the Plaid Utility Configuration~ This is tied to gem version; so we can't reall swap between them.
  - [x] Add Validations to the Plaid  Utility Configuration
- [x] Spike using Plaid Link to Connect an Account and Auth to retrieve Account Numbers
  - Auth Docs: https://plaid.com/docs/api/products/#auth
  - Link Docs: https://plaid.com/docs/link/
  - https://fin.plaid.com/articles/inside-ach-payments-with-stripe-and-plaid/
  - How Do?
    - [x] Make a Form and Active Record object that stores a "Check" (Name of Payee, Date, Amount, eSignature, Plaid Account Link ID)
    - [x] Set up Link to get the Plaid Account ID
    - [x] Set up Auth to retrieve the Account Number and Routing Number
    - [x] Expose all the check(s) to Space Members
    - [x] ~~Expose just the check(s) the person deposited to non Space Members~~
- [ ] Payment Form shouldn't explode the Space when Plaid isn't setup.
- [ ] Allow a `FurniturePlacement` to have many `:utility_hookups` so the `FurniturePlacement` knows which utilities to use.
- [ ] Spike building a NACHA file (Perhaps using [the ach gem?](https://github.com/jm81/ach)
  - Some docs on NACHA: https://www.mercbank.com/pdfs/Standard%20NACHA%20ACH%20File%20Format.pdf

- [x] Is Stripe A no-go? https://stripe.com/restricted-businesses

Notes from audit 12/15/2021
   - [x] add validations to form
   - [x] reframe to "payment" and away from "check" terminology
   - [ ] add some kind of help text that clarifies that this is a payment authorization, and the payment won't happen immediately
   - [ ] add input mask for "Amount" field, so that it looks like dollars
   - [ ] color on hover on the "Create" button looks weird
[...]

PR #193, opened 2021-02-06
Initial draft of feature file for check drop off feature

zspencer:


user512 (approved):
I think this is mergeable.

PR #195, opened 2021-02-07
Start Invitation logic

cjoulain:

TODO:

- [x] Add Authorization to ensure that only Space Owners can create Invitations to a Space
- [x] Build out the Invitation Form
- [x] Create an Invitation to the Space in the Invitation Controller
- [x] Feature flag Invitations so we can merge this!

### Maybe?
- [ ] Beautify the Sent Invitations List
- [ ] Rename SpaceInvitationsMailer to InvitationMailer to match the Model and Controller
- [ ] Don't swallow the input when failing to save an invitation
- [x] Add a `belongs_to :space, inverse_of: :invitations`. See [the general belongs_to guide](https://guides.rubyonrails.org/association_basics.html#belongs-to-association-reference) and the [inverse_of option](https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-inverse-of)
- [ ] Move Invitation email and movement from unsent to pending to a background job
- [ ] Flesh out the user journey for accepting an invitation from the link (especially for guests)


zspencer:
Thank you @eg-bet and @KellyAH ! I've allocated 80 points of Patronage each:

Egbet, this places your patronage balance at 480, or 0.67% of outstanding patronage: https://airtable.com/tblndyi9VR3shxtmS/viwxgldsOPhzLxfpV/recYSFdgWTTJongTc?blocks=hide


Kelly, this places your patronage balance at 640, or 0.89% of outstanding patronage: https://airtable.com/tblndyi9VR3shxtmS/viwxgldsOPhzLxfpV/recC50u4lQN01Ojjl?blocks=hide

Thank you so much for your regular contributions to Convene! 

Note: As CJ is a maintainer, their patronage balance is updated via submitting time-sheets.

convene-web/app/views/spaces/edit.html.erb
```diff
@@ -0,0 +1,12 @@
...
+<% end %>
+
+<%- @space.members.each do |member| %>
+  <%= member.name %>
+  <%= member.email %>
+<%- end %>
+
+<%= form_with model: @space.space_memberships.new do |member_form| %>
+
+<% end %>
```
zspencer, line 12: Before we merge, we will probably want to change this to use the new `@space.invitations` relationship instead of the `@space.space_memberships` relationship.

convene-web/config/routes.rb
```diff
@@ -22,6 +22,9 @@
     end
   end
 
+  resources :space_memberships do
```
zspencer, line 25: Before we merge, we will probably want to rename this to `:invitations` and restrict it to only the`create` and destroy actions.

See an example of a resource being restricted to only particular actions here: https://guides.rubyonrails.org/routing.html#configuring-the-rails-router

convene-web/app/models/invitation.rb
```diff
@@ -0,0 +1,2 @@
+class Invitation < ApplicationRecord
+end
```
zspencer, line 2: Not blocking:

We may want to communicate what the "Job" of the invitation is by adding some comments.

Things that are useful in this kind of documentation are:

1. What other parts of the system does an Invitation interact with and why?
2. What Feature files are particularly relevant for this model?


Here's an example, written in [YARDoc syntax](https://yardoc.org/)

```
# An {Invitation} offers a {Person} the opportunity to join a {Space}. When an {Invitation} is accepted, it creates a {SpaceMembership}. 
# For more information, review the {file:features/spaces/invitation.feature Space Invitation Feature Definition.}

File: convene-web/app/models/invitation.rb [hunk 4]
```diff
@@ -0,0 +1,2 @@
+class Invitation < ApplicationRecord
```
zspencer, line 1: We may want to add a `belongs_to :space, inverse_of: :invitations`. See [the general belongs_to guide](https://guides.rubyonrails.org/association_basics.html#belongs-to-association-reference) and the [inverse_of option](https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-inverse-of)

convene-web/app/controllers/spaces_controller.rb
```diff
@@ -2,4 +2,8 @@ class SpacesController < ApplicationController
   def show
     @space = current_space
   end
+
+  def edit
+    @space = current_space
+  end
```
zspencer, line 8: Not blocking: Since we're using `space` in both `show` and `edit` we could expose a helper method instead that lazily retrieves the space, removing that responsibility from the show and edit actions:

```
helper_method def space
  @space ||= current_space
end
```

And replace the references to `@space` in the view  with `space`. 

As an aside,My personal preference is to leverage [decent_exposure](https://github.com/hashrocket/decent_exposure) for a consistent pattern for retrieval and building of objects for controllers.

convene-web/spec/requests/spaces/invitations_request_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe "/spaces/:space_id/invitations", type: :request do
+  it "doesn't expsose" do
```
zspencer, line 4: Maybe blocking: This `it` statement probably wants a better wording. Perhaps `it "creates an invitation to the Space"`

convene-web/spec/spec_helper.rb
```diff
@@ -43,7 +43,9 @@
   # inherited by the metadata hash of host groups and examples, rather than
   # triggering implicit auto-inclusion in groups with matching metadata.
   config.shared_context_metadata_behavior = :apply_to_host_groups
-
+  config.define_derived_metadata do |meta|
```
zspencer, line 46: Not Blocking: We may want to add a comment explaining our reasoning for why we prefer to `aggregate_failures`.

I, personally, prefer that we `aggregate_failures` because without it, we can wind up in situations where we don't have a clear picture of what parts of the test are failing; which can make it hard to orient our understanding as we decide what actions to take.

zspencer (commented):
This is looking great! 

I've added the todos from todays session as checkboxes in the pull request body so that we have an idea of where we may want to start our next session! (Or if someone wants to take a stab at them between now and then they can do so!)

convene-web/app/controllers/invitations_controller.rb
```diff
@@ -0,0 +1,13 @@
+class InvitationsController < ApplicationController
+
```
cjoulain, line 2: Add authorization to only allow members to invite and remove To Do note since create action exists

convene-web/app/controllers/invitations_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  # TODO: fix create action to create invitation data so spec test passes
+  def create
+    @space_invitation = current_space.invitations.new(invitation_params)
```
cjoulain, line 5: Remove instance variable and rename variable to reference controller name for clarity

convene-web/app/controllers/invitations_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+    @space_invitation.save!
```
cjoulain, line 6: To Do: what happens if save action fails?

convene-web/app/views/spaces/edit.html.erb
```diff
@@ -0,0 +1,12 @@
...
+<%= form.text_field :name %>
+<% end %>
+
+<%- @space.members.each do |member| %>
+  <%= member.name %>
+  <%= member.email %>
+<%- end %>
+
+<%= form_with model: @space.invitations.new do |member_form| %>
+
```
cjoulain, line 11: Add form fields for name and email

convene-web/spec/requests/spaces/invitations_request_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe "/spaces/:space_id/invitations", type: :request do
+  it "doesn't expose" do
```
cjoulain, line 4: Rename for more descriptive action

convene-web/spec/requests/spaces/invitations_request_spec.rb
```diff
@@ -1,12 +1,35 @@
...
+
+  # ------ #
+  # NOTES
+  # ------ #
+  # Kelly H, Zee, Egbet decided these are the possible invitation statuses:
+  # pending - email was not sent yet due to mailer issues etc.
+  # sent - email was sent but not accepted
+  # accepted - receiver accepted the invitation
+  # rejected - receivers rejected the invitation
+
```
eg-bet, line 32: Have been thinking about adding an additional invitation status of `resent` that logs the amount of times resent, default being `0`. I hear Zee in the back of my head saying this may be getting ahead of myself, but with a query like `space.invitations.find_by(name: 'foobar', email: 'foobar@example.com')` we can run into an issue where an invite is sent to the same name/email more than once and this query not working. I think we may have actually seen this happen in our last session where there were two invitation records of the same name/email and we saw an error relating to expecting to find just one record.

zspencer: This is an interesting corner case to think through!

There are a number of paths we could go:

1. Ignore it for now (depending on what we anticipate the impact is)
2. Use a counter
3. Use an Event pattern to track when we send invitations; so that we don't lose context and can make more informed decisions (either programmatically or as a human)
4. ???

KellyAH: I like 1. + adding it as a todo or notes somewhere for now. and 3 when we tackle eventually it.

cjoulain:
I updated a few files to get the CI to pass, as we were getting a failure related to the invitations_requests spec (SpaceInvitationMailer was not defined).

I don't know if what I did was correct so feel free to comment/improve. 

One thing I don't understand is with the mail variable, it's listed as anonymous when using binding.pry and doesn't carry any data with it. I'm guessing that's because it's a stub?

```
[1] pry(#<RSpec::ExampleGroups::SpacesSpaceIdInvitations>)> mail
=> #<Double (anonymous)>

```
I also tried using ActionMailer instead of Double to send out a test message but it was also anonymous. 
```
=> #<InstanceDouble(ActionMailer::MessageDelivery) (anonymous)>
```

Do we want to add Mailcatcher settings to the docker.yml? Or is that something to add further down the line?

convene-web/spec/requests/spaces/invitations_request_spec.rb
```diff
@@ -16,9 +16,9 @@
 
     # TODO finish building logic to get below assertions to pass
     # generate mailer and views
-    expect(SpaceInvitationMailer).to have_received(:space_invitation_email).with(invitation)
-    expect(mail).to have_received(:deliver_now)
-    expect(invitation.status).to eq("sent")
+    expect(SpaceInvitationMailer).to have_received(:space_invitation_email)
+    expect(mail).to have_received(:deliver_later)
+    expect(invitation.status).to eq("pending")
```
zspencer, line 21: @cjoulain - I'm curious about the removal of the `.with(invitation)` constraint in the assertion? It feels useful to confirm that the Mailer was given the correct invitation to deliver.

I'm also curious about the change from confirming the Invitation status was set to 'sent' to 'pending'; as it seems like we would want the invitation to be sent upon creation in this particular case. 

Love the change to expecting this to be delivered later; I'm not entirely sure we have ActiveJob configured yet; so we may want to create a ticket in the Contributor Experience milestone for setting up background jobs; otherwise it's possible these emails won't deliver in production

convene-web/app/controllers/invitations_controller.rb
```diff
@@ -3,8 +3,12 @@ class InvitationsController < ApplicationController
   # TODO: fix create action to create invitation data so spec test passes
   def create
     @space_invitation = current_space.invitations.new(invitation_params)
-    @space_invitation.save!
-    redirect_to [current_space, @space_invitation]
+    if @space_invitation.save!
+      SpaceInvitationMailer.space_invitation_email(@person).deliver_later
```
zspencer, line 7: Could you clarify where you expect `@person`  to be set and/or where it came from? I'm looking through the "flow" of the code; and it seems like it's going to be a `nil` value.

Also, clarifying point: `save!` will return either `true` or raise an exception; so we will never reach the `else` branch. A more "railsy" way would be to use `save` without the bang; and then render the `new` action so any validation errors are presented back to the user to be corrected.

zspencer:
> I updated a few files to get the CI to pass, as we were getting a failure related to the invitations_requests spec (SpaceInvitationMailer was not defined).
> 
> I don't know if what I did was correct so feel free to comment/improve.


This is great! Thank you! I left a few points of clarification, but I think this is on the right track!

> One thing I don't understand is with the mail variable, it's listed as anonymous when using binding.pry and doesn't carry any data with it. I'm guessing that's because it's a stub?
> 
> [1] pry(#<RSpec::ExampleGroups::SpacesSpaceIdInvitations>)> mail
=> #<Double (anonymous)>
> 
> I also tried using ActionMailer instead of Double to send out a test message but it was also anonymous.
> 
> => #<InstanceDouble(ActionMailer::MessageDelivery) (anonymous)>

Yep, when we use `double` or `stub` within a test context; the values that are returned from those functions are instances of the class `Double`  and when we use `instance_double` they are of the class `InstanceDouble`. 

That said, if you only include the output of your expressions, it's very difficult for me to know what the context you're evaluating those expressions within. And in programming; context is _everything_; because we cannot make any assumptions about what is true or not without knowing exactly where we are in the call-stack and what data is within scope.

> Do we want to add Mailcatcher settings to the docker.yml? Or is that something to add further down the line?

We currently have [Maildev enabled in our Procfile](https://github.com/zinc-collective/convene/blob/development/Procfile#L7). We could move it to Docker, but I don't know what benefit that provides; and would mean that anyone who wanted to contribute and run the tests would also need to have docker configured for their machine.

You can see the JavaScript code we use to interact with the Maildev API in [features/lib/MailServer.js](https://github.com/zinc-collective/convene/blob/development/features/lib/MailServer.js). 

That said, for our rspec tests, we don't need to deliver the emails to Maildev, and can leverage a library like [mailspec](https://github.com/email-spec/email-spec#rspec)

convene-web/app/views/spaces/edit.html.erb
zspencer: We did it! YAYYYY!

convene-web/app/models/invitation.rb
```diff
@@ -0,0 +1,28 @@
...
+  belongs_to :space
+
+  belongs_to :invitor, class_name: :Person
```
zspencer, line 4: We may also want to add a `has_many :invitations` on the Person class to indicate who they have invited

convene-web/spec/requests/spaces/invitations/rsvp_spec.rb
```diff
@@ -0,0 +1,27 @@
+RSpec.describe "/spaces/:space_id/invitations/:invitation_id/rsvp", type: :request do
+  describe "GET /spaces/:space_id/invitations/:invitation_id/rsvp" do
+    context 'as a guest' do
+    end
+
+    context 'as a neighbor' do
+    end
+
+    context 'as a space member' do
+    end
```
zspencer, line 10: Probably need to add a `@todo` to flesh out this more fully so that someone else can grab it after we merge.

features/spaces/invitations.feature
```diff
@@ -0,0 +1,61 @@
+Feature: Spaces: Invitations
+  Invitations allow Space Owners to bring additional people into the Space
+
+  @unstarted @andromeda
```
zspencer, line 4: This should be flipped to `@built @unimplemented-steps`

convene-web/spec/requests/spaces/room/furniture_placements_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+
+      it 'allows updating a placement' do
+        patch placement_path, params: { furniture_placement: { furniture_attributes: { content: 'updated content' }} }
+        expect(response).to have_http_status(302)
+        expect(placement.reload.settings['content']).to eq('updated content')
+      end
+    end
+  end
+end
```
zspencer, line 32: We should probably figure out why that is in this branch, is it in `development` already?

convene-web/app/controllers/invitations_controller.rb
zspencer: We did it!

zspencer (approved):
This looks like it is good enough to merge, especially since it's behind a feature flag!

PR #196, opened 2021-02-07
Fix review apps

zspencer:
There is probably a smarter way to do this, but what I think is
happening is because `db:prepare` is being called in the same process
that is doing the loading of the data, the models haven't had their
underlying mapping to the database tables "refreshed" which is causing
the builds to fail.


zspencer:
Screen Shot 2021-02-07 at 1 22 31 PM

IT WORKS!

user512 (approved):

user512:
Make sense!

PR #197, opened 2021-02-07
Upgrade Ruby to 2.7.2 and Rails to 6.1

zspencer:
This should be a safe upgrade, given how little we have and how good our
test coverage is


convene-web/Gemfile
```diff
@@ -10,14 +10,13 @@ ruby '2.7.1'
 gem 'dotenv-rails', groups: [:development, :test]
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 6'
+gem 'rails', '~> 6.1'
```
user512, line 13: I think we can keep `gem 'rails', '~> 6'` and just run `bundle update rails` to achieve the same result and avoid locking to `6.1`.

user512 (approved):
LGTM

convene-web/Gemfile
zspencer: What does that buy us?

PR #198, opened 2021-02-14
Expire sessions after seven days

user512:
Connects: https://github.com/zinc-collective/convene/issues/118

We haven't figured out how to expire a session due to inactivity,
so we just expire the session after seven days regardless.


anaulin (approved):
✅  🚢

PR #202, opened 2021-02-18
Sprout Utility/UtilityConnection for connecting a Space to an external Service

zspencer:
This is very rough, but the goal here is to make it so the Plaid (or possibly DocuSign) configuration can be set on a per-space basis for #186; as well as open up opportunities for folks to connect their existing infrastructure for furniture that leverages it.


convene-web/app/lib/utilities.rb
```diff
@@ -0,0 +1,25 @@
+# Utilities are External Services that provide functionality to a piece of
```
anaulin, line 1: This comment immediately makes me think "Should we call this an `ExternalService` then?". Naming is hard, and I am myself a fan of using "Utility" when I can think of anything better, but I also recognize it is kinda like calling something "Thingie" -- not a huge value add. 😹

convene-web/app/lib/utilities.rb
```diff
@@ -0,0 +1,25 @@
...
+# - A Jitsi-Meet or BigBlueButton instance for videoconferencing
+# - An Airtable Base or Google Apps account that acts as a human-editable
+#   backend-data store for a piece of Furniture.
+# - A Dropbox, OneCloud or Google Drive account for file sharing.
+# - ???
+#
+module Utilities
+  # @return [Utility] configured as appropriate given the connection
+  def self.from_connection(connection)
+    Utility.new(settings: connection)
```
anaulin, line 17: Sounds like the "connection" is actually a "configuration" or "settings" (as opposed to a server-to-server connection, which is what I would have assumed from the name).

convene-web/app/models/utility_connection.rb
```diff
@@ -0,0 +1,24 @@
+# Links a {Utilities::Utility} to a {Space} or {Person}.
+class UtilityConnection < ApplicationRecord
```
anaulin, line 2: Ok, this is silly, but hear me out: if it is a "utility" in the gas, water, and electricity sense, then this is class is really a "hook up". (I told you this was silly.)

convene-web/app/models/utility_connection.rb
```diff
@@ -0,0 +1,24 @@
...
+  belongs_to :space
+
+  # Human-friendly name for disambiguating when a single kind of Utility has
+  # multiple Connections.
+  attribute :name
+  attribute :description
+
+  # Which type of {Utility} is connected, for example Airtable or  Asana or
+  # Jitsi-Meet or BBB or Plaid or Stripe.
+  attribute :utility_type
```
anaulin, line 13: Noodling out loud: would this eventually be an enum with a registry?

anaulin (commented):
This structure makes sense to me so far. I left some thinking-out-loud comments that are around naming, trying to make things a little bit more self-explanatory to a future reader of the code.

convene-web/app/models/utility_connection.rb
zspencer: No that's amazing I love it.

convene-web/app/models/utility_connection.rb
zspencer: Yea, I wonder if there's a Registry class provided by Rails  that we could use for this...

convene-web/app/lib/utilities.rb
zspencer: Yea, that's a good point

convene-web/app/models/utility_connection.rb
zspencer: There is not a registry class provide by Rails for this :(

Issue #206, opened 2021-02-21
Contributor Experience: Update bin/run script to let people know to open localhost:3000/spaces/system-test


zspencer:
Recommended by @KellyAH during today’s mobbing session, it’s not quite clear where to go to start futzing with Convene as a person!

user512:
@KellyAH, @anaulin added some message to guide developer to go to localhost:3000, let us know if this resolve this issue.

zspencer:
`bin/run` now tells folks to open `http://localhost:3000/spaces/system-test/`

PR #207, opened 2021-02-21
Add message about Rails host and port to bin/run

anaulin:
This addresses https://github.com/zinc-collective/convene/issues/206


user512 (approved):

zspencer:
I believe the missing stair was the /spaces/system-test or /spaces/demo path(s) but this is also super useful!

PR #208, opened 2021-03-07
Furniture encapsulates Routes and Views

zspencer:

One thing that was concerning me was the way that Furniture code
definitions was starting to spread throughout the codebase, particularly
how if someone wanted to define a piece of furniture, they would have to
define stuff in `app/views` as well as `app/lib/furniture` and
`config/routes.rb`

This consolidates all those entrance points, so that Furniture can be
defined in a relatively localized place.


convene-web/app/lib/furniture.rb
```diff
@@ -1,9 +1,16 @@
 module Furniture
   REGISTRY = {
-    breakout_tables: Furniture::BreakoutTables,
-    videobridge_jitsi: Furniture::VideobridgeJitsi,
+    breakout_tables: Furniture::BreakoutTablesByJitsi,
```
anaulin, line 3: Minor: we could consider renaming the symbol as well, to include "jitsi" in the name.

anaulin (approved):
🛋️  🚢

convene-web/app/lib/furniture.rb
zspencer: Done!

PR #212, opened 2021-03-21
Remove unused require, update package-lock.json and schema.rb with diffs

anaulin:
This PR removes an unused import in `support/env.js`, and also updates `db/schema.rb` and `package-lock.json` with the diffs I get after cloning the repo from scratch and running `bin/setup` successfully. (I can also now run `bin/test` successfully.)

I'm not 100% if these diffs are all legit things, or if there is e.g. something about version mismatch. I am using node v15.7.0 and npm v7.4.3.


user512 (dismissed):

user512:
Might be a good idea to move `convene-web/.nvmrc` to root and upgrade to latest LTS version so we are all align on node version.

anaulin:
🤦‍♀️  I didn't realize `.nvmrc` is a thing. I've updated the `package-lock.json` in this PR using the correct node version, after which I don't get any diffs.

user512:
It's probably not a bad change tbh, node 15 comes with npm v7 which will bring this new changes anyway.
I am not opposed to upgrading node to the current version but I think the last time we settled on LTS instead.

user512 (approved):

anaulin:
Agreed that upgrading might be a good idea, but it seems separate from the changes in this PR of mine. We could have a separate change where we update the `.nvmrc` and the `package-lock.json` accordingly.

(I don't have strong opinions about this, I am not much of a Node person, as you've seen. 😉 )

zspencer (approved):
😻 😍 😻 Thank you!!!

PR #213, opened 2021-03-21
Use Overmind to run convene-web

user512:
Use [Overmind](https://github.com/DarthSim/overmind) to manager different processes (App server, webpack dev server & maildev).
Use similar setup on CI to ensure things don't break in our dev environment.

Overmind also nicely exit all running maildev or webpack process when we `Ctrl C` out of `bin/run` process.


.github/workflows/test-convene-web.yml
```diff
@@ -34,6 +34,13 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Install Overmind
+      run: |
+        sudo apt-get install tmux
+        sudo mkdir ~/bin/
+        sudo wget -O ~/bin/overmind https://github.com/DarthSim/overmind/releases/download/v2.2.1/overmind-v2.2.1-linux-386.gz
```
zspencer, line 41: May need to `gunzip` the resulting file, since it's likely gzipped given the `.gz` file extension :)

convene-web/.overmind.env
```diff
@@ -0,0 +1 @@
+OVERMIND_PROCFILE=Procfile.dev
```
zspencer, line 1: My instincts are that overmind is a project-level dependency, not a module-level one; so perhaps we could have a Procfile in the top-level directory that starts maildev, the web server, the asset compilation process, and (when we have them) background jobs?

convene-web/bin/run
```diff
@@ -10,7 +10,4 @@
 # visited in browser without doing port forwarding.
 : ${BINDING:=localhost}
 
-echo "******************************************************"
-echo "*** Starting Rails server at http://$BINDING:$PORT ***"
-echo "******************************************************"
```
zspencer, line 15: The point of this particular message was so that it was easier for folks who are getting started to know where they should point their browser to get started; does it get swallowed by overmind?

anaulin (approved):
This is working for me locally. 👍

anaulin (approved):
I insist.

PR #214, opened 2021-03-21
Test release task

user512:
Making sure release won't fail because of bad release task.

### DIFF
--- a/convene-web/spec/tasks/release_after_build_spec.rb
+++ b/convene-web/spec/tasks/release_after_build_spec.rb
@@ -1,11 +1,13 @@
 require "rails_helper"
 
 RSpec.describe "rake release:after_build" do
-  subject do
+  subject(:run_rake_task!) do
     ConveneWeb::Application.load_tasks
     Rake::Task["release:after_build"].reenable
     Rake::Task["release:after_build"].execute
   end
 
-  it { is_expected.not_to raise_error }
+  it "does not fail" do
+    expect { run_rake_task! }.not_to raise_error
+  end
 end
--- a/convene-web/spec/tasks/release_after_build_spec.rb
+++ b/convene-web/spec/tasks/release_after_build_spec.rb
@@ -1,0 +1,13 @@
+require "rails_helper"
+
+RSpec.describe "rake release:after_build" do
+  subject(:run_rake_task!) do
+    ConveneWeb::Application.load_tasks
+    Rake::Task["release:after_build"].reenable
+    Rake::Task["release:after_build"].execute
+  end
+
+  it "does not fail" do
+    expect { run_rake_task! }.not_to raise_error
+  end
+end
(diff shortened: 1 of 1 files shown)


convene-web/spec/tasks/release_after_build_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+RSpec.describe "rake release:after_build" do
+  subject(:run_rake_task!) do
+    ConveneWeb::Application.load_tasks
+    Rake::Task["release:after_build"].reenable
```
user512, line 6: We probably don't need this line but if we intended to have multiple it block, this is needed.

zspencer: Seems useful regardless!

convene-web/spec/tasks/release_after_build_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+    Rake::Task["release:after_build"].execute
+  end
+
+  it "does not fail" do
+    expect { run_rake_task! }.not_to raise_error
```
zspencer, line 11: Since we're using `subject`, could we shrink this down to `it { is_expected.not_to raise_error }`? (I always get confused by how to use rspec in general when dealing with errors :-p)

zspencer (approved):
Thanks!  Love it!

convene-web/spec/tasks/release_after_build_spec.rb
user512:
```suggestion
    it { is_expected.not_to raise_error }
```

convene-web/spec/tasks/release_after_build_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+RSpec.describe "rake release:after_build" do
+  subject(:run_rake_task!) do
```
user512, line 4:
```suggestion
  subject do
```

convene-web/spec/tasks/release_after_build_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+    ConveneWeb::Application.load_tasks
+    Rake::Task["release:after_build"].reenable
+    Rake::Task["release:after_build"].execute
+  end
+
+  it "does not fail" do
+    expect { run_rake_task! }.not_to raise_error
+  end
```
user512, line 12:
```suggestion
  it { is_expected.not_to raise_error }
```

convene-web/spec/tasks/release_after_build_spec.rb
user512: Looks like I can't use one-liner syntax here.
https://github.com/zinc-collective/convene/runs/2163178859

PR #216, opened 2021-03-21
Massive Refactor of Steps, Pages, and Components

zspencer:
OK, the goal here is to:

- Shovel all the interactions with `selenium-webdriver` down into `Page` and `Component` classes; so that our step definitions never have to think about Selenium
- Get JSDoc plugged in for native type-hinting without requiring us to go all-in on TypeScript
- Move us away from having multiple items on `this`; or at least keeping the `this` items down to something like `this.session` and/or `this.factory`so that we have much less implicit connections between steps.
- Pull responsibility off of the `SpacePage` and into child `Component` objects or sibling `Page`s.

Hopefully this will also make it a bit easier to reason about what particular pieces of the test suite are doing; by giving us more affordances for dropping documentation  and type signatures.


anaulin:
💚  🥦  🧹

zspencer:
The two remaining failures appear to be because we're not creating a brand-new-space on each test-run... I want to muse a bit on that; because I'd _love_ for us to spawn new spaces with the blueprint(s) that we expect for the particular test case we're running; but to do so I'd need to expose an API for creating Spaces... which may be fine but I think I'm out of spoons on it for not...

anaulin (approved):
This looks like a great refactor that will enable us to write tests more sustainably. Thank you!

anaulin (approved):
But really.

PR #220, opened 2021-04-03
Better configure Passwordless, add an initializer

anaulin:
This PR addresses some of the issues we uncovered when doing manual testing for https://github.com/zinc-collective/convene/issues/118:
* add a better "from" address for magic link emails
* set an explicit token timeout setting
* set explicit sign in, sign out, and failure redirect paths
* add an explicit setting for redirecting "back" after sign in (where possible)


user512 (approved):
👍

zspencer (approved):
Thank you!!!

Issue #153, opened 2020-11-15
🌫️ `Member`: Profiles


zspencer:
- https://github.com/zinc-collective/convene/issues/117
- https://github.com/zinc-collective/convene/issues/118

People have different ways of representing themselves, and the way they represent themselves in one Space may not be how they want to represent themselves in other Spaces.

## Use Cases
- [x] Create, Update and Remove a `Member` `Profile` (https://github.com/zinc-collective/convene/issues/165)
- [ ] `Person` has a default `Profile`
- [ ] `Member` sets `Profile` for `Space`

zspencer:
@colombene / @anaulin - Does this feel like it synthesizes the different pathways someone may take to enter a room through the Entry Hall?  


(Note: The "Scenario Outline" feature is a way to replace the bit of the step in the `<Actor>` brackets with the value; so that we can run the same test against a number of different user-types. Not sure if we will wind up using it in the actual test side of things. Thoughts @KellyAH / @user512 / @eg-bet / @cjoulain?)

user512:
I think both pathways and the feature outline makes sense.

anaulin:
> customized Identity is the default when they begin to Enter other Rooms in the Space

I had assumed that a person's last used Customized Identity would be the default when they begin to Enter other Rooms in _any_ Convene Space. Or do we want to have a per-Space default? (Seems easier to have one default to begin with, but I see the value in having a per-Space default.)

anaulin:
But yes, I think this makes sense and clarifies (for me) what our planned user flows are.

zspencer:
If it's not too much of a burden, I'd appreciate if our implementation respected that someone's identity is contextual to the space they are within; so that people who do have to present differently in different contexts aren't asked to carry the cost of making that decision each time they enter a space.

That said, I'm OK with us getting there incrementally.

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #221, opened 2021-04-04
WIP - Identity object to store name and pronoun

user512:
Connects: https://github.com/zinc-collective/convene/issues/153

Putting my thought here before I forgot:
1. Create a page behind a feature flag (env var) for person to enter name/ pronoun before entering video room
2. When person enter name/ pronoun, store in session storage so video room can read
3. When person enter another room, the user can see the prior entered name/ pronoun before entering video room and they can change that if they wanted to.
4. If person authenticate thru email, then set session storage name using person.name from the Ruby side.

### DIFF
--- a/convene-web/app/javascript/src/identity.js
+++ b/convene-web/app/javascript/src/identity.js
@@ -1,0 +1,18 @@
+export default class Identity {
+  set name(name) {
+    sessionStorage.setItem("name", name);
+    return true;
+  }
+
+  set pronoun(pronoun) {
+    sessionStorage.setItem("pronoun", pronoun);
+    return true;
+  }
+
+  get personInfo() {
+    return {
+      name: sessionStorage.getItem("name"),
+      pronoun: sessionStorage.getItem("pronoun"),
+    };
+  }
+}
--- a/convene-web/app/javascript/src/video_room.js
+++ b/convene-web/app/javascript/src/video_room.js
@@ -1,5 +1,5 @@
 import { EventTarget } from "event-target-shim";
-import Identify from "./identity";
+import Identity from "./identity";
 
 const eventBus = new EventTarget();
 export default class VideoRoom {
@@ -41,12 +41,12 @@ export default class VideoRoom {
   }
 
   get displayName() {
-    const identify = new Identify();
-    return `${identify.personInfo.name} (${identify.personInfo.pronoun})`;
+    const identity = new Identity();
+    return `${identity.personInfo.name} (${identity.personInfo.pronoun})`;
   }
 
   jitsiApiOption() {
-    const identify = new Identity();
+    const identity = new Identity();
     return {
       roomName: this.roomName,
       parentNode: this.parentNode,
@@ -70,7 +70,7 @@ export default class VideoRoom {
         disableDeepLinking: true,
       },
       userInfo: {
-        displayName: identify.personInfo.name,
+        displayName: identity.personInfo.name,
       },
     };
   }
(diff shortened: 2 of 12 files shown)


convene-web/app/javascript/src/identity.js
```diff
@@ -0,0 +1,18 @@
...
+    sessionStorage.setItem("pronoun", pronoun);
+    return true;
+  }
+
+  get personInfo() {
+    return {
+      name: sessionStorage.getItem("name"),
+      pronoun: sessionStorage.getItem("pronoun"),
+    };
+  }
```
zspencer, line 17: What advantages does the `personItem` method provide over a `get name() { ... }` that encapsulates the sessionStorage call?

convene-web/app/javascript/src/identity.js
```diff
@@ -0,0 +1,18 @@
+export default class Identity {
```
zspencer, line 1: I'm trying to decide between using `Profile` and `Identity` to describe the collection of name/pronouns/avatar. Curious what @anaulin / @cjoulain think?

convene-web/app/javascript/src/video_room.js
```diff
@@ -33,27 +34,44 @@ export default class VideoRoom {
...
         DISABLE_RINGING: true,
         SETTINGS_SECTIONS: [],
-        DEFAULT_REMOTE_DISPLAY_NAME: 'Participants',
+        DEFAULT_REMOTE_DISPLAY_NAME: "Participants",
       },
       configOverwrite: {
         disableDeepLinking: true,
       },
+      userInfo: {
+        displayName: identify.personInfo.name,
```
zspencer, line 73: Does this want to be `this.displayName`?

zspencer (commented):
I think your trajectory sounds reasonable; by "new page" do you mean you want to add an interstitial component that gets served up using [Turbo](https://turbo.hotwire.dev/)/[Hotwire](https://hotwire.dev/) or do you mean something more akin to our "Waiting Room" resource that intercepts when a Room is locked? If the latter, my gut says we could expand the existing waiting room to collect the persons name/pronouns before entering an unlocked room; which could get bypassed for authenticated users who have set their name/pronouns in the database.

Thoughts?

convene-web/app/javascript/src/video_room.js
user512: Yes!

convene-web/app/javascript/src/identity.js
user512: I think `get name()` makes more sense.

user512:
I think the reason I went for `sessionStorage` is because of the different identity in different space and our people model isn't designed for permanently linking to a space.

But maybe what I should be doing is to have an identity Rails model that links to a space so it's more predictable than using `sessionStorage`/ `localStorage`.

convene-web/db/schema.rb
```diff
@@ -44,6 +44,18 @@
     t.index ["room_id"], name: "index_furniture_placements_on_room_id"
   end
 
+  create_table "identities", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "name"
+    t.string "pronoun"
+    t.uuid "person_id"
+    t.uuid "space_id"
```
zspencer, line 51: Maybe instead of `space_id` on `Identity`, we have a `profile_id` on `SpaceMembership` so that an Identity can be used in multiple Spaces?

user512: Good, let me look into it and think it thru. 👍

user512: Something to think about, move avatar_url to Identity or have avatar_url on Person as a fallback?

convene-web/app/javascript/src/identity.js
anaulin: "Profile" feels like the more commonly used term for this kind of thing. But I'm not sure I have a strong opinion about it.

user512:
closing this to redo

PR #223, opened 2021-04-04
Use tweaked custom mailer to email magic links that point the user to the domain they were using

anaulin:
This PR attempts to resolve an issue we found when testing https://github.com/zinc-collective/convene/issues/118, where a magic link URl was directing the user to the incorrect subdomain.

Another option might have been monkeypatching Passwordless, but to me this feels a little less brittle, more explicit, and hopefully easier to further extend in the future.


zspencer (approved):
Thanks Ana!!

PR #225, opened 2021-04-04
Wait for assets before running bin/run

user512:
Before Hivemind, we curl system-test.zinc.local to trigger asset built before running `bin/test`.
Now with Hivemind, we have dedicated asset process to run `webpack-dev-server` so we need to check if the `manifest.json` is ready.


zspencer (approved):
Ahhh nice! Thank you!

PR #226, opened 2021-04-11
Sprout Markdown Text Block Furniture

zspencer:
You can add text to a room now! Very exciting!
Screen Shot 2021-04-11 at 20 28 16
Screen Shot 2021-04-11 at 20 28 10
Screen Shot 2021-04-11 at 20 27 20

This also:

1. Adds Breadcrumbs to the Room and Edit Room pages
2. Sets our default "Update <Model>" text to "Save"
3. Allows folks to edit Room names and slugs
4. Allows folks to edit furniture in their room, assuming the Furniture has attributes that are editable.


convene-web/app/controllers/furniture_placements_controller.rb
```diff
@@ -0,0 +1,21 @@
+class FurniturePlacementsController < ApplicationController
+  def update
+    furniture_placement.update(furniture_placement_params)
```
user512, line 3: It's probably ok because we control the params in the markup but should be always do `update!`?

convene-web/app/lib/furniture/markdown_text_block/_in_room.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="card">
+  <%= furniture.to_html.html_safe %>
```
user512, line 2: I know why it's safe here but do you think we should add `Redcarpet` is making it safe here?

convene-web/app/views/application/_error.html.erb
```diff
@@ -0,0 +1,3 @@
+<%- if model.errors[field].present? %>
+<p class="error-message"><%= model.errors[field].join(', ') %></p>
```
user512, line 2: [minor] Indent line 2?

convene-web/config/locales/en.yml
```diff
@@ -35,6 +35,8 @@ en:
       waiting_room:
         access_code: "Enter Access Code"
     submit:
+      update: "Save"
+      create: "Create"
```
user512, line 39: Ok to check this in, but I can't find the usage for these in this pull request.

user512 (approved):

PR #232, opened 2021-04-18
Specify the node version for the entire project

zspencer:
Turns out we don't specify the Node version for everything!


user512:
I tested by running `nvm use` in both `convene/` & `convene/convene-web/` and both were able to use the specified node version.
Also good for `bin/setup` on both `convene/` & `convene/convene-web/`.

user512 (approved):

Issue #233, opened 2021-04-18
Entering space with no entrance caused a 500 error


user512:
Entering a space with no entrance cause 500, use safe operator will fix that.

zspencer:
@user512 Re: Ending a Call returning someone to the Space Show page

I would recommend is we clarify the intent of the VideoBridge by Jitsi furniture in a Feature test:

```features/furniture/videobridge-by-jitsi.feature
# Someone may want to be in a Room, but not be in an ongoing video call; especially if there
# are other pieces of Furniture in the Room the person may use.
Feature: Ending a Call
  Given a Space Member is in a Room with an active VideoBridge by Jitsi
  When the Space Member ends the call
  Then the Space Member remains in the Room
  And the VideoBridge by Jitsi is no longer open
```

user512:
Confirmed space show page is now working again when there is the space doesn't have an entrance.

PR #236, opened 2021-04-18
Gracefully handle error when space has no entrance

user512:
Connects: https://github.com/zinc-collective/convene/issues/233

I didn't write test for this because:
1. This is breaking all `SpaceController#show` page so it's better to fix it sooner than later. I manually test it locally with a space with no entrance.
2. I want to get some advice/ directions on how to generate seed for different space configuration for feature test and implement in another pull request.

### DIFF
--- a/convene-web/app/views/spaces/show.html.erb
+++ b/convene-web/app/views/spaces/show.html.erb
@@ -4,7 +4,7 @@
   <% end %>
 </div>
 
-<%- space.entrance.furniture_placements.each do |furniture_placement| %>
+<%- space.entrance&.furniture_placements&.each do |furniture_placement| %>
   <%= render partial: furniture_placement.furniture.in_room_template,
              locals: { furniture: furniture_placement.furniture, room: space.entrance,
                        person: current_person } %>
--- a/features/spaces.feature
+++ b/features/spaces.feature
@@ -13,3 +13,9 @@ Feature: Spaces
     Given a Space with a Room specified as its Entrance
     When Anyone visits the Space
     Then they see the Furniture in the Entrance Room
+
+  @built @unimplemented-steps
+  Scenario: Space without Entrances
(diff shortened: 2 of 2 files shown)


convene-web/app/views/spaces/show.html.erb
```diff
@@ -4,7 +4,7 @@
   <% end %>
 </div>
 
-<%- space.entrance.furniture_placements.each do |furniture_placement| %>
+<%- space.entrance&.furniture_placements&.each do |furniture_placement| %>
```
zspencer, line 7: Thank you! 🤦

features/spaces.feature
```diff
@@ -12,4 +12,10 @@ Feature: Spaces
...
     When Anyone visits the Space
-    Then they see the Furniture in the Entrance Room
\ No newline at end of file
+    Then they see the Furniture in the Entrance Room
+
+  @built @unimplemented-steps
+  Scenario: Space without Entrances
+    Given a Space with no Entrance
+    When Anyone visits the Space
+    Then they only see the Rooms
```
zspencer, line 21: TO Perhaps we could unpack this into two assertions:

1. `Then they do not see any Furniture`  and 
2. `And they see the Room Picker for the Space` (which may already exist?)

zspencer (approved):
Love it! Going to merge to fix things :X

zspencer:
Regarding how to test this:

1. We're "missing" a pretty important part of our test infrastructure, which is the ability to create data programmatically in our feature tests. I think it makes sense for us to start tilting at that windmill; and I'd recommend we do so by creating  a `Scenario: Creating a Space via the API` in `features/spaces.feature` which lays out the JSON the API will take in the `When` statement and make assertions about the response + being able to visit the space in the Then(s); as that way you're 'calling your shot' on what the API will look like in a way that other contributors can see.

PR #238, opened 2021-04-18
Eliminate error when running `bin/setup` on a Mac without shared-mime-info installed

zspencer:
Turns out, some dev machines do not yet have the mime info installed; and it's a system-level dependency
So we added it in our `bin/setup` script.


user512 (approved):
nice

PR #240, opened 2021-04-18
Aggregate failures in rspec by default

zspencer:
It's OK for a test to have multiple asserts; especially given how
expensive most of the test setup/teardown is in rails.


user512 (approved):
I love that.

PR #243, opened 2021-04-24
Change Root to Spaces#Show and provide a Default Space

anaulin:


convene-web/config/routes.rb
```diff
@@ -30,7 +32,6 @@
   resources :guides, only: %i[index show]
 
   constraints BrandedDomain.new(Space) do
-    root 'spaces#show'
```
zspencer, line 33: This would break `https://meet.zinc.coop` or other spaces with custom domains. We _probably_ have a test for that? Maybe? Or if we don't, we may want to add one.

That said, there is probably some benefit in making the `root 'spaces#show'` the global root; and have it render a `ConveneSpace` that we expect to always exists (kinda like how `SystemTestSpace` and `DemoSpace` exist in development and review apps). As we build out furniture, it may even serve as the formal public landing page of https://convene.zinc.coop!

zspencer (changes requested):
Once we figure out how to handle the "root when a branded domain shows the branded domains space" I think this will be good-to-go.

convene-web/config/routes.rb
anaulin: I brought back `space#show` as the root route everywhere, and added a dummy blank `Space` to show when there is no space specified.

anaulin:
This ended up not being necessary because I found a workaround to the issue in my spec that makes it so that I don't need to redirect to `/` in the spec. Do we want to keep any of this code, or should I just close this PR, @zspencer ?

zspencer:
Deprecated by https://github.com/zinc-collective/convene/pull/251

PR #244, opened 2021-04-25
Do not keep installing shared-mime-info

user512:
Check if a brew package is installed using `brew ls --versions`
Not all packages is in `PATH`

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -14,8 +14,8 @@ function install_package {
 }
 
 function setup_package {
-  output=$(command -v $1)
-  if [ -z $output ]; then
+  output=$(brew ls --versions $1)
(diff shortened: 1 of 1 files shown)


bin/setup
```diff
@@ -22,8 +22,17 @@ function setup_package {
...
+  output=$(brew ls --versions $1)
+  if [ -n $output ]; then
+    install_package $1
+  else
+    echo "$1 installed, continuing..."
+  fi
+}
+
+setup_cli_tool "overmind"
+setup_cli_tool "yarn"
```
zspencer, line 35: Instead of splitting this into two functions, could we do an or, or a nested if or else if?

zspencer (approved):

PR #245, opened 2021-04-25
Authorize updates to Furniture Placements in controller

anaulin:
To prevent non-space-members from messing with a Space's furniture.


user512 (approved):

zspencer (approved):
So gooooood!

PR #247, opened 2021-05-02
Prefix feature flag environment variables with `FEATURE_`

zspencer:
I was getting a bit worried about figuring out what env variables were
feature flags and what weren't; as well as making sure feature flags are
disabled by default.

So I took a refactor pass!


user512 (approved):
I like that.

PR #248, opened 2021-05-02
Restructure Header and Footer to prioritize the Space

zspencer:
- Moves Convene to a Footer
- Adds Breadcrumbs to the Header and Footer
- Gets rid of Avatars since we don't have it yet
- Gets rid of the tap to Sign out / Sign In

Screen Shot 2021-05-01 at 7 06 36 PM
Screen Shot 2021-05-01 at 7 06 48 PM
Screen Shot 2021-05-01 at 7 07 02 PM

### DIFF
--- a/convene-web/app/javascript/src/base/layout.scss
+++ b/convene-web/app/javascript/src/base/layout.scss
@@ -1,5 +1,5 @@
 @layer base {
-  html {
+  body {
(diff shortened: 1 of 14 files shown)


user512 (approved):
Thanks for also attaching the screenshot

PR #249, opened 2021-05-02
Docs/move root level instructions for development back readme

KellyAH:
move development documentation to a doc that makes it easier to find and makes more sense.


zspencer (approved):
Super appreciate this! 

Basically we're going for "Start at project root, then each child/module-folder has stuff that is _only_ relevant to that particular module. Both Ruby and Node are project-level dependencies; so pulling this up to the top-level CONTRIBUTING guide and leaving `convene-web` to reference it makes a lot of sense!

Issue #252, opened 2021-05-07
🌱🥔⛰️🔌 `Utilities`


zspencer:
Utilities are how Convene Neighborhoods, Spaces, and Furniture integrate with other services on the web.  Some examples:

- The Stripe Utility allows a Space or Furniture to receive online payments. 
- The Plaid Utility allows 
- The Jitsi Utility provides real-time Video and Audio support.
- The AirTable Utility allows Furniture to reflect data from a shared Spreadsheet.
- The Google Groups Utility allows Furniture to manage Google Group Memberships programmatically.

Etc.

Utilities tend to emerge from Furniture that would like to leverage a 3rd-party service, and get extracted into the `app/utilities` folder once a few pieces of Furniture are talking to the same service.

## Use Cases
- [ ] https://github.com/zinc-collective/convene/issues/1430
- [x] https://github.com/zinc-collective/convene/issues/1002
- [x] Managing Utilities
  - [x]  🥔 Adding a Utility to a Space
  - [x] 🥔 Utilities can validate themselves
  - [x] 🥔 Utilities may be configured
  - [ ] Removing a Utility
    - [ ] In-use Utilties may not be removed
  - [ ] https://github.com/zinc-collective/convene/issues/891

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #253, opened 2021-05-07
Initial implementation of Utilities

zspencer:
See: https://github.com/zinc-collective/convene/issues/252


convene-web/app/lib/furniture.rb
```diff
@@ -1,19 +1,22 @@
...
     videobridge_by_jitsi: Furniture::VideobridgeByJitsi,
     markdown_text_block: Furniture::MarkdownTextBlock
-  }
+  }.freeze
 
-  # Allows Furniture to expose their controllers
+  # Appends each {Furniture}'s CRUD actions
   def self.append_routes(routing_context)
-    REGISTRY.values.each do |furniture|
+    REGISTRY.each_value do |furniture|
```
user512, line 14: TIL about `each_value`

user512 (approved):
I feel like I am going to need to have you to demo this a bit to completely understand what's going on but this looks good to me.

convene-web/app/models/space.rb
```diff
@@ -56,4 +53,22 @@ def unlocked?
...
+  has_many :utility_hookups
+
+  def jitsi_meet_domain
+    jitsi_hookup = utility_hookups.find_by(utility_slug: :jitsi)
+    return if jitsi_hookup.blank?
+
+    Utilities.from_utility_hookup(jitsi_hookup).meet_domain
+  end
+
+  def hookups
```
anaulin, line 69: Is this used anywhere? I couldn't find a usage example in this PR (but its big, so likely I missed it?). Looking at the implementation, it reads like it returns Utilities, not UtilityHookups, so I was trying to understand how it is used.

anaulin (approved):
It's starting to feel quite fleshed out! 🍖

convene-web/app/models/space.rb
zspencer: Probably not! Thank you!

PR #254, opened 2021-05-07
Clean up how population of the development and prod databases

zspencer:
- Added a Blueprint for Zinc's meet.zinc.coop domain
- Blueprints can add members using either a hash or a Person
- Call `release:after_build` as part of `bin/setup`


user512 (approved):
Looks great!

PR #259, opened 2021-05-08
Replace phone icon with "open door" icon for the "Enter Room" button.

anaulin:
Small visual tweak, so that we no longer have a phone icon for rooms, since rooms might or might not have "call" capabilities.

Looks like this:
Screen Shot 2021-05-08 at 3 26 11 PM


zspencer (approved):
This is nice! Thank you!

PR #267, opened 2021-05-22
Use pessimistic version constraints in Gemfile

zspencer:
When we do not specify the version, we run the risk that a `bundle
update` will update package(s) in a way that is incompatible. If we use
the twiddle-wakka `~>` syntax, we are less likely to run into an incidental
incompatibility.

There is something to be said for leaving dependecy resolution unbounded
and expecting that our test suite will catch incompatibilities; but I'm
not sure I'm there yet.

see:
- https://guides.rubygems.org/patterns/#pessimistic-version-constraint
- https://thoughtbot.com/blog/rubys-pessimistic-operator


convene-web/Gemfile
```diff
@@ -34,42 +34,42 @@ gem 'jbuilder', '~> 2.7'
...
+gem 'pg', '~> 1.2'
 
 # User-facing features
 #
 # Allows logging in via magic links
-gem 'passwordless'
+gem 'passwordless', '0.10.0'
 # Provides powerful administration features
-gem 'administrate'
+gem 'administrate', '0.16.0'
```
user512, line 58: Why lock this one to 0.16.0?

user512 (approved):

convene-web/Gemfile
zspencer: Because the version number is less than 0, semver does not have guarantee that patch or minor version bumps will not break the APIs it exposes.

Issue #235, opened 2021-04-18
Github actions should run on forked repo


user512:
Github now supports running action workflow by using `pull_request_target`
It's likely as simple as just changing [this line](https://github.com/zinc-collective/convene/blob/5c9a2f01d4838165e7f498a49d42ab1278527be4/.github/workflows/test-convene-web.yml#L5) to 
`pull_request_target`.

Ref: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

PR #268, opened 2021-05-22
Run GH actions on forked repo

user512:
Connects: https://github.com/zinc-collective/convene/issues/235
so we can run test when contributors open pull request on their forked repo.


zspencer (approved):
This is great!

PR #271, opened 2021-05-23
Fix broken symbolic link for nvmrc

zspencer:
When working with CJ, we noticed that their environment was haivng some
weird issues. Turns out, the .nvmrc file in the convene-web directory
was not a valid symbolic link; which may have been throwing their
node-version manager for a loop.

Symbolic Links: https://www.lifewire.com/create-symbolic-links-ln-command-4059723

### DIFF
--- a/convene-web/.nvmrc
+++ b/convene-web/.nvmrc
@@ -1,1 +1,1 @@
-.nvmrc
+../.nvmrc


zspencer:
@cjoulain - If you check out this branch, does it resolve the issue you were experiencing with the node version being flakey?

cjoulain:
I don't know if this is more of a shell-specific issue? When opening a new terminal, my default seems  to be  an outdated version of Node (v 8.16.0). Example here:

node-version

The Node version doesn't update with this branch but I ran `nvm use 12.18.3` and that seemed to take when testing on a new terminal. I'm not sure if there's a way to indicate an alias setting within .nvmrc or, for those facing this issue, if it's easier to mention this in the README (to run `nvm use {node_version}`. What do you recommend?

zspencer:
What happens if you use `nvm use` without specifying a version? My hope is that will pick up the `.nvmrc` file?

I would prefer not to specify fixes for a particular version manager in our README. Each version manager has it's own foibles and flaws and weirdness, and it feels more like the responsibility of the version manager to provide clear debugging instructions. 

That said, it looks like [nvm has some guidance on deeper shell integration](https://github.com/nvm-sh/nvm#deeper-shell-integration) that may be worth exploring.

user512 (approved):

user512:
Tested locally.
```bash
❯ pwd
/Users/tomlee/dev_work/zinc/convene
❯ nvm which
Found '/Users/tomlee/dev_work/zinc/convene/.nvmrc' with version <12.18.3>
/Users/tomlee/.nvm/versions/node/v12.18.3/bin/node
❯ cd convene-web
❯ nvm which
Found '/Users/tomlee/dev_work/zinc/convene/.nvmrc' with version <12.18.3>
/Users/tomlee/.nvm/versions/node/v12.18.3/bin/node
```

cjoulain:
Just confirming this helped. I had been running `nvm use {version}` when it looks like `nvm use` does the trick.

Screen Shot 2021-05-24 at 8 01 29 PM

zspencer:
Woo! Glad to hear it!

Issue #270, opened 2021-05-22
Bug: `bin/setup: line 21: [: shared-mime-info: binary operator expected`


zspencer:
Looks like our setup shell script has a sad. It may be that we do not even need to include this package anymore, especially since mimemagic was ditched at some point. 

To reproduce:

1. Clone the repo
2. Run `bin/setup`
3. Haz-a-sad

### To prove your fix works

1. `brew uninstall shared-mime-info` and `gem uninstall mini_mime`
2. `bin/setup`
3. `bin/test`

zspencer:
Fixed by #277

PR #277, opened 2021-05-29
Small improvements to bin/setup

anaulin:
This PR tweaks a couple of things I noticed re-setting this repo from scratch:
* make our string handling in bash resilient to the cases where `$brew_output` or `$command_output` might contain spaces
* fix the check for existence of `$brew_output` to check for an empty string (I believe this check was previously logically inverted)
* make `convene-web/bin/setup` make a copy of `.env.example` if there isn't an existing `.env`

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -3,7 +3,7 @@
 function install_package {
   echo "Attempting to install $1"
   output=$(command -v brew)
-  if [ -n $output ]; then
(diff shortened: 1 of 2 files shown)


zspencer (approved):
Oooh, this looks like it fixes the shell warning I was getting: https://github.com/zinc-collective/convene/issues/270

PR #278, opened 2021-05-29
Update webpacker, avoid depending on node-sass

anaulin:
This fixes an issue when running the server within overmind on Mac OS X, which had some issues with `node-sass v 4.14.1`.


zspencer (approved):
A patch a day keeps the bitrot away

PR #279, opened 2021-05-30
Streamline Space and Room configuration

zspencer:
Coops not cages wanted to update their content, but because they have an Unlisted Room as their Entrance Hall, there isn't any way for them to find the configure button!

We:

- Switched to a "Gear" for configure, which seems more in line with visual conventions
- Got our ass kicked by the CSS trying to get rid of the icon underline (But we did it!)
- Added a list of Rooms with links to Configure them on the Space Configuration page.


anaulin (approved):
⚙️  ✅

PR #280, opened 2021-05-30
Cache brew caches between CI runs

anaulin:
To speed up each run:
* cache the Homebrew cache between CI runs
* stop installing `shared-mime-info`, since we  no longer need it


zspencer (approved):
This shaved off ~2m! Nice!

PR #282, opened 2021-05-30
Use 'Sign in' when submitting the AuthenticatedSession form

zspencer:
- Decomposes locale files on a per-domain basis


anaulin (approved):
Thank you for organizing things better! 🗄️

PR #284, opened 2021-05-30
Allow email's to be delivered from a real address

zspencer:
This way folks can reply and ask for help and such

- Fixes `convene-web/bin/setup` to copy `convene-web/.env.example`
when `convene-web/.env` does not exist.
- Fixes the default URL builder to use `protocol` and not `scheme` when
  building URLs


anaulin (approved):
Much nicer to have customers be able to yell back! ✉️

PR #290, opened 2021-06-06
Remove deprecation warnings related to autoloading

anaulin:
This change adds a `Rails.application.reloader.to_prepare do` around an initializer call to `ActionMailer::Base`. This removes the deprecation warning:
```
DEPRECATION WARNING: Initialization autoloaded the constants ActionText::ContentHelper and ActionText::TagHelper.
```

`Rails.application.reloader.to_prepare` is meant explicitly for autoloading during application boot, with the caveat that the contents of the block must be idempotent (which this code is) since it might be called twice. Rails docs: https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-when-the-application-boots

We still have a handful of errors `warning: already initialized constant` on startup, but that seems more innocuous and also harder to fix.

### DIFF
--- a/convene-web/config/initializers/action_mailer.rb
+++ b/convene-web/config/initializers/action_mailer.rb
@@ -1,21 +1,23 @@
-if ENV["SMTP_ADDRESS"]
-  ActionMailer::Base.smtp_settings = {
-    address: ENV["SMTP_ADDRESS"],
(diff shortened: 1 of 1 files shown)


zspencer (approved):
Thank you! I was very confused about how to tackle that warning.

PR #291, opened 2021-06-06
Administrate is causing constant lookup issues

zspencer:
For whatever reason, Administrate is making things go sideways with
zeitwerks constant lookup.

Since we're not _using_ Administrate, it seems to me that removing it
may be the simplest path forward; and instead start building out our
client-management and operator-facing features directly into the
application.


anaulin (approved):
🔪  it with 🔥

PR #306, opened 2021-06-27
Clean up more Administrate issues

zspencer:
Heroku wasn't able to deploy Convene because Administrate was trying to compile it's assets, because I forgot to remove it :X


anaulin (approved):
🧹

PR #312, opened 2021-07-10
Plaid can be configured with secrets and environment

zspencer:

Also, Utility Hookup configuration is now encrypted at rest; which
should make things a bit safer.

Co-authored-by: Ana Ulin <ana@ulin.org>


anaulin (approved):
🔒 🤫

PR #318, opened 2021-07-25
Wire up Plaid furniture and basics of linking a bank account

anaulin:
As part of a solution to https://github.com/zinc-collective/convene/issues/186, this PR sketches up a piece of furniture that uses a Plaid utility hookup to make it possible to "link" a bank account. (The data is not yet being saved anywhere, that will come in a later PR.)

Demo:
Kapture 2021-07-24 at 18 01 29


zspencer (approved):
This is probably safe for now, since I don't expect anyone to actually be adding the Plaid utility :X

PR #326, opened 2021-08-08
Check Drop Box: Exchange Plaid's Public Token for Access Token

zspencer:
We still have to:
- [ ] Persist the Check to the Database
- [ ] Figure out how we want the Space Owner to list the checks to deposit
- [ ] ???

But we can do that in a later PR!

### DIFF
--- a/convene-web/app/lib/furniture/check_dropbox.rb
+++ b/convene-web/app/lib/furniture/check_dropbox.rb
@@ -1,50 +1,23 @@
 # frozen_string_literal: true
 
 module Furniture
-  # Renders some HTML in a {Room}.
+  # Allows a Space to receive eChecks
+  # @see features/furniture/check-drop-off.feature
   class CheckDropbox
     include Placeable
 
     def link_token(person)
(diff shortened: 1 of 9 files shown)


anaulin (approved):
🏦  👍

convene-web/app/lib/furniture/check_dropbox.rb
```diff
@@ -1,50 +1,67 @@
...
+        self.access_token = response.access_token
+        self.item_id = response.item_id
 
-    def plaid_configuration
-      Plaid::Configuration.new do |configuration|
-        configuration.server_index = Plaid::Configuration::Environment[plaid_environment]
-        configuration.api_key['PLAID-CLIENT-ID'] = plaid_client_id
-        configuration.api_key['PLAID-SECRET'] = plaid_secret
-        configuration.api_key['Plaid-Version'] = plaid_version
+        byebug
```
zspencer, line 36: This probably needs to go before merging.

PR #327, opened 2021-08-08
Guest may accept Invitation

zspencer:

TODO:
- [x] Fix the failing tests!
- [x] Let's make it less brittle so it doesn't explode when trying to
  accept an invitation for an email address that is already registered

Co-authored-by: KellyAH <KellyAH@users.noreply.github.com>
Co-authored-by: CJ Joulain <cjoulain@users.noreply.github.com>


anaulin:
@zspencer did you mean to take this out of `draft`? (since you put it up for review already)

convene-web/app/models/rsvp.rb
```diff
@@ -0,0 +1,19 @@
...
+  def update(attributes)
+    invitation.update(status: attributes[:status])
+    person = Person.create(name: invitation.name, email: invitation.email)
+    person.space_memberships.create(space: invitation.space)
+    authentication_method = person.authentication_methods.create(contact_method: :email, contact_location: invitation.email)
+
+    authentication_method.send_one_time_password!(invitation.space)
+  end
+
+  def persisted?
```
anaulin, line 16: Is this here to make some other ActiveRecord magic work? It's fine if there is no other workaround, but it feels a little meh to have a this non-persisted model claim that it is persisted.

anaulin (approved):
👍  Looks good, once the email test assertion is fixed.

zspencer:
@anaulin - I use "Draft" to indicate that it's not ready for Merge (because of the broken test). I like to request feedback whenever I put something down for the day; so that I can integrate that feedback earlier.

convene-web/app/models/rsvp.rb
zspencer: Sooo yeaaaaaaaa this is a sneaky butt to make Rails' URL builder play nicely.

`persisted?` is checked when determining whether to do a PUT or a POST, and POST actions default to the plural form of the model; so instead of `invitation_rsvp_path` being called when we do `form_with model: [:invitation, rsvp]` it calls `invitation_rsvps_path`. 

This could imply that what we _want_ is for an `Invitation` to have many RSVPs, but since an `Invitation` has an implicit 1-to-1 relationship with an `Invitee`; and it felt like allowing someone to RSVP multiple times was a bit odd in that case.

anaulin: _grumbles something unintelligible about magic and Rails_

Might be worth summarizing some of that in a code comment here, for future readers of this code.

_walks away grumbling some more_

zspencer: Hah, yea, I think we have a couple options; one is to ... simply accept that an invitation may "have many" rsvps; even though we don't persist them...

zspencer:
Redis is not wanting to work on CI; so gonna punt on that for a sec and look into it later.

PR #342, opened 2021-08-22
Add Rails support to Solargraph

zspencer:
For those of us who use Solargraph for type-hinting, method definition / documentation previews and similar delightful features; you'll notice that Rails classes are often ... missing.

This should make that a little more consistent!

See:
- https://solargraph.org/guides/rails
- https://github.com/castwide/solargraph/issues/87
- https://github.com/iftheshoefritz/solargraph-rails

### DIFF
--- a/.solargraph.yml
+++ b/.solargraph.yml
@@ -1,0 +1,23 @@
+---
+include:
+- "**/*.rb"
+exclude:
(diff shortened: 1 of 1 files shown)


user512 (approved):
purffect!

PR #349, opened 2021-08-28
Make Blueprint more Idempotent

zspencer:
When running `bin/setup`, we re-apply the blueprint for the System Test space.

Thich was clobbering the Furniture Placements and Utility Hookups and other such changes.
Now it should not clobber changes made on the local system.


anaulin:
This change looks ok on the surface of the the diff, but it is consistently breaking an existing spec: `./spec/models/demo_space_spec.rb:16`. 🤔

user512 (approved):

Issue #159, opened 2020-12-07
Authentication: Sign In by entering emailed One-Time code


zspencer:
Extends https://github.com/zinc-collective/convene/issues/118

Per @user512's [comment](https://github.com/zinc-collective/convene/issues/118#issuecomment-739436282); it may make sense for us to defer implementing the identification via one-time emailed code  so we can stay focused on the main thrust of the Andromeda release.

- [x] Identification via following an emailed link
  - [x] Write scenario
  - [x] Implement step definitions
  - [x] Implement feature
  - [x] Update guide to identificiation
```features/identification.feature
  Scenario: Identification via entering a one-time code delivered by Email
    Given I have submitted my email in the sign in form
    When I enter the code emailed to me 
    Then I am identified within the Space for a week
```

zspencer:
This is in production, and the guide has been updated. Wooo! Closin' tickets!

PR #367, opened 2021-09-11
Update authentication guide to include one time code

zspencer:
See: https://github.com/zinc-collective/convene/issues/159

This also un-pends the feature tests, and updates the test harness to
expose affordances for signing in via one-time-code.


anaulin (approved):
Thanks for buttoning this up! ✅

PR #378, opened 2021-09-19
Check Dropoff's have a Status

zspencer:
We want to be able to track the status of Checks so we can remove the Items from Plaid eventually.


anaulin (approved):
💚

PR #394, opened 2021-10-02
Remove FontAwesome as a dependency

zspencer:
We are using more emojis for our iconography; instead of font-awesome.


anaulin (approved):
👏

PR #397, opened 2021-10-05
Define a policy for who should be allowed to do what with Checks

zspencer:

A quick proof of concept for writing policy specs in furniture.


convene-web/spec/furniture/check_dropbox/check_policy_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    permissions :show?, :update?, :edit?, :destroy? do
+      it { is_expected.not_to permit(actor, check) }
+    end
+
+    permissions :create?, :index? do
+      it { is_expected.to permit(actor, check) }
+    end
+  end
+
+  context 'as a neighbor' do
```
anaulin, line 18: You're using double quotes in the other `context` descriptions. :-)

zspencer: Doh! Thank you!

zspencer:
Yea, I was thinking I want to start figuring out how to do request specs at the very least; and that led me down getting FactoryBot working for furniture.

I wound up bumping up into not having a clear way to provide a fake plaid; but I think that's the next step.

PR #425, opened 2021-10-28
Check Dropbox: Only show link to all checks to those who can see them

zspencer:
See: http://github.com/zinc-collective/convene/issues/186


anaulin (approved):
Yay policies! 🔒

PR #428, opened 2021-10-30
Move `lib/furniture` and `lib/utilities` up a level

zspencer:
Instead of burying `furniture` and `utilities` beneath lib, let's make them top-level concepts like `model` etc.


anaulin (approved):
Sending thoughts and prayers. 🙏

PR #429, opened 2021-10-30
Use port 3000 by default

zspencer:


anaulin (approved):
Thanks for all the fixings! 🥪

PR #433, opened 2021-11-06
Something is causing method / constant lookup to go sideways

zspencer:
Errors like `Superclass Mismatch` and `undefined method app_views_...`
are getting thrown every which way in a few hours of coding. I don't
quite get why this is, but I am hopeful that upgradine zeitwerk may
help?

Alternatively, we may want to dig into puma?


anaulin (approved):
🛕 🤞 🍀 🙏

PR #442, opened 2021-11-14
Automate check for accepting an Invitation as a Guest

zspencer:

Strangely, there seems to be some Unobtrusive JavaScript left over in our application from the move to Turbo, which had broken the invitation flow!

Thankfully, our TDD ensemble discovered it as part of automating that check; and we fixed it!

This also:

- Reframes the Accepting an Invitation as a Guest Scenario to use existing Invitation Steps and the newer pattern we follow for defining steps
- Finds and resolves some greedy regexs in parameter types

### DIFF
--- a/features/spaces/invitations.feature
+++ b/features/spaces/invitations.feature
@@ -39,13 +39,17 @@ Feature: Spaces: Invitations
     And that Invitation is Accepted
     And all other Invitations to that Invitation's contact info are Resolved
 
-  @unstarted @andromeda
+  @built @andromeda
   Scenario: Accepting an Invitation as a Guest
-    Given an Invitation was sent less than 14 days ago
-    When a Guest accepts that Invitation
-    Then the Guest becomes a Space Member
-    And that Invitation is Accepted
-    And all other Invitations to that Invitation's contact info are Resolved
+    Given a "System Test" Space
+    And the "System Test" Space has a Space Owner "space-owner@example.com"
+    And an Invitation to the "System Test" Space is sent by Space Owner "space-owner@example.com"
+      | name | email                       |
+      | A Guest | a-guest@example.com |
+    When the Invitation to "a-guest@example.com" for the "System Test" Space is accepted by the Guest "a-guest@example.com"
+    Then the Guest "a-guest@example.com" becomes a Space Member of the "System Test" Space
+    And the Invitation to "a-guest@example.com" for the "System Test" Space has a status of "accepted"
+    And all other Invitations to "a-guest@example.com" for the "System Test" Space have a status of "accepted elsewhere"
 
   # We want to make sure that people who were invited but
   # didn't take action can't suddenly appear and disorient
--- a/features/spaces/invitations.feature
+++ b/features/spaces/invitations.feature
@@ -49,7 +49,8 @@ Feature: Spaces: Invitations
     When the Invitation to "a-guest@example.com" for the "System Test" Space is accepted by the Guest "a-guest@example.com"
     Then the Guest "a-guest@example.com" becomes a Space Member of the "System Test" Space
     And the Invitation to "a-guest@example.com" for the "System Test" Space has a status of "accepted"
-    And all other Invitations to "a-guest@example.com" for the "System Test" Space have a status of "accepted elsewhere"
+    # @todo implement me!
+    # And all other Invitations to Space Member "a-guest@example.com" for the "System Test" Space no longer have a status of "pending"
 
   # We want to make sure that people who were invited but
   # didn't take action can't suddenly appear and disorient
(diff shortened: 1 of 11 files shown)


zspencer:
Thanks for rebasing `development` into this!

anaulin:
@zspencer is this ready for review? (not sure because it is not wip or draft, but also there are no reviewers)

zspencer:
@anaulin - We are planning to merge it as the test ensemble; but we didn't gather yesterday (so we may merge it in a couple weeks)

If you have thoughts on it, feel free to drop them! Especially because it would give the ensemble an opportunity to read and respond to asynchronous feedback (rather than just the in-the-moment design review we do during session!)

features/spaces/invitations.feature
```diff
@@ -39,13 +39,18 @@ Feature: Spaces: Invitations
...
-    And all other Invitations to that Invitation's contact info are Resolved
+    Given a "System Test" Space
+    And the "System Test" Space has a Space Owner "space-owner@example.com"
+    And an Invitation to the "System Test" Space is sent by Space Owner "space-owner@example.com"
+      | name | email                       |
+      | A Guest | a-guest@example.com |
+    When the Invitation to "a-guest@example.com" for the "System Test" Space is accepted by the Guest "a-guest@example.com"
+    Then the Guest "a-guest@example.com" becomes a Space Member of the "System Test" Space
+    And the Invitation to "a-guest@example.com" for the "System Test" Space has a status of "accepted"
+    # @todo implement me!
```
anaulin, line 52: Is there a more "proper" Cucumber annotation that could be used here, instead of a comment? That way we would get reminders in the Cucumber output when we run the tests.

features/steps/invitation_steps.js
```diff
@@ -15,6 +18,26 @@ When(
...
+  /**
+   * @param {Invitation} invitation
+   * @param {Space} space
+   * @param {Actor} actor
+   */
+  function (_, invitation, _2, space, _3, actor) {
+    return (
+      invitation
+        .rsvpLink()
+        // @todo Refactor to live in the harness
```
anaulin, line 32: I am not sure what is our workflow with TODOs like this one. Do we just leave them in the code for later? Do we make issues to remind ourselves to revisit this? Mostly I'm wondering how will we make sure that a TODO like this will eventually be either addressed (and removed), or declared to be not-worth-doing (and removed).

features/steps/invitation_steps.js
```diff
@@ -36,3 +59,44 @@ Then(
...
+);
+
+Then(
+  "all other Invitations to {actor} for {a} {space} no longer have {a} status of {string}",
+  /**
+   * @param {Actor} actor
+   * @param {Space} space
+   * @param {string} status
+   */
+  function (actor, a, space, a2, status) {
```
anaulin, line 93: Seems like `a` and `a2` params here be underscores instead of `a`s, to be consistent with what we have on line 70, and to make it clearer that those are unused.

anaulin (approved):
Looks good! I think this is ready to merge. 🚢 

I have a couple of minor, non-blocking comments below, mainly about style and thinking about future-us and handling of TODOs. Feel free to address those or not, if you think they don't make sense, and merge when you're ready.

features/spaces/invitations.feature
zspencer: It looks like we could have two options:

1. We could mark the step as `pending,` which would fail the test, and let us know we didn't implement this step yet.
2. We could make the step a no-op, where it would just "pass," but that would obscure the fact that it's not quite there yet.

I think I'm leaning towards leaving it as "@todo" in the code; so that someone who notices it and has the capacity can go attempt to implement it. 

That said, we don't really have norms or guidance around how to handle `@todo.` I tend to think of them as an 'aspirational message for future Zee' and not a "someone else cleans this up because I don't give any fucks." Still, I anticipate that others may have different reads on what `@todo` means).

features/steps/invitation_steps.js
zspencer: I think we can tackle this before merging, since it's pretty concrete; and doesn't require adding new steps or implementing parts of the feature.

features/spaces/invitations.feature
zspencer: (Note: We also have a checkbox for finishing this) https://github.com/zinc-collective/convene/issues/117

anaulin: I like trying to align on `{@}todo` as meaning "someone who notices it and has the capacity can go attempt to implement it". We could also document this intent in the `CONTRIBUTING` guide or something, if we have the capacity for that.

features/steps/invitation_steps.js
zspencer: Good catch! My robot is supposed to get those but I did not.

features/spaces/invitations.feature
zspencer: Yea, I need to run atm but would happily merge someone adding it!

PR #454, opened 2021-12-04
Update turbo-rails version to one that actually exists

anaulin:
This fixes a build breakage caused by `turbo-rails` erroneously bumping their version number, then dependabot updating our dependency to the bad version, and `turbo-rails` then yanking the bad version. See: https://github.com/hotwired/turbo-rails/releases/tag/v0.8.2

This PR also bumps the `globalid` version to 1.0.0, which should be safe, since it is not a code change and a "now we are stable" change (see: https://github.com/rails/globalid/releases/tag/v1.0.0)


anaulin:
Duh this is already updated in main branch, just not in the other branch I was looking at. 🤦

zspencer:
yea this ate an hour for me last Saturday. Thank you for tackling it! Apologies for not leaving a clear enough trail that it had been resolved.

anaulin:
Nah, I think I wasn't braining right on Saturday and missed this and some other stuff. That's on me.

PR #455, opened 2021-12-05
Don't list unlisted rooms in Space

anaulin:
This PR fixes a small bug introduced in https://github.com/zinc-collective/convene/pull/441, by making unlisted rooms not listed (but accessible with a URL).

I also tweaked the feature tests to avoid assuming that rooms always have a video panel.

I believe this should fix the feature test failures we've been seeing.

### DIFF
--- a/app/views/spaces/show.html.erb
+++ b/app/views/spaces/show.html.erb
@@ -1,6 +1,6 @@
 <% breadcrumb :root %>
 <div class="space-container card-list">
-  <% policy_scope(space.rooms).each do |room| %>
+  <% policy_scope(space.rooms.listed).each do |room| %>
     <%= render partial: 'spaces/room_card', locals: { room: room } %>
   <% end %>
 </div>
--- a/features/steps/room_steps.js
+++ b/features/steps/room_steps.js
@@ -82,8 +82,8 @@ When("the {actor} taps the {room} in the Room Picker", function (actor, room) {
 });
 
 Then("the {actor} is placed in the {room}", async function (actor, room) {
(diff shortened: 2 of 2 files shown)


anaulin:
Sigh.

I don't understand why this test failure happens, and I cannot reproduce it locally. It does seem to be fail consistently on CI, on the `development` (main) branch and others.

One thing I notice looking at the screenshots of the failed tests on CI is that the actor, despite being "Guest", is both signed in and has access to the edit link for the Space, so they aren't really a Guest. Even so, locally a space owner doesn't see the unlisted rooms, so I'm not sure what to make of this. Here is the screenshot:
Guest_may_not_discover_Unlisted_Rooms

zspencer:
That is strange; I would expect scoping to listed would resolve this.

I've been thinking that the Room List probably wants to be Furniture; and rather than each Room knowing its listed/unlisted state; people add Rooms to the Room List to provide affordances for navigating the space. 

This may simplify things significantly; but would be a substantative change that we don't want to nibble at now. Thoughts?

zspencer:
Note: The reason I went in with a Big Hammer in https://github.com/zinc-collective/convene/pull/447 so that each test runs against it's own Space was because I _think_ what's happening is we're winding up in a place where all the unlisted rooms have been flipped to listed; but that may be an invalid hypothesis.

anaulin:
Ugh, I didn't notice https://github.com/zinc-collective/convene/pull/447/files in the list of open PRs 🤦‍♀️ . Closing this PR as duplicated / unnecessary work. We can focus on getting #447 working instead.

PR #462, opened 2021-12-16
Two classes in the same file breaks Rails autoloading

zspencer:
It turns out, by storing these Policy objects with their Utitility
definition, Zeitwork (or the Rails Autoloader, which I think is
Zeitwerk?) pulls them in, redefining their classes; and causing a
superclass type mismatch.

By splitting them out, Rails autoloader loads them each only once and
only when they have been changed (which triggers some kind of class
definition class flush)

Co-Authored-By: Ana Ulin <ana@ulin.org>


anaulin (approved):
🧹 🥦  👍

PR #463, opened 2021-12-16
Add Validation to Checks

zspencer:

As we were doing a "cleanup" review of the interaction flow for the payment form, we realized that there are no validations on any of the data. This was in part because we were rushin', and also because the way we're doing Item as a thing independent of ItemRecord was making error handling a bit harder than if it were an ApplicationRecord instead of a Model.

- [x] Make saving the form actually work
- [x] [Independently merge in the changes for consolidating to attribute from field or attribute](https://github.com/zinc-collective/convene/pull/469)
- [x] `check_dropbox/_in_room`  and `check_dropbox/checks/new` are basically copies
- [x] get specs to 🟢 
  - [x] in `check_request_spec.rb`, the faking of `Plaid::PlaidUtility` in the `before` block needs to be ~~killed with fire~~ adjusted to match the new utility structure 
- [ ] Maybe? Figure out how to indicate that a piece of Furniture relies on a Utility that requires loading javascript asset(s); so that it's not something we have to write into each view that happens to use the utility.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -89,6 +89,10 @@ group :development, :test do
   gem 'rspec-rails'
   gem 'shoulda-matchers', '~> 5.1'
   gem 'rails-controller-testing'
+
+  # Let the robots do the request/response faking.
+  gem 'vcr'
+  gem 'webmock'
 end
 
 group :development do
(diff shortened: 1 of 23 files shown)


anaulin (approved):
🦭

PR #470, opened 2021-12-19
Prevent `ParentClassMismatchError`

zspencer:
Turns out, having two classes in the same file with Zeitwerk is a very
not great idea.


zspencer:
Woops we already did this :X

PR #493, opened 2022-01-06
[FIX] Make sure Markdown-generated HTML gets styled like HTML

anaulin:
**Before**
image

**After**
image


zspencer (approved):
Nice! This is much nicer!

PR #494, opened 2022-01-06
Remove background gradient, and make things less gray

anaulin:
The gradient was causing some odd issues with different contrast in different
parts of the page. And the gray was a little... gray? So we made it less gray.


zspencer (approved):
🎨 👩‍🎨 👩‍🎓

PR #557, opened 2022-02-10
Remove extraneous ItemRecord layer 🧹 

anaulin:
Remove unnecessary extra `ItemRecord` layer. This makes `Item` into a fully-fledged `ActiveRecord` model, and furniture-specific Item types can inherit directly from it.

**TODO**

* [x] figure out how to return `Payment`s from `PaymentForm`, instead of play `Item`
  * maybe look into https://github.com/jeremyevans/sequel 
* [x] finish globally renaming / unifying naming of `Item` and `ItemRecord` (look at methods, attributes, comments)
* [x] fix it


zspencer (approved):
THIS IS SO MUCH BETTER

Issue #476, opened 2021-12-23
Spaces: Blueprints


anaulin:
`Blueprints` provide `Residents` with a tailor-made set of tools and/or toys for the particular experience they're hoping to curate with their `Space`. 

For example, a "Book Club" `Blueprint` could be used to create a `Space` with a single `Room` which has a:
- `MembershipForm` which asserts the purpose of the club, collects contact information of potential members, and confirms Potential Members will abide by its agreements.
- `GatheringPost` which shows non-Members upcoming and past events, collects and confirms RSVPs, etc.
- `MediaBox` to record the club's history and Recommendations.

While a "Garden" `Blueprint` could be used to create a `Space` with several rooms, with `Furniture` that's more focused on curating one's online presence.

`Blueprints` can be used to create a new `Space`, or can be applied to an existing `Space`.

- [ ] Create a `Space` from a `Blueprint`
  - [ ] Creates the `Space`
  - [ ] Creates `Utilities`
  - [ ] Creates `Rooms`
  - [ ] Creates `Furniture` within the `Space`
- [ ] Applying a `Blueprint` to a `Space`
  - [ ] Adds `Rooms` to the `Space`
  - [ ] Adds `Utilities` to the `Space`
  - [ ] Does not remove or overwrite existing `Utilities`
  - [ ] Adds `Furniture` to matching `Rooms`
  - [ ] Does not remove or overwrite existing `Furniture`

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #568, opened 2022-02-15
Sketch out the initial feature file for Space Blueprints

zspencer:
See: https://github.com/zinc-collective/convene/issues/476

Ana had noticed that our Blueprint's overwrite stuff they probably
shouldn't; and I had noticed something similar previously as well. I had
initially tossed Blueprints into the codebase as an affordance for
developers to quickly create a Space with initial data.

However, I'm beginning to think that it makes sense to be a bit more
intentional with them, and am promoting them to a sub-feature of the
Space feature.


anaulin (approved):
🦭 👍🏼

PR #570, opened 2022-02-15
Closing the VideoBridge does not exit a Room

zspencer:
Early in Convene's development, we were thinking it was _primarily_ a video/audio chat platform.

However, as we've evolved with the "what do we actually have capacity for/find interesting" the VideoBridge has become a piece of Furniture in the broader Convene Ecosystem.

This:
- Sprouts a Feature definition for the VideoBridge furniture
- Renames everything to `VideoBridge` or `video_bridge` as appropriate
- Consolidates the asset files for the VideoBridge furniture into `app/furniture/video_bridge`
- Makes it so closing the VideoBridge doesn't navigate the browser.


user512 (approved):
I like it, but can you let me know where was the code that closing the video bridge kick a user out of the room?

app/javascript/controllers/video_room_controller.js
```diff
@@ -1,22 +0,0 @@
...
-export default class extends Controller {
-  static targets = [ "wrapper" ]
-
-  connect() {
-    this.videoRoom = new VideoRoom(this.data.get('videoHost'), this.wrapperTarget);
-    this.videoRoom.addEventListener('enteredRoom', () => {
-      this.wrapperTarget.classList.add('active-room');
-    });
-    this.videoRoom.addEventListener('exitedRoom', () => {
-      window.location.replace(this.data.get('spacePath'));
```
zspencer, line 13: @user512 - Here's where we were moving.

PR #574, opened 2022-02-17
Set up LocaleApp for content translations

zspencer:
OK so I don't want to merge this unless @anaulin  + @user512 are down for it; but basically because Convene is a public project we get free i18n translation hosting; but we would still need to do the actual translation.

The main benefit is folks can translate the application without needing to use Github; which could be cool!


zspencer:
Closing this for now, was just using it as a dry-run

PR #576, opened 2022-02-17
Use just Turbo, not UJS

zspencer:
While UJS and Turbo are compatible, they serve similar roles and we
don't (intentionally) use UJS anywhere.


user512 (approved):

zspencer:
Deprecated in favor of #579

PR #579, opened 2022-02-20
Rails-ujs is probably safe to bump to 7

zspencer:
In https://github.com/zinc-collective/convene/pull/495, we thought we
didn't want to jump from 6 to 7 without bumping Rails first.

We also thought that maybe we didn't even need UJS, because we have no
use-cases where we are explicitely using it.

However, after removing UJS
https://github.com/zinc-collective/convene/pull/576 we discovered that
we _do_ use it for signing people out!

We did a review of the changelog and the commits between 6.X and 7.0 for
ujs, and they seemed computationally meaningless; so we think it's safe
to just bump and defer the deletion.

Co-authored-by: Neer Malathapa <nirmalathapa@users.noreply.github.com>
Co-authored-by: Naomi Quinones <naomiquinones@users.noreply.github.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@hotwired/turbo-rails": "^7.1.1",
     "@rails/actioncable": "^7.0.2",
     "@rails/activestorage": "^7.0.2",
-    "@rails/ujs": "^6.1.4",
+    "@rails/ujs": "^7.0.2",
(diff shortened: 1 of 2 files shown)


anaulin:
Thank you for digging into this, everyone! 😻

PR #580, opened 2022-02-20
Invitations include name _or_ email of the Invitor

zspencer:
See: https://github.com/zinc-collective/convene/issues/117#issuecomment-962519714

@anaulin had noticed that the invitation emails were left dangling after
"Invited by..." when the Invitor doesn't have a name.

We've pulled out a `display_name` method for the Person, and delegate it
explicitely in the `Invitation` as the `invitor_display_name`.

This gives us a seam for making more intelligent decisions about how we
present peoples names/emails/identities/etc in the application, but does
not exercise that seam to aggressively.

### DIFF
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -26,9 +26,9 @@ class Invitation < ApplicationRecord
     status.to_sym == :accepted
   end
 
-  def invitor_display_name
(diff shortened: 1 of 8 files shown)


user512 (approved):
👍

PR #592, opened 2022-03-01
Also destroy Person's AuthenticationMethods and SpaceMemberships

zspencer:
I was doing some poking at the data on my machine, deleted a person, and
discovered that some of their PII and Context sticks around when it
probably shouldn't.

Now when a Person is destroyed, their Space Membershipsand
Authentication Methods do as well.


anaulin (approved):
👍🏼

PR #594, opened 2022-03-06
Invitations: Redesign Invitations

zspencer:
There were some strange behaviors with how Invitations work, in particular:

- We never confirmed that Guests had access to the email address; we just assumed if they had the link they were totally the right person. That's not a good idea; so we made it necessary for them to confirm their access to the email address by using a OneTimePassword
- Neighbors couldn't accept Invitations at all; which seemed not-great.
- Once an Invitation was accepted, the user  was just kinda left hanging.

This resolves those rough edges, as well as fleshes out the full feature test suite for inviting people to a Space.

- [x] Define the Scenario for inviting a Neighbor to become a Member of a different Space more precisely
- [x] Adjust the test framework so that we can use parameterized space and space member names
  - [x] Create AuthenticationMethod API endpoint (and merge it independently!) (#630)
  - [x] Fix Spaces API endpoint, because it doesn't return any data!
  - [x] ~Create Person API Endpoint (and merge it independently!) [See Review](https://github.com/zinc-collective/convene/pull/630#discussion_r844568812)~
  - [x] Create SpaceMembership API Endpoint (and merge it independently!)
  - [x] Figure out how to uniquify the Actor email / Space Name + Slug on Create
  - [x] Figure out how to share Spaces/Actors/etc. between Steps in the same Scenario
- [x] Get it Green!
  - [x] Tidy up the UI
- [x] Refactor
  - [x] ~~Stop piggybacking on the AuthenticationMethod email, when it's an Invitation Email (which has a different endpoint)~~
  - [x] ~~Pull the `this.actors[..]` warehouse into the `actor` parameter type~~
  - [x] ~~Set `last_authentication_method_at` on create of AuthenticationMethod~~

### DIFF
--- a/features/steps/space_steps.js
+++ b/features/steps/space_steps.js
@@ -4,6 +4,9 @@ const { linkParameters, Actor, Space } = require("../lib");
 const appUrl = require("../lib/appUrl");
 const { Api } = require("../lib/Api");
 const AuthenticationMethod = require("../lib/AuthenticationMethod");
+const SpaceMembership = require("../lib/SpaceMembership");
+const crypto = require("crypto");
+
 
 Given("{a} {space}", function (_, space) {
   return this.api().spaces().create(space);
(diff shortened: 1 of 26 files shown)


features/steps/space_steps.js
```diff
@@ -3,17 +3,62 @@ const { RoomPage, SpacePage, SpaceEditPage } = require("../harness/Pages");
...
+   * @param {*} _
+   * @param {Actor} actor
+   * @returns
+   */
+  function (_, space, _, actor) {
+    this.actors = this.actors || {};
+    this.actors[actor.email] = actor;
+
+    space = this.spaces[space.name] || space;
+    actor.email = `${this.testId()}+${actor.email}`;
```
zspencer, line 42: TODO: Let's pull this out to the `actor` parameter type, so that it doesn't have to be done in the steps.

features/steps/space_steps.js
```diff
@@ -3,17 +3,62 @@ const { RoomPage, SpacePage, SpaceEditPage } = require("../harness/Pages");
...
+    const api = new Api(appUrl(), process.env.OPERATOR_API_KEY);
+    const toCreate = new AuthenticationMethod({
+      contactMethod: "email",
+      contactLocation: actor.email,
+    });
+
+    return api
+      .authenticationMethods()
```
zspencer, line 50: Let's swap this to `api.people().create(...)` and create the Person, rather than an `AuthenticationMethod`

spec/factories/authentication_methods.rb
```diff
@@ -6,5 +6,6 @@
 
     contact_method { :email }
     contact_location { person.email }
+    last_one_time_password_at { Time.now }
```
zspencer, line 9: Does this want to be set on create to the current time?

PR #601, opened 2022-03-10
Purchasing a Space: First Step

zspencer:
We've decided that we'll use an embedded AirTable form, which we'll make look nice and work well.

TODO:

- [x] ~`url_field` needs styling since it's not an `input[type="text"]`~
- [x] ~Why the heck are forms so huge? Do we want to give an affordance for people to set the high when they're plugging in the form?~
- [x] ~Create a pending space from the submitted form~
- [x] ~Accept the pre-payment?~


anaulin (approved):
💪🏼 💰

zspencer:
We've pulled the EmbeddedForm furniture into it's own PR, and this feature-definition is probably ok to merge on it's own.

PR #615, opened 2022-03-23
Prototype: Themes!

zspencer:
We had talked briefly about how, if we want to  move momentpark.com off SquareSpace and into Convene, we probably needed a way to give it different colors / typography / layout, so I made it possible for a Space to set a theme, which will then prioritize loading views from the `app/themes/#{space.theme}` directory.

Introducing "Desert Dunes", a totally unique theme! 
Screen Shot 2022-03-22 at 8 08 18 PM

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,3 +1,4 @@
 //= link_tree ../images
 //= link_tree ../builds
 //= link application.js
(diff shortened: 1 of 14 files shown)


zspencer:
Now with _moar fancy
Screen Shot 2022-03-22 at 20 38 35
_

PR #618, opened 2022-03-24
Only load Jitsi in the VideoBridge

zspencer:


anaulin (approved):
Very nice. Thanks for the cleanup! 🧹 🙏🏼

PR #625, opened 2022-03-31
Upgrade to latest Rails patch version

anaulin:
A failed attempt at getting rid of some warnings about re-defined constants. Instead, we just get a slightly fresher Rails.


zspencer (approved):
🥳 🥳

PR #630, opened 2022-04-03
Sprout API for Creating Authentication Methods

zspencer:
The AuthenticationMethod API wants to be used in the feature tests to
make it a little bit easier to create People and authenticate as them
programmatically.

This pulls the work done in
https://github.com/zinc-collective/convene/pull/594 for exposing that
API out into it's own PR that we can merge independently, then rebase
into the feature branch for Invitations.

Co-authored-by: Neer Malathapa <nirmalathapa@users.noreply.github.com>
Co-authored-by: Naomi Quinones <naomiquinones@users.noreply.github.com>

### DIFF
--- a/app/controllers/authentication_methods_controller.rb
+++ b/app/controllers/authentication_methods_controller.rb
@@ -4,6 +4,7 @@
 class AuthenticationMethodsController < ApplicationController
   def create
     skip_policy_scope
+    authentication_method_params = policy(AuthenticationMethod).permit(params)
     authentication_method = AuthenticationMethod.new(authentication_method_params)
     authorize(authentication_method)
     if authentication_method.save
@@ -11,9 +12,5 @@ class AuthenticationMethodsController < ApplicationController
     else
       render json: AuthenticationMethod::Serializer.new(authentication_method).to_json, status: :unprocessable_entity
     end
-  end
-
-  def authentication_method_params
-    params.require(:authentication_method).permit(:contact_method, :contact_location)
   end
 end
--- a/app/controllers/authentication_methods_controller.rb
+++ b/app/controllers/authentication_methods_controller.rb
@@ -6,14 +6,33 @@ class AuthenticationMethodsController < ApplicationController
     skip_policy_scope
     authentication_method = AuthenticationMethod.new(authentication_method_params)
     authorize(authentication_method)
-    authentication_method.save!
+    if authentication_method.save
+      render json: authentication_method_as_json(authentication_method)
+    else
+      render json: render_errors(authentication_method), status: :unprocessable_entity
+    end
+  end
 
-    render json: {
-      id: authentication_method.id,
-      contact_method: authentication_method.contact_method,
-      contact_location: authentication_method.contact_location,
-      person: {
-        id: authentication_method.person.id
+  private def render_errors(model)
+    {
+      errors: [{
+        title: t('.failure'),
+        detail: model.errors.full_messages.join('.')
+      }]
+    }
+  end
+
+  # @todo Maybe we want to try?
+  #       http://jsonapi-rb.org/guides/serialization/defining.html
+  private def authentication_method_as_json(authentication_method)
+    {
+      authentication_method: {
+        id: authentication_method.id,
+        contact_method: authentication_method.contact_method,
+        contact_location: authentication_method.contact_location,
+        person: {
+          id: authentication_method.person.id
+        }
       }
     }
   end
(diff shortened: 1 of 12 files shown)


spec/requests/authentication_methods_request_spec.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe '/authentication_methods/', type: :request do
+  include ActiveJob::TestHelper
```
anaulin, line 6: Is this really needed? 🤔

app/controllers/authentication_methods_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    authentication_method = AuthenticationMethod.new(authentication_method_params)
+    authorize(authentication_method)
+    if authentication_method.save
+      render json: AuthenticationMethod::Serializer.new(authentication_method).to_json
+    else
+      render json: AuthenticationMethod::Serializer.new(authentication_method).to_json, status: :unprocessable_entity
+    end
+  end
+
+  def authentication_method_params
```
anaulin, line 16: Should this be in a `private` section?

anaulin: Or move it into the policy?

app/models/authentication_method.rb
```diff
@@ -20,6 +24,8 @@ def set_one_time_password_secret
     self.one_time_password_secret ||= ROTP::Base32.random
   end
 
+  # @todo This is dangerous because we're creating people incidentally
+  #       instead of intentionally...
```
anaulin, line 28: Yeah, this is something to think about. I definitely find it counter-intuitive that in order to create a new `Person` that can log in we are creating an `AuthenticationMethod`. It seems like we would want to know which `Person` is associated with an `AuthenticationMethod`?

It feels more natural that we start with the `Person`. We could have a `Person` API that accepts nested attributes for an `AuthenticationMethod`, and have minimal required attributes for the `Person` (a name? something else?).

anaulin: If we go that way, we probably want to remove the `optional: true` on line 6, and enforce that an `AuthenticationMethod` is always associated with a `Person`.

config/locales/authentication_method/en.yml
```diff
@@ -0,0 +1,4 @@
+en:
+  authentication_methods:
+    create:
+      failure: "Authentication Method could not be saved. 🙀"
```
anaulin, line 4: I feel like I belong here. 😻

anaulin (approved):
This looks good and is a great step towards both making our feature tests more powerful and robust, and also I love that we are starting to have a lil' API. 🌱 

I left some comments below about the philosophy of going from `AuthenticationMethod` to `Person` or vice-versa. Personally I like the idea of giving primacy to `Person` over other things, but that's just me. If y'all want to go with the current `AuthenticationMethod`-first way, that's fine by me as well.

app/models/authentication_method/serializer.rb
```diff
@@ -0,0 +1,37 @@
+# @todo Maybe we want to try?
+# http://jsonapi-rb.org/guides/serialization/defining.html
```
anaulin, line 2: I think we 💯 want to follow the JSONAPI spec if/when we are going to have a Proper API. I don't want to gate this PR on adjusting to the JSONAPI spec, though, especially since for now this is an affordance for our tests.

But if y'all feel moved to include JSONAPI in this PR, then 👏🏼 👏🏼 .

app/models/authentication_method.rb
zspencer: That's a great point! I think for this one, we're going to:

1. leave Person as optional
2. Not fix the `set_person` just yet (but flag it as deprecated and remove it eventually)
3. Probably create a Person endpoint as well, that _does_ accept the Nested Attributes, so that our Tests can create a Person right-off the bat without making two API calls.

The reason we're going "backwards" is because we want to support people "inviting" someone who is not yet a Registered Person in Convene; and once the person sign in with that authentication method, we bind it to the Person who we create on Sign In.

app/models/authentication_method/serializer.rb
zspencer: Yea, I think that's too much for this PR as well, as we want to get it merged in the next 30m

spec/requests/authentication_methods_request_spec.rb
zspencer: Oh yea, probably not... lol.

app/controllers/authentication_methods_controller.rb
zspencer: Done!

config/locales/authentication_method/en.yml
zspencer: Because u do!

Issue #632, opened 2022-04-06
🥔🗜️✨ `Gizmo`: `Livestream`


zspencer:
- https://github.com/zinc-collective/convene/issues/709

- Embeds a Video from a Livestream Provider, such as Twitch, in #1155

### Use Cases
- [x] Streaming from Twitch
- [x] Streaming from OwnCast

### Watering Hole

> Hoping for a federated, secure platform to share content with our band's fans. VOD, livestreaming, merchandise sales would be ideal, as would free and paid subscriptions.
> - [JB](https://airtable.com/appI2Bdc9NBhWOfTM/tbllaq3XwtiHUfNcF/viwJihQgUU8fCGYvr/recePpmhJRkBdxhbK)

rosschapman:
Appears to be implemented. 🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #633, opened 2022-04-06
Twitch Livestream may be embedded in a Room

zspencer:
See: https://github.com/zinc-collective/convene/issues/632

This sprouts a new Livestream Furniture, and makes it available to
Rooms.

### DIFF
--- a/app/furniture/livestream.rb
+++ b/app/furniture/livestream.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# Renders a Twitch Livestream in a Room
+class Livestream
+  include Placeable
+  def channel=(channel)
+    settings['channel'] = channel
+  end
+
+  def channel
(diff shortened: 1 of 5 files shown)


app/furniture/livestream.rb
```diff
@@ -0,0 +1,35 @@
...
+  def layout=(layout)
+    settings['layout'] = layout
+  end
+
+  def layout
+    settings.fetch('layout', '')
+  end
+
+  # @todo can we make it so we don't need to define this?
+  # and the `settings.fetch` bits?
```
anaulin, line 31: 💯  maybe a bit of metaprogramming in our future

anaulin (approved):
🎥 👍🏼 

There is something with the size of the embed yet again that wants tweaking:
image

But we can address that in a follow-up.

PR #640, opened 2022-04-14
Fix handling of FurniturePlacement slot order

anaulin:
Fixes a bug we noticed while updating furniture in production.


zspencer (approved):
Yay! I am glad we found something that does this for us, because oh god it's a hard problem.

PR #644, opened 2022-04-17
Spaces: Use Spaces API to create a Space in Scenarios

zspencer:
This swaps our scenarios to append test data to the local database, rather than re-use the `system-test` space over and over and over again.

It runs safely on CI, but if you run the test suite twice you start to see errors.

I believe this is because we're not applying a the `testId` to the Invitation Email, so when an invitation goes out it goes out to the same email address each time; which can result in race-conditions when following links.

It may be worth tilting at that windmill a bit more before merging; but I'm not sure.

Thoughts?

### DIFF
--- a/features/lib/Api.js
+++ b/features/lib/Api.js
@@ -1,38 +1,57 @@
-const axios = require('axios');
+const axios = require("axios");
+const { camelCase } = require('change-case');
+const applyCaseMiddleware = require("axios-case-converter").default;
+const Space = require("./Space");
 class Api {
   constructor(host, apiKey) {
     this.host = host;
     this.apiKey = apiKey;
-    this.axios  = axios.create({
-        baseURL: host
-      });
+    this.axios = applyCaseMiddleware(
+      axios.create({
+        baseURL: host,
+      })
+    );
 
     // Alter defaults after instance has been created
-    this.axios.defaults.headers.common['Authorization'] = `Token token="${apiKey}"`;
-    this.axios.defaults.headers.common['Content-Type'] = 'application/json';
-    this.axios.defaults.headers.common['Accept'] = 'application/json';
+    this.axios.defaults.headers.common[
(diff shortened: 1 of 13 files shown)


features/support/env.js
```diff
@@ -14,6 +15,10 @@ class CustomWorld {
   constructor() {
     this.driver = driver;
   }
+
+  testId() {
```
zspencer, line 19: There may already be something like this in cucumber, but I didn't take the time to find out :X

app/models/space/serializer.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# Serializes {Space}s for programmatic consumption
+class Space::Serializer < ApplicationSerializer
+  # @return [Space]
+  alias space resource
+
+  def to_json(*_args)
```
anaulin, line 8: Feels like it is time to start introducing https://github.com/jsonapi-serializer/jsonapi-serializer (or similar), making this all a little better. Maybe in a follow-up PR.

features/lib/Api.js
```diff
@@ -1,38 +1,57 @@
-const axios = require('axios');
+const axios = require("axios");
+const { camelCase } = require('change-case');
+const applyCaseMiddleware = require("axios-case-converter").default;
```
anaulin, line 3: Maybe run an autoformatter to get all these quotes consistent.

anaulin (approved):
I'm fine with getting this merged in even if right now there are issues running the suite twice. Hopefully we will address that sooner rather than later.

features/lib/Api.js
zspencer: Done!

app/models/space/serializer.rb
zspencer: Concur.

PR #647, opened 2022-04-19
Creating a Space via the API doesn't Assume the SystemTest Blueprint

zspencer:
As I was working on making it possible to create arbitrary Spaces via
Cucumber in https://github.com/zinc-collective/convene/pull/644, I ran
into a bump because the API assumes you want to always create a
SystemTest space.

But that's still way too big of a step to do in a single PR, so I pulled
out a smaller step towards that larger goal.

Right now, the SpacesController#create is more of a nupsert than a true
Create, in that it:

1. Finds or Creates the Space based upon it's name
2. But if it _does_ exist, it does nothing. Hence Nupsert. Kinda like a
   no-op upsert.


app/controllers/spaces_controller.rb
```diff
@@ -17,20 +17,19 @@ def update
...
 
   def create
-    @space = Space.new(space_params)
     skip_policy_scope
-    authorize(@space)
-    Blueprint.new(client:
-      { name: space_params[:name],
-        space: SystemTestSpace::DEFAULT_SPACE_CONFIG.merge(space_params).with_indifferent_access }).find_or_create!
+    client = Client.find_or_create_by(name: space_params[:name])
+    authorize(client)
```
anaulin, line 22: FWIW, `authorize` returns the thing that it authorizes, so you can use it for chaining and assigns, so you could do:
```ruby
client = authorize(Client.find_or_create_by(name: space_params[:name]))
```

app/models/space.rb
```diff
@@ -61,4 +61,13 @@ def jitsi_meet_domain
 
     Utilities.from_utility_hookup(jitsi_hookup).meet_domain
   end
+
+  attr_accessor :blueprint
+  after_create :apply_blueprint, if: -> { blueprint.present? }
+
+  def apply_blueprint
```
anaulin, line 68: To summarize our discussion during ensemble: let's move this into a factory method like `Space.create_with_blueprint`, then adjust the controller action to use this new method when a blueprint is provided.

anaulin (approved):
🚢 🚢 🚢

PR #653, opened 2022-04-22
Collect responsibility for applying a Space::Blueprint

zspencer:
OK, so I agree that the callback system was probably a little fraught.

Complexity emerges at those seams.

It may make sense for us to pull the responsibility into a factory.

Now if we want to create a space without executing a pile of callbacks,
we can.

And if we want to keep the controllers focused on the secure translation
of the HTTP protocol into domain objects we can.

And it sides steps a `Service` class (which I find reminiscent of the
`Manager` classes of my JavaDays.

### DIFF
--- a/app/controllers/spaces_controller.rb
+++ b/app/controllers/spaces_controller.rb
@@ -20,12 +20,7 @@ class SpacesController < ApplicationController
     skip_policy_scope
     authorize(Client)
     authorize(Space)
-
-    if space_params[:blueprint].present?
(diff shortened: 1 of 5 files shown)


anaulin (approved):
Nice, thanks for the clean-up! 🧹 🥦 💚

anaulin (approved):
Tests _are_ 🆒 ! This is the happiest I've ever been! 🎉

PR #655, opened 2022-04-22
Spaces: API Serializes response Data

zspencer:
I also went ahead and added rswag so that we can generate API
documentation directly from our rspec tests

https://user-images.githubusercontent.com/50284/164836924-4179bcea-99e4-4c4a-9649-9fdd41c0e6be.mov

.

### DIFF
--- a/app/controllers/spaces_controller.rb
+++ b/app/controllers/spaces_controller.rb
@@ -20,8 +20,14 @@ class SpacesController < ApplicationController
     skip_policy_scope
     authorize(Client)
     authorize(Space)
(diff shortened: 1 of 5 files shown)


anaulin (commented):
Just at a glance, it looks like the diff in this PR is a subset of the changes in https://github.com/zinc-collective/convene/pull/644. What am I missing?

zspencer:
Yea, this was pulled out of #644 but Github isn't picking up the rebase-cascade quite right. I'll merge this then rebase 644, since it's likely we'll be working from that on Sunday.

PR #671, opened 2022-05-08
Sprout API for PUT /spaces/:space_id/rooms/:room_id

zspencer:


app/models/furniture_placement/serializer.rb
```diff
@@ -0,0 +1,12 @@
+class FurniturePlacement::Serializer < ApplicationSerializer
```
anaulin, line 1: missing `frozen_string_literal: true` bit

anaulin (approved):
🎉 🌮 👍🏼

PR #675, opened 2022-05-11
Furniture: Spotlight!

anaulin:
**TODO**

- [x] fix broken tests (they rely on System Test being a certain shape, which we changed)
- [x] Rename to Spotlight!
- [x] Fix Words to be Less Silly
- [x] add tests
- [x] test this with S3 (Ideally in CI?)
- [x] (maybe) add feature definition
- [x] Rename from `image_file` to `image`


app/furniture/spotlight/image.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Spotlight
+  class Image < Item
+    has_one_attached :file
```
zspencer, line 5: It would be nice if this was namespaced to the space so it is easier to see what is owned by the space

zspencer (approved):
I luv it.

app/furniture/spotlight/image.rb
anaulin: We did some research, and this is not possible to do with ActiveStorage out of the box, so we decided this is not worth it for now.

PR #690, opened 2022-05-26
Add GitHub Action to automatically deploy to Heroku

anaulin:
This is to replace the no-longer-working Heroku<->GitHub autodeploy.

For this to work, I'll need to set the `HEROKU_API_KEY` secret under Settings -> Secrets (I do not have access to this section of the settings at the moment).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,18 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - master
+
+# Uses https://github.com/AkhileshNS/heroku-deploy to deploy to Heroku
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,18 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - master
```
zspencer, line 6: This isn't the branch we deploy!

zspencer (changes requested):
FINALLY! A valid reason to request changes!

.github/workflows/deploy.yml
anaulin: Fixed. PTAL. 🙏🏼

zspencer (approved):
🚢 🇮🇹

PR #691, opened 2022-05-26
Update Rails to fix latest security vulnerabilities

anaulin:
Fixes:
https://github.com/zinc-collective/convene/security/dependabot/26
https://github.com/zinc-collective/convene/security/dependabot/25


zspencer (approved):
It's strange that we had to remove the `RAILS_ENV=test` that is part of the autogenerated rails setup script; but alas.

PR #695, opened 2022-05-29
Always place Furniture inside a TurboFrame

zspencer:
This may be a bit risky, but the intention here is to make all Furniture
live inside a Turbo Frame, so the "modify in-line" behavior "just works"


anaulin (approved):
A little magick-y, but seems like a relatively safe experiment. LGTM! 🧙🏼

PR #698, opened 2022-05-29
Furniture: Spotlight: Test automatin'

zspencer:
- [x] Loosen up the device-specific rendering checks (out of scope for now)
- [x] Get rid of extra parameterTypes that are likely unnecessary now
- [ ] Send all the attributes being passed into the Furniture step over the wire to the server
  - [x] Get at least the non-file ones sent across the wire
  - [ ]  Figure out how to do that with files and shit :X.
- [ ] ???


anaulin (commented):
Looking good so far! 🥦

PR #715, opened 2022-06-12
Invitations: Test authenticated and adversarial Invitations

zspencer:

This demostrates that Authenticated people can successfully respond to invitations, as well as proves that nefarious actors cannot respond to invitations to other people.

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -2,7 +2,7 @@
   <%= render partial: "breadcrumbs" %>
 
 
-  <nav aria-label="Menu" class="profile-menu">
+  <nav aria-label="Menu" class="profile-menu" data-person-email="<%= current_person.email %>">
     <%- if current_person.authenticated? %>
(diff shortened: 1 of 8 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= render partial: "breadcrumbs" %>
 
 
-  <nav aria-label="Menu" class="profile-menu">
+  <nav aria-label="Menu" class="profile-menu" data-person-email="<%= current_person.email %>">
```
anaulin, line 5: Is `current_person` guaranteed to be non-nil? What is this set to if the person is not logged in?

anaulin (approved):
Thank you for adding more thorough tests for this critical path. 😻

I only have one clarification question below, just making sure this works when the user is logged out. Otherwise LGTM!

app/views/layouts/_header.html.erb
zspencer: It returns an instance of the `Guest` class, which has a nil email.

PR #732, opened 2022-06-23
Rooms: Respect Furniture Placements' Ranks

zspencer:
Furniture Placement's have a Slot, but the Slot doesn't get respected in ordering.


anaulin (approved):
🎖️

PR #733, opened 2022-06-23
Convert our JavaScript to ESM

anaulin:
ESM is the future!


zspencer (approved):
Wow, we lost 300 lines of JS!

PR #754, opened 2022-07-14
Use Active Record `store_accessor` to add accessors to `data` key/value pairs

anaulin:
This PR replaces our hand-made JSONB accessors in Payment with the Active Record `store_accessor` method.


zspencer (approved):
Neat, I haven't seen `store_accessor` before!

anaulin:
@zspencer I only learned about it earlier this week and I thought it would be useful for our use case here!

PR #761, opened 2022-07-24
Furniture: Marketplace

zspencer:

- [ ] Tidying
  - [x] Pull everything that isn't specifically about the Marketplace Furniture into tiny PRs that can be merged independently.
    - [x] `FurnitureController` changes
    - [x] `money` gem
    - [x] `button` CSS specificity fixes
    - [x] `email_field` partial
- [ ] Sketch `Checkout` flow
- [ ] Rename `Marketplace::Item` to `Marketplace::LineItem`
- [ ] Persist the `Marketplace::Order`
- [ ] Persist `Marketplace::LineItems` in the `Marketplace::Order`?

### Maybe?
- [ ] Sprout `Marketplace::Stall` for segmenting `Marketplace::Product` based upon who is selling?


features/furniture/marketplace.feature.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Scenario: Place a Delivery Order
+- When a Guest places a Delivery Marketplace Order in the "Piikup Marketplace" Space for:
+  | item                 | quantity | price |
+  | 1lb of Bananas       | 4        | $5.56 |
+  | 32oz of Greek Yogurt | 1        | $7.99 |
+  | Delivery by Piikup   | 1        | $6.99 |
+- Then the Guest is charged $20.54 for their Marketplace Order
+- And the Marketplace Vendor "Piikup" receives the Marketplace Order placed by the Guest in the "Piikup Marketplace" Space
+- And the Marketplace Vendor "Piikup" receives a Payment of $6.99
```
zspencer, line 27: I'm not sure we want Piikup to be a Marketplace Vendor; or if we want them to be the Stripe account holder. I think the latter? 

I've been reading more about Stripe Connect (since I haven't implemented it yet) and I think we want to try and get it to work with [Direct Charges](https://stripe.com/docs/connect/charges#types), because that means we can [add the delivery fee as an "application fee"](https://stripe.com/docs/connect/direct-charges#collecting-fees) rather than trying to [separate the charges](https://stripe.com/docs/connect/charges-transfers)

This puts the payment directly into the Stripe account of the seller, less the stripe platform fees (2.9% + .50c?) and the Application Fee; and makes the seller responsible for refunds, etc; which seems like the lightest weight way to do this.

PR #762, opened 2022-07-24
Ensemble/ignore invitations  117

KellyAH:
Zee, Naomi and I fixed the test for unignoring an invitation and fixed some of the messaging for the unignore button.


anaulin (approved):
Thank you all for the fix!

PR #783, opened 2022-08-07
Invitations: Revoking an Invitation

zspencer:

This makes it possible for a Space Owner to Revoke an Invitation that was sent; and prevents people from accepting revoked invitations.

We also took the time to remove some stuff that was getting a little crufty, as well as ran our autoformatter (since we don't have it happening in CI yet...)


app/views/spaces/edit.html.erb
```diff
@@ -46,6 +46,9 @@
             &lt;<%= invitation.email %>&gt;
             <%= invitation.last_sent_at %>
             (<%= invitation.status %>)
+            <%- if invitation.pending? %>
+              <%= button_to t('icons.remove'), [space, invitation], title: 'Revoke', method: :delete, class: 'no-underline' %>
```
anaulin, line 50: Title should be i18n-ed.

features/spaces/invitations.feature
```diff
@@ -54,11 +54,15 @@ Feature: Spaces: Invitations
...
-    And no further Invitations can be sent to "aang-the-avatar@example.com" for the "System Test" Space
+  @unimplemented-steps @unstarted
+  Scenario: Revoking a Pending Invitation
+    Given an Invitation to "invitee-a@example.com" for the "System Test" Space is sent by Space Owner "space-owner@example.com"
+    When the Invitation to "invitee-a@example.com" for the "System Test" Space is Revoked by Space Owner "space-owner@example.com"
+    Then the Invitation to "invitee-a@example.com" for the "System Test" Space has a status of "revoked"
+    Then the Invitation to "invitee-a@example.com" for the "System Test" Space cannot be Accepted
+
+  @unimplemented-steps @unstarted
+  Scenario: Revoking Accepted Invitations
```
anaulin, line 65: This is technically built, by hiding the "revoke" button unless the invitation is "pending". So maybe update the tags on this scenario and maybe even tweak the scenario description to clarify that this works by hiding the controls.

anaulin (approved):
Thank you for adding this! Minor comments below.

app/views/spaces/edit.html.erb
KellyAH: Thanks fixed in b097307

features/spaces/invitations.feature
KellyAH: Thanks fixed in 467c4fa

PR #791, opened 2022-08-11
Add a sequential number to Factory's Space name to avoid failures due to duplicated names

anaulin:
Test run https://github.com/zinc-collective/convene/runs/7779315597?check_suite_focus=true failed due to the factory trying to make a Space with a duplicated name. This PR adds a sequential number to the Space Name to avoid such collisions.

Bonus: replaced FFaker with Faker, which is identical and better maintained.


zspencer (approved):
I love it! Thank you!

PR #798, opened 2022-08-21
Address feedback for revoking invitations #783

KellyAH:
Thanks @anaulin 😄 
https://github.com/zinc-collective/convene/pull/783

- marks test Revoking Accepted Invitations as built. 
- translates revoking of invitations button


zspencer:
Thanks @KellyAH, @nirmalathapa and @naomiquinones!!!

PR #806, opened 2022-08-25
Fix Failing Deployments

zspencer:
Initially, I thought the issue was caused by Heroku's recent changes to pg_extras; but maybe it's because values with the enum were still in the database?

Anyway, this fixes it!?

https://devcenter.heroku.com/changelog-items/2446
https://blog.testdouble.com/posts/2022-08-15-migrating-postgres-extensions-to-the-heroku_ext_schema/

### DIFF
--- a/db/migrate/20220807184623_remove_sent_from_invitation_status_enum.rb
+++ b/db/migrate/20220807184623_remove_sent_from_invitation_status_enum.rb
@@ -3,7 +3,7 @@ class RemoveSentFromInvitationStatusEnum < ActiveRecord::Migration[7.0]
     reversible do |direction|
       direction.up do
         execute "UPDATE invitations SET status='pending' WHERE status='sent'"
-        remove_enum_value :invitation_status, 'sent'
+        # remove_enum_value :invitation_status, 'sent'
       end
       direction.down do
         add_enum_value :invitation_status, 'sent'
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -22,6 +22,7 @@ ActiveRecord::Schema[7.0].define(version: 2022_08_07_184623) do
     "expired",
     "ignored",
     "revoked",
+    "sent",
   ], force: :cascade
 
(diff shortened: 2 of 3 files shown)


db/migrate/20220807184623_remove_sent_from_invitation_status_enum.rb
```diff
@@ -3,7 +3,7 @@ def change
     reversible do |direction|
       direction.up do
         execute "UPDATE invitations SET status='pending' WHERE status='sent'"
-        remove_enum_value :invitation_status, 'sent'
+        # remove_enum_value :invitation_status, 'sent'
```
anaulin, line 6: Huh. I assume this migration had already run in on prod and this enum value exists there.

Would it make sense to also comment out the down-path, just for cleanliness sake?

And did you get a sense if we're going to have a problem with these enum migrations in the future? Or did you run some commands on the Heroku Postures instance to make them play nice going forward?

anaulin (approved):
Thanks for fixing this.

Below questions are non-blocking and just for my own education. I trust that you've tested this.

db/migrate/20220807184623_remove_sent_from_invitation_status_enum.rb
zspencer: Technicaly, I ran this migration against production, so that the enum value isn't getting removed anymore :'( ... so now this PR doesn't need to get merged and everything will keep working...

But it's probably worth fixing the actual drift; I just didn't have any enthusiasm for it beyond unfucking the deploy and wanted to move on to something fun.

db/schema.rb
```diff
@@ -22,6 +22,7 @@
     "expired",
     "ignored",
     "revoked",
+    "sent",
```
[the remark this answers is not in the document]
zspencer: Nope, this is because the column is no longer being "removed" so it gets added back into the schema 🤦

zspencer:
I mean I `git push --force heroku HEAD:main` which is *kinda* like testing :X

PR #813, opened 2022-09-01
Rename Space Memberships to just Memberships

anaulin:
Because the "Space" is redundant.


zspencer (approved):
`:%s/space_membership/membership/g`

PR #835, opened 2022-09-18
Invitations UI Cleanup Pass

zspencer:

- [x] Split Invitations to it's own page
- [x] Improve presentation of collection of Invitations
- [x] Remove "Tab Group" UI for nested navigation; so that we stay on single level as much as possible
- [x] Request spec for creating invalid invitations
- [x] Tidy up flow when creating invitations
  - [x] Send people to the Invitations page, not Memberships
  - [x] Don't discard the users input when can't save an invitation
- [x] Don't show Invitation Revoke button if Invitation already Revoked
- [x] Fix Failing Tests
  - [x] Rspec
  - [x] Adjust PageObjects 
  - [x] WHY ARE CUKES FAILING?!
- [x] ~~Extract "Table" component?~~
Screen Shot 2022-09-18 at 12 39 05 PM
Screen Shot 2022-09-18 at 12 38 59 PM
Screen Shot 2022-09-18 at 12 38 52 PM
?!


app/views/invitations/_invitation.html.erb
```diff
@@ -1,9 +1,33 @@
...
-  <%= invitation.name %>
-  &lt;<%= invitation.email %>&gt;
-  <%= invitation.last_sent_at %>
-  (<%= invitation.status %>)
-  <%- if invitation.pending? && policy(invitation).destroy? %>
-    <%= render "buttons/remove", href: [space, invitation], title: t('invitation.destroy') %>
-  <%- end %>
-</li>
\ No newline at end of file
+
```
anaulin, line 1: Extra blank line.</high-value comment>

app/views/invitations/index.html.erb
```diff
@@ -1,16 +1,43 @@
-<fieldset>
+<% breadcrumb :invitations, space %>
```
anaulin, line 1: Maybe set this in the controller, instead of the view? But we set it in the view elsewhere, so it makes sense to put it here for consistency. Hrmph. 🤔

app/views/memberships/_membership.html.erb
```diff
@@ -22,7 +22,13 @@
     </time>
     <br>
     <% if membership.invitation.present? %>
-        Invited by: <%= membership.invitation.invitor.name %>
+      Invited by <%= membership.invitation.invitor.name %>
+      on
+      <br />
```
anaulin, line 27: I think in 2022 we no longer self-close <br> tags? Or is it the other way around? Not that it matters functionally. Only matters OCD-ally.

anaulin (approved):
Very nice cleanup, y'all! 💅🏼

Inconsequential and non-blocking comments below.

PR #843, opened 2022-09-22
Fix test flakiness caused by poor path construction

anaulin:
Also tweaked the Space factory to stop creating bogus slugs.


zspencer (approved):
Love it!

PR #850, opened 2022-09-29
Switch back to the mainline rails-spec gem

anaulin:
I set us to use the release candidate for the `rspec-rails` version that support Rails 7.x. (Looks like `rspec-rails` are in the process of changing their versioning scheme, so there isn't a non-RC version that fits us for now.)


zspencer (approved):
Love it!

Issue #201, opened 2021-02-18
🌫️ `Gizmo`: `Spotlight`


zspencer:
## Use Cases
- [ ] `Spotlight` with a Title and Description...
  - [ ] a `Room`
  - [ ] a Movie
    - [ ]  With Captions
   - [ ] an Image
     - [ ] With Alt-Text https://usability.yale.edu/web-accessibility/articles/images
  - [ ] an Arbitrary Website 
    - [ ] With a Image
    - [ ] Prepopulates Content via Open Group Protocol - https://ogp.me
- [ ] As a...
  - [ ] Hero, which is full-width on screen
  - [ ] Card Collection

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #852, opened 2022-09-29
Spotlight accessability improvements

sadiejay:
Part of https://github.com/zinc-collective/convene/issues/201

Tried to add alt-text as an attribute to spotlight, but it's not working. Please help


zspencer:
Hey @sadiejay! I've created a patch into your branch with what fixes it. Apparently we still have polishing to do on how creating Furniture works... The symbol vs string keys thing should not be something a `Furniture` developer really should need to worry about; so @anaulin and I will do some pondering on how to tidy that up!

anaulin (approved):
Nice improvement, thank you!

PR #863, opened 2022-10-06
Strip out VScode Workspace Settings

zspencer:
It doesn't seem like something that's useful to share; especially since it benefits from being tailored on a per-device basis.

It also seems like it gets in the way of people using whatever linting setup works best for them on their machine; since in order for Solargraph to work for linting, you need to disable vscode Ruby's linter; which works fine with rubocop.

We're also removing Solargraph from the recommend extensions since it seems to not really work well on most contributors computers.


anaulin (approved):
Thanks for cleaning up. Too bad this didn't work better.

PR #864, opened 2022-10-08
installs cypress

puffenstein:


puffenstein:
This was meant to go to my forked repo. I'm experimenting with cypress and @KellyAH suggested that Convene app would be a great app to practice writing cypress tests against. I guess it decided to go to the main repo instead! Oopsie! 😬

Issue #898, opened 2022-10-21
Furniture: Journal


zspencer:
It would be nice to be able to publish short and long-form multi-media content to a Room.

- [ ] Adding a Journal to a Space
- [ ] Publishing a Journal Entry
- [ ] Subscribing to a Journal
  - [ ] Via Email
  - [ ] Via ActivityPub
  - [ ] Via RSS
- [ ] Adding a Distribution Channel
  - [ ] Twitter
  - [ ] PRNewsWire?
  - [ ] GoodReads?
- [ ] Distributing Published Journal Entries
  - [ ] To Subscribers
  - [ ] To Distribution Channels

PR #899, opened 2022-10-23
Memberships: Adds Breadcrumb from Membership: Show page

zspencer:
I noticed when working on the
[Journal](https://github.com/zinc-collective/convene/issues/898) that when you click on the `Account` link, it kind of drops you into the middle of no where.

This adds a breadcrumb, so we can find our way back! Yay!

## Before
Screen Shot 2022-10-23 at 1 39 29 PM

## After!
Screen Shot 2022-10-23 at 1 39 20 PM

### DIFF
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -15,6 +15,10 @@ class Membership < ApplicationRecord
 
   validates :member, uniqueness: { scope: :space_id }
 
+  def member_name
+    member.display_name
+  end
+
   enum status: {
     active: 'active',
     revoked: 'revoked'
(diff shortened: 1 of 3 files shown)


KellyAH:
Does breadcrumbs need any test coverage added? Or is it such a core feature of Rails that it's not going to break/is tested in other areas?

zspencer:
I could imagine a world where it's useful to write unit tests for crumbs, especially for ones with conditional logic but I didn't want to spend time figuring out how to do it in a useful way beyond reloading the page.

anaulin (approved):
Nice, thank you! 🥖

PR #900, opened 2022-10-23
Journal: Initial MVP

zspencer:
See: [Furniture: Journal](https://github.com/zinc-collective/convene/issues/898)

I wanted to explore what it would look like to do a pretty standard Rails-only No TurboLinks piece of Furniture; and this is what I got!

https://user-images.githubusercontent.com/50284/197422802-fdab1d85-c446-4efd-a9b5-54c8656cb464.mp4


features/furniture/journal.feature.md
```diff
@@ -0,0 +1,28 @@
+# Feature: Journal
+
+Many people maintain an online journal. Often, these journals are intended to connect with a broader collection of people a one-on-one or small group message.
```
KellyAH, line 3: typo?
```suggestion
Many people maintain an online journal. Often, these journals are intended to connect with a broader collection of people via one-on-one or small group messages.
```

KellyAH:
If this "standard Rails-only No TurboLinks" way of implementing furniture is faster/easier to build/maintain/debug, that's great. But what risks do we take or features do we lose by not using TurboLinks?

The UI/UX for this feature still could use some improvements in a future iteration. Will that be possible with this current implementation?

zspencer:
re: no Hotwire - Theoretically, the way Rails apps are supposed to work is you:

1. Build it without using Hotwire/Turbo, so that everything works without JavaScript
2. Layer in turbo frames/streams as appropriate
3. Profit!

So it should be relatively straightforward to layer in hotwire to improve the UX.

app/furniture/journal/breadcrumbs.rb
```diff
@@ -0,0 +1,24 @@
...
+    parent :room, entry.room
+    link 'Add a Journal Entry', [:new, entry.space, entry.room]
+  end
+
+  crumb :edit_journal_entry do |entry|
+    parent :journal_entry, entry
+    link 'Edit', [:edit, entry.space, entry.room, entry]
+  end
+rescue NoMethodError => e
+  throw e unless e.message.start_with?("undefined method `crumb'")
```
anaulin, line 21: What's this about? In which cases would we be throwing an `undefined method crumb` error, and why is that ok to swallow?

app/furniture/journal/room.rb
```diff
@@ -0,0 +1,7 @@
+class Journal::Room < Room
+  has_many :journal_entries, inverse_of: :room, class_name: 'Journal::Entry'
```
anaulin, line 2: Is the only purposes of this new Room class to have the `:journal_entries` association?

I don't love it. It seems confusing to me, overloaded with our Room-Room, and causing us to have bits of code like `room.becomes(Room)...`.

If the main purpose of this is to load the `entries` in `Journal` (without adding a `has_many` to our proper Room), I'd prefer we removed this class and instead implemented `entries` with an ActiveRecord lookup, like `Entry.where(room_id: room).recent`.

anaulin (approved):
I have a couple of comments below, the one about `Journal::Room` more important to me than the other one. However, neither of these two is blocking for merging this -- feel free to ignore or not the comments and then merge when you're ready.

app/furniture/journal/room.rb
zspencer: Ya, this gets wiped in the fancier version of the PR.

app/furniture/journal/breadcrumbs.rb
zspencer: OK this is HORRIBLE and I *HATE IT* but apparently Gretel's way of re-loading  breadcrumbs uses an `instance_eval`; but if gets autoloaded by Rails and then it explodes.

zspencer:
Moved to draft because I wanna do it like we did the Marketplace.

app/furniture/journal/breadcrumbs.rb
zspencer: I fixed it by adding breadcrumbs to Zeitwerks ignore directive!

PR #901, opened 2022-10-24
Journal: But Designed like the Marketplace

zspencer:
I wanted to see what would happen if I made the way the Furniture works a little bit more like Marketplace. Next steps are probably to pull out the shared bits into a PR that can be merged independently so the design improvements in the Marketplace branch aren't stuck there until we merge.

PR #862, opened 2022-10-06
WIP: Working on buyer adding to cart feature

sadiejay:
Found a seam where we can start adding products to cart

Co-authored-by:  Kelly <KellyAH@users.noreply.github.com>
Co-authored-by:  Zee <zspencer@users.noreply.github.com>
Co-authored-by:  Ana <anaulin@users.noreply.github.com>

PR #902, opened 2022-10-24
Furniture: Extract seam to ease the elimination of Placeable

zspencer:
OK, so the trajectory I am going with is we can get rid of the `Furniture`, `FurniturePlacement` and `Placeable` classes and the weird things they do in favor of standard-ass inheritance. I expect that most of the ways those pieces fit within the system will mostly coalesce into the `Furniture` class; but I'm not quite ready to make that leap just yet.

This will let us go through the `MarkdownTextBlock`, `Spotlight`, `VideoBridge`, and `PaymentForm` Furniture and cut their LoC by quite a bit; as well as get us into a much smoother groove when extending those features.

The refactor seemed to serve both the `Marketplace` and the `Journal` feature when we did [the first in an
Ensemble](https://github.com/zinc-collective/convene/pull/862/files#diff-d2a16da4475658ee74f5ff09ab97db7f2ec836b017ae760f4cc13687b150b656R34-R35) and the second [in my own
branch](https://github.com/zinc-collective/convene/pull/901/files)

### DIFF
--- a/app/furniture/furniture.rb
+++ b/app/furniture/furniture.rb
@@ -31,10 +31,7 @@ module Furniture
 
   # @return [Placeable]
   def self.from_placement(placement)
-    REGISTRY[placement.furniture_kind.to_sym].new(placement: placement)
-  end
-
-  def self.use_relative_model_naming?
-    true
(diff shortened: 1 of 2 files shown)


anaulin (approved):
🎉

PR #905, opened 2022-10-27
Turn on an explicit locale backend for the Money gem, to get rid of a deprecation message

anaulin:
As suggested in the instructions at https://github.com/zinc-collective/convene/pull/new/au/fix-deprecation


zspencer (approved):
LOVE IT!

PR #915, opened 2022-11-03
Remove UJS, which we don't use, and which argues with turbo.

anaulin:


zspencer (approved):
Thank you this is literally the happiest I have ever been

Issue #831, opened 2022-09-15
🥔🗜️✨ `Gizmo`: `Marketplace`


zspencer:
- https://github.com/zinc-collective/convene/issues/709

Community eCommerce for collective liberation.

### Live Examples

- https://marketplace.piikup.com - Primarily Catering, Grocers, and Restaurants

_Wish this was further along? Consider sponsoring our [Zinc Community Fund](https://opencollective.com/zinc-community), which we use to cover our operating costs and pay contributors!_

### Client Owners:
- https://github.com/zinc-collective/convene/issues/832 - [Example ](https://www.jotform.com/app/221485534343152)
- https://github.com/zinc-collective/convene/issues/888 - [Priorities](https://artisans.coop/blog/marketplace-priorities-poll-results/)
- https://github.com/zinc-collective/convene/issues/889
- https://github.com/zinc-collective/convene/issues/1572

### [Fauxsonas](https://www.usability.gov/how-to-and-tools/methods/personas.html)
  - [ ] #2044
  - [ ] #2043
  - [ ] #2045

### Use Cases & features
- [x] #1641
- [x] #1324
- [x] #1326
- [x] #1331
- [x] #1325
- [x] #1327
- [x] https://github.com/zinc-collective/convene/issues/1137
- [ ] #1330
- [ ] #1238

### Works With
- [x] #1002 
- [ ] #1501

### Design Library
- [ ] #2220

### Comparison

| Product | Fees | Funding Model | Corporate Structure | License | Sells on | 
|---------|------|----------------|----------------------|--------|--------|
| [Convene: Marketplace](https://convene.zinc.coop/marketplace) | Your CoOps Commercial Member Fee + [~3% Stripe Fees](https://stripe.com/pricing) | Bootstrapped | Worker and Client Cooperative | CoOp Source | Your CoOps Domain, Your Domain or supported [Cross-Channels](https://github.com/zinc-collective/convene/issues/1446)|
| [Etsy](https://www.etsy.com/) | [9.5%~25%](https://www.etsy.com/sell) | Venture Backed, Public Company | For-Profit Corporation | Proprietary | Etsy Only |
| [Shopify](https://www.shopify.com/) | [$39/mo+ ~3% Credit Card Fees](https://www.shopify.com/pricing) | Venture Backed, Public Company | For-Profit Corporation | Proprietary | Shopify or Your Domain |
[...]

PR #927, opened 2022-11-10
Marketplace: Cart Total Updates as Products are Added and Removed

zspencer:


anaulin (approved):
Look at those totals!!1! 📈

PR #934, opened 2022-11-12
Add money field with tailwind stylings

nirmalathapa:
Added money input field with dollar sign in front of the money and the end of the number with @zspencer to see with the tailwind component. For now it is always dollar sign as a hard coded string. 

input-filed

### DIFF
--- a/app/furniture/marketplace/products/_form.html.erb
+++ b/app/furniture/marketplace/products/_form.html.erb
@@ -2,5 +2,6 @@
   <%= render "text_field", { attribute: :name, form: f} %>
   <%= render "text_area", { attribute: :description, form: f} %>
   <%= render "number_field", { attribute: :price_cents, form: f} %>
+  <%= render "money_field", { attribute: :amount, form: f} %>
   <%= f.submit %>
 <% end %>
--- a/app/furniture/marketplace/products/_form.html.erb
+++ b/app/furniture/marketplace/products/_form.html.erb
@@ -1,7 +1,6 @@
 <%= form_with model: product.location do |f| %>
   <%= render "text_field", { attribute: :name, form: f} %>
   <%= render "text_area", { attribute: :description, form: f} %>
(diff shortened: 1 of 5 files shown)


app/furniture/marketplace/products/_form.html.erb
```diff
@@ -2,5 +2,6 @@
   <%= render "text_field", { attribute: :name, form: f} %>
   <%= render "text_area", { attribute: :description, form: f} %>
   <%= render "number_field", { attribute: :price_cents, form: f} %>
+  <%= render "money_field", { attribute: :amount, form: f} %>
```
zspencer, line 5: Nice! I'm assuming that right now, setting this value doesn't get saved to the database?

If you want to try to get it to flow through (i.e. get saved, then rendered on the menu screen), there's a few places to look to make changes:

1. The `Marketplace::Product` model class definition should tell you the name of the attribute that stores the product's price
2. The `Marketplace::ProductPolicy` Policy class definition should expose a set of `permitted_attributes` (or something like that), which is how we allow attributes to be written by HTTP requests.
3. Then we'll want to make sure that when we submit the new field it is being written as the full price, rather than the price in cents; which may require some poking and reading of how the `money-rails` gem works!

Once that's working, we'll want to get rid of the `... "number_field", {attribute: :price_cents, ...` line and then we can merge this and through a cash-money-party!

zspencer (commented):
🎉 this is great I love it! Soooo good!

app/furniture/marketplace/products/_form.html.erb
nirmalathapa: Thanks @zspencer :) Yes you are right it's a dumb UI for now. 

I think I am a bit lost in between the number step 1. and step 2. may be we also want a loading text while saving the data in the database as a part of the user flow? I think we could continue from this PR today probably if not there is other stuff?

zspencer (approved):
Yesssss!!!

PR #955, opened 2022-11-27
Marketplace: Destroy Products when Destroying Marketplaces

zspencer:
See: https://github.com/zinc-collective/convene/issues/831

I keep forgetting that foreign keys need to be taken care of explicitly
in relations. I'm pretty sure there's a gem that notices mismatches
between dependent settings and database schemas but I don't remember
what it is :X.


anaulin (approved):
~~@dependabot squash and merge~~

Sorry, wrong button. I meant to say LGTM. 👍🏼

PR #974, opened 2022-12-07
Tidy up Architecture Diagramt

zspencer:
Attempting to consolidate some of the over-detailed bits, as well as expose some of the cooler stuff (like, each Space stores their own Assets but shares a Postgresql Database, VM Cluster, and Redis Queue


anaulin (approved):
It's so pretty. 🥹

Thanks so much for updating this! (If someone else wanted to make more updates to this diagram, is the login+password for whatever app you use for this shared? Or is there some other way to build on the work you've done here?)

In the spirit of inclining our language towards human-friendly plain English, I suggest the following non-blocking edits:
* "Furniture provides the core value-propositions for a Space, and is the main entrance point for interactivity between people and data" => "Furniture provides the core functionality for a Space, and is the main entrance point for interaction between people and data."
* "Utilities allow Spaces or Furniture to interact with external vendors." => s/vendors/services/ (or maybe APIs?)
* "Spaces are how we gather people and data together towards a common purpose." => get rid of the "together", which seems redundant with "gather" and implied by "common purpose"?

zspencer:
The password is currently not shared, but  it can be! And I like those edits!

PR #976, opened 2022-12-08
Explicitly skip_authorization in Memberships index action

anaulin:
Another small fix after we changed our system to require us to explicitly authorize or scope data in Actions.


zspencer (approved):
Thank you!

KellyAH (approved):

PR #977, opened 2022-12-08
Marketplace: Cart attach to Shopper

KellyAH:

When there were multi Marketplaces in a neighborhood, the page would crash. Now the Cart is attached to a Shopper which prevents the conflict of trying to find the Cart by the same Id across Marketplaces.

### DIFF
--- a/app/furniture/marketplace/marketplaces/_marketplace.html.erb
+++ b/app/furniture/marketplace/marketplaces/_marketplace.html.erb
@@ -1,5 +1,5 @@
 
-<%- cart = marketplace.carts.find_or_create_by(id: session[:current_cart] ||= SecureRandom.uuid) %>
+<%- cart = marketplace.carts.find_or_create_by(shopper: current_person) %>
 
 <%= render cart %>
 
--- a/app/furniture/marketplace/marketplaces/_marketplace.html.erb
+++ b/app/furniture/marketplace/marketplaces/_marketplace.html.erb
@@ -1,5 +1,13 @@
 
-<%- cart = marketplace.carts.find_or_create_by(shopper: current_person) %>
+
+
+<%- shopper = if current_person.is_a?(Guest) 
+     Marketplace::Shopper.find_or_create_by(id: session[:current_cart] ||= SecureRandom.uuid)
+  else
+    Marketplace::Shopper.find_or_create_by(person: current_person)
+end %>
+
+<%- cart = marketplace.carts.find_or_create_by(shopper: shopper) %>
 
 <%= render cart %>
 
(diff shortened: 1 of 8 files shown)


app/furniture/marketplace/marketplaces/_marketplace.html.erb
```diff
@@ -1,5 +1,13 @@
 
-<%- cart = marketplace.carts.find_or_create_by(id: session[:current_cart] ||= SecureRandom.uuid) %>
+
+
+<%- shopper = if current_person.is_a?(Guest) 
+     Marketplace::Shopper.find_or_create_by(id: session[:current_cart] ||= SecureRandom.uuid)
+  else
+    Marketplace::Shopper.find_or_create_by(person: current_person)
+end %>
```
anaulin, line 8: This is fine for now, but it feels like this might want to eventually live in a controller helper method, instead of the template (and maybe also the cart?). It just seems like a fair amount of non-trivial logic to have in a view.

db/migrate/20221208013832_add_shopper_to_marketplace_carts.rb
```diff
@@ -0,0 +1,12 @@
+class AddShopperToMarketplaceCarts < ActiveRecord::Migration[7.0]
+  def change
+    create_table :marketplace_shoppers, id: :uuid  do |t|
+      t.references(:person, type: :uuid, foreign_key: {to_table: :people}, index: {unique: true})
```
anaulin, line 4: Does this `unique: true` mean that we intend there to only ever be at most one Shopper linked to a Person, among all Marketplaces? Or is this also scoped by Marketplace, and I'm just not seeing it?

anaulin (approved):
LGTM 🚢

app/furniture/marketplace/marketplaces/_marketplace.html.erb
KellyAH: I tried moving this to a helper in `app/furniture/marketplace/marketplaces_controller.rb`  but it forced me to repeat the partial `_marketplace.html.erb` which is used in at least 3 controllers. This would lead to a lot of duplicate code. 

@anaulin @zspencer  Should I just move it to application helper like below or leave the code as in `_marketplace.html.erb`.

```
module ApplicationHelper
  def shopper
    params = if current_person.is_a?(Guest) 
      {id: session[:current_cart] ||= SecureRandom.uuid}
    else
      {person: current_person}
    end

    Marketplace::Shopper.find_or_create_by(params)
  end
end

```

zspencer: Putting it in the `ApplicationHelper` would take over the name `shopper` in all execution contexts; which seems less than ideal.

WHAT IF!

We [generated a `MarketplaceComponent`](https://viewcomponent.org/guide/generators.html) that takes an instance of a `Marketplace::Marketplace` and exposes a `shopper` method? It could access `current_person` and `session` through the [Helpers Proxy](https://viewcomponent.org/guide/helpers.html#proxy)

PR #994, opened 2022-12-15
Marketplace: Sprout Checkout

KellyAH:
Continuing building out checkout flow in https://github.com/zinc-collective/convene/issues/831

# TODO

- [ ] add [request specs](spec/furniture/marketplace/checkouts_controller_request_spec.rb)
- [ ] add checkout form to [checkout page](app/furniture/marketplace/checkouts/new.html.erb)
  - [ ]  - save inputted shopper address
  - [x] 🌱  - display carts items and quantities & total
- [ ] save checkout when checkout form submits


KellyAH:
Oops I forgot to assign ya'll as reviewers after opening this PR on Wednesday. I think we wanted to get it merged ASAP right? Or are we waiting to continue it on Sunday?

zspencer:
Well, we could merge it as is; since it seems like the tests are passing, but I would prefer to wait to merge until we showed the `Cart`, it's `Products` and Total on the `CheckoutsController#new` page; rather than a blank page.

I don't think we need to get the actual "Create a Checkout" bit working pre-merge, but some tidying of the flow may be nice on Sunday!

zspencer (approved):
Nice! This is good enough to merge! WOOO!!!!

app/furniture/marketplace/cart_products_controller.rb
```diff
@@ -24,6 +24,7 @@ def create
             turbo_stream.replace("cart-product-#{cart_product.product_id}", cart_product),
             turbo_stream.replace("cart-footer-#{cart.id}",
               partial: "marketplace/carts/footer", locals: {cart: cart}),
+              turbo_stream.replace("cart-total-#{cart.id}", partial: "marketplace/carts/total", locals: {cart: cart})
```
zspencer, line 27: I notice we make this particular change 3 different times in this controller; what if we gathered these three replace calls into a method so they aren't duplicated everywhere? What arguments would the method need to take to work appropriately across it's different calls?

app/furniture/marketplace/carts/_footer.html.erb
```diff
@@ -1,9 +1,16 @@
+<% 
+  marketplace = cart.marketplace
+  room = marketplace.room
+  space = room.space
+%>
+
 <tfoot id="cart-footer-<%= cart.id%>" class="bg-gray-50">
   <tr>
     <td></td>
+    <th> <%= link_to("checkout", new_space_room_marketplace_checkout_path(space, room, marketplace), html_options = {}) %> </th>
```
zspencer, line 10: We may be able to use `[:new, space, room, marketplace, :checkout]` in place of `new_space_room_marketplace_checkout_path(space, room, marketplace)`

zspencer (approved):

PR #998, opened 2022-12-18
Journal: Entries with @handle@example.com link to https://example.com/@handle

zspencer:
https://github.com/zinc-collective/convene/issues/898

There's probably a better way to do this that takes into account that handles may not use the same format on every fediverse instance; but we'll cross that bridge when we come to it.

### DIFF
--- a/spec/furniture/journal/entry_spec.rb
+++ b/spec/furniture/journal/entry_spec.rb
@@ -1,0 +1,18 @@
+require "rails_helper"
+
+RSpec.describe Journal::Entry do
+  subject(:entry) { build(:journal_entry, body: body) }
+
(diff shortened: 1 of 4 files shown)


spec/furniture/journal/entry_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+RSpec.describe Journal::Entry do
+  subject(:entry) { build(:journal_entry, body: body) }
+
+  let(:body) { Faker::Books::CultureSeries.civ }
```
anaulin, line 6: I don't think you need this `body` declaration, it is never used (the one on line 12 is used instead).

zspencer: True!

Issue #74, opened 2020-08-03
🥔✨ `Space`: `Domain`s


zspencer:
While `Space`s in a neighborhood can be accessed at their neighborhood address (i.e. https://neighborhood.example.com/spaces/space-handle); Commercial Clients expect People would visit a Space via a domain that represents their organization. 

## Use Cases
- [x] 🥔✨ Visiting a `Space` via a `Domain`
- [x] #1513
- [ ] 🌫️ Registering a `Domain`
- [ ] 🌫️ Emailing from a `Domain`

## Operator Experience
- [ ] Automate allocation of the `Space` Domain with the `Neighborhood` Infrastructure
  - [ ] DNS Provider (Cloudflare)
  - [ ] Web Host (Heroku)
  - [ ] Transactional Email Service?

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #1000, opened 2022-12-18
Space: BYO Domains drop Space from polymorphic_urls

zspencer:

This had been bugging me for a bit, where the paths on the page would always include the spaces, even if they were being routed to a custom domain.

Now, we send folks to the domain that the space claims and drop the Space from the path, but only on urls built using Rails Polymorphic url builders.

For example:

`space_room_path(space, room)` will still go to
`current_domain.example.com/spaces/:space_id/rooms/:room_id`, while `url_for([space, room])` or `link_to([space, room])` will go to
`space_domain.example.com/rooms/:room_id`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,32 +27,37 @@ class ApplicationController < ActionController::Base
 
   helper_method :current_person
 
+  # Overrides built in `url_for` to better support branded domains
+  # @see http://api.rubyonrails.org/classes/ActionController/Metal.html#method-i-url_for
   helper_method def url_for(options)
-    space = if options[0].is_a?(Space) && options[0].branded_domain.present?
+    space = if options[0].try(:branded_domain).present?
       options.delete_at(0)
-    elsif [:edit, :new].include?(options[0]) && options[1].is_a?(Space) && options[1].branded_domain.present?
+    elsif [:edit, :new].include?(options[0]) && options[1].try(:branded_domain).present?
       options.delete_at(1)
-    elsif options.is_a?(Space) && options.branded_domain.present?
+    elsif options.try(:branded_domain).present?
       options
     end
 
-    if space
-      if options.respond_to?(:last) && options.last.is_a?(Hash)
-        options.last[:domain] = space.branded_domain
-      elsif options.respond_to?(:<<) && options.length > 0
-        options << {domain: space.branded_domain}
-      else
-        options = [:root, {domain: space.branded_domain}]
-      end
+    return super unless space
+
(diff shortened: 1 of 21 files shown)


app/controllers/application_controller.rb
```diff
@@ -27,6 +27,38 @@ class ApplicationController < ActionController::Base
...
+  helper_method def url_for(options)
+    space = if options[0].is_a?(Space) && options[0].branded_domain.present?
+      options.delete_at(0)
+    elsif [:edit, :new].include?(options[0]) && options[1].is_a?(Space) && options[1].branded_domain.present?
+      options.delete_at(1)
+    elsif options.is_a?(Space) && options.branded_domain.present?
+      options
+    end
+
+    if space
```
anaulin, line 39: You could return early `unless space.present?`, I think it would be a tad more readable

app/controllers/application_controller.rb
```diff
@@ -27,6 +27,38 @@ class ApplicationController < ActionController::Base
 
   helper_method :current_person
 
+  helper_method def url_for(options)
```
anaulin, line 30: These two methods are arcane enough that I think they warrant at least a one-sentence comment explaining that they augment the standard Rails helpers to make URLs on branded domains nicer.

app/controllers/application_controller.rb
```diff
@@ -27,6 +27,38 @@ class ApplicationController < ActionController::Base
...
+    space = if options[0].is_a?(Space) && options[0].branded_domain.present?
+      options.delete_at(0)
+    elsif [:edit, :new].include?(options[0]) && options[1].is_a?(Space) && options[1].branded_domain.present?
+      options.delete_at(1)
+    elsif options.is_a?(Space) && options.branded_domain.present?
```
anaulin, line 35: Not sure if it is worth it, or if it would make the code even harder to understand, but these checks for "is this a Space and does it have a branded domain" could be tightened up like:
```ruby
thing_were_checking.try(:branded_domain).present?
```
Which is a little bit more flexible (will work for any object that has a `branded_domain`, even if it is not a `Space`), but I think works for our purposes.

anaulin (approved):
Thank you for improving this! ™️ 🔗

I left some minor non-blocking comments below.

app/controllers/application_controller.rb
zspencer: I always forget about `.try(:foo)`!

app/controllers/application_controller.rb
zspencer: I think I need to `return super unless space` but ya!

PR #1007, opened 2022-12-21
Spaces: Support Sub-domains in BYO Domains

zspencer:
https://github.com/zinc-collective/convene/issues/23

When I was testing #74 on https://beta.zeespencer.com/ I discovered that it didn't quite work, and was nesting sub-domains (i.e. `beta.zeespencer.com` became `beta.beta.zeespencer.com`)

This is likely because the [`domain` option for `url_for`](http://api.rubyonrails.org/classes/ActionDispatch/Routing/UrlFor.html#method-i-url_for) doesn't actually do the whole domain; just the TLD!

Womp womp!


anaulin (approved):
Thanks for testing and fixing! 🛠️

PR #1016, opened 2022-12-22
PaymentForm: Restart from Scratch!

zspencer:
We learned a lot since we started with this, in particular around how Turbo works, how more complex furniture wants to be structured, and how to integrate with Utilities.

All that said, we don't need it anymore; and when we start over we'll do it better.


anaulin (approved):
Yes, burn it with fire! I feel so light now. 🪶

PR #1022, opened 2022-12-27
Spotlight: Reset!

zspencer:
Since we're not really using it yet; and we've learned a lot since we started it  it seems easiest to toss and restart.


anaulin (approved):
Nice!

PR #1023, opened 2022-12-27
Expose a `#location` method for Railsy-Routing

zspencer:
There should be documentation and tests and such, but I smoke tested this pretty thoroughly?


anaulin (approved):
🗺️

PR #1024, opened 2022-12-28
Marketplace: MarketplacesController#edit Spec

zspencer:
Just a basic "get the route, confirm it renders a template successfully for authenticated user and is a 404 when not"


anaulin (approved):
Thank you for adding tests! 😻

PR #1027, opened 2023-01-02
Journal: Don't explode on unpublished posts

zspencer:
See: https://github.com/zinc-collective/convene/issues/1026
See: https://github.com/zinc-collective/convene/issues/898

Welp, that's what I get for not testing that case! Now there's a test, it's not great... needs some actual assertions; but it exercised the error!

### DIFF
--- a/spec/furniture/journal/entries_request_spec.rb
+++ b/spec/furniture/journal/entries_request_spec.rb
@@ -6,7 +6,17 @@ RSpec.describe Journal::EntriesController, type: :request do
   let(:room) { journal.room }
   let(:member) { create(:person, spaces: [space]) }
 
-  describe "POST /entries" do
+  describe "#index" do
+    it "shows all the journal entries" do
+      sign_in(space, member)
+      entries = create_list(:journal_entry, 3, journal: journal)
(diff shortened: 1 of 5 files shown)


spec/furniture/journal/entries_request_spec.rb
```diff
@@ -6,7 +6,17 @@
   let(:room) { journal.room }
   let(:member) { create(:person, spaces: [space]) }
 
-  describe "POST /entries" do
+  describe "#index" do
+    it "shows all the journal entries" do
+      sign_in(space, member)
+      entries = create_list(:journal_entry, 3, journal: journal)
```
anaulin, line 12: Would be nice to explicitly make sure in the test that we have both published and un-published entries and are displaying all of them. But not blocking to merge this in.

anaulin (approved):
🐛🔨

spec/furniture/journal/entries_request_spec.rb
zspencer: Good call! I'll do that then merge!

anaulin: Thank you for adding that! ✅

PR #1030, opened 2023-01-05
Make sure we compare HTML-encoded stuff to HTML-encoded stuff, to avoid test flakiness

anaulin:
Fixes a flaky test


zspencer (approved):
lolsob that's hilarious thank you

PR #1031, opened 2023-01-05
Marketplace: Accept Payments

zspencer:
https://github.com/zinc-collective/convene/issues/831

- [x] Accept Payments via Stripe Checkout
  - [x] Happy Path
    - [x] Prevent change to CartProducts Quantity when Cart is Checked Out
    - [x] Prevent destruction of CartProducts when Cart is Checked Out
    -  (Zee's note: Perhaps this is best done at the Policy level, so Sellers/Distributors may update + destroy CartProduct(s) after Checkout but Shoppers cannot?
    - [x] make sure that the current cart is correctly reset after a checkout
  - [x] Sad Path
- [x] Tidy
  - [x] Request Specs
  - [x] Model Specs
  - [x] Pull Stripe API bits out of `Marketplace::CheckoutsController` and into `Marketplace::Checkout`?

Co-authored-by: Ana <anaulin@users.noreply.github.com>
Co-authored-by: Dalton <daltonrpruitt@users.noreply.github.com>
Co-authored-by: Kelly Hong <KellyAH@users.noreply.github.com>

PR #1044, opened 2023-01-15
Update CheckoutPolicy to match spec expectations

anaulin:
(this PR is for merging into WIP https://github.com/zinc-collective/convene/pull/1031)

This PR updates the `CheckoutPolicy` implementation to satisfy the expectations in the spec. I do not know if this is precisely the behavior we want -- e.g. maybe we want a "marketplace admin" (space owner?) to be able to always see a Checkout -- but we can expand the behavior later.

This fixes the test failures in https://github.com/zinc-collective/convene/pull/1031

### DIFF
--- a/app/furniture/marketplace/checkout_policy.rb
+++ b/app/furniture/marketplace/checkout_policy.rb
@@ -11,7 +11,7 @@ class Marketplace
     end
 
     def show?
-      true
+      create?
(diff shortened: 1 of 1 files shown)


zspencer (approved):
Thank you! I think this makes a ton of sense!

PR #1045, opened 2023-01-16
Handle the "failed to initiate checkout" case and add request specs

anaulin:
(merging into WIP https://github.com/zinc-collective/convene/pull/1031/)

There are a few things that I don't exactly love about this (the weird validation issue, needing the stripe url attributes on checkout), but I think it at least moves things in the right direction.

### DIFF
--- a/app/furniture/marketplace/carts/_footer.html.erb
+++ b/app/furniture/marketplace/carts/_footer.html.erb
@@ -7,7 +7,7 @@
 <tfoot id="cart-footer-<%= cart.id%>" class="bg-gray-50">
   <tr>
     <td></td>
-    <th> <%= link_to("checkout", new_space_room_marketplace_checkout_path(space, room, marketplace), data: { turbo: false }) %> </th>
+    <th> <%= button_to("checkout", [space, room, marketplace, :checkouts], data: { turbo: false }) %> </th>
     <th scope="row" class="text-right px-1 py-3.5">Total: </th>
     <td class="text-left px-3 py-3.5 font-bold">
       <%= render "marketplace/carts/total", cart: cart %>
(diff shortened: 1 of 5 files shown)


app/furniture/marketplace/carts/_footer.html.erb
```diff
@@ -7,7 +7,7 @@
 <tfoot id="cart-footer-<%= cart.id%>" class="bg-gray-50">
   <tr>
     <td></td>
-    <th> <%= link_to("checkout", new_space_room_marketplace_checkout_path(space, room, marketplace), data: { turbo: false }) %> </th>
+    <th> <%= button_to("checkout", [space, room, marketplace, :checkouts], data: { turbo: false }) %> </th>
```
zspencer, line 10: Theoretically, `marketplace.location(:checkouts)` should return `[space, room, marketplace, :checkouts]`

zspencer (approved):
omg, thank you for pulling this out of the controller and into a model! 

Also, I really appreciate that this is now a post and not a GET!

app/furniture/marketplace/carts/_footer.html.erb
anaulin: Noted. We can update it on Wednesday.

anaulin: That doesn't seem to work. `marketplace.location(:checkouts)` is creating the path `/spaces/system-test/rooms/entrance-hall/marketplaces/000c9416-a6ff-49f0-9a55-3a86aad16c3a`, which is not right.

zspencer: Blah, I must not have finished applying`HasLocation` to the `Marketplace` namespace.

PR #1050, opened 2023-01-19
`Neighborhood`: Bump Rails version to patch CVE-2023-22794 and CVE-2023-22797

anaulin:
Addresses https://github.com/zinc-collective/convene/security/dependabot/47 and https://github.com/zinc-collective/convene/security/dependabot/44


zspencer (approved):
Thank you!

PR #1051, opened 2023-01-19
Remove Jitsi video-bridge furniture and infrastructure code

anaulin:
- [x] delete the `video_bridge` placements
- [x] delete all the associated code

This PR does _not_ remove the Vultr terraform configuration code, and the Vultr instance is still running.


zspencer (approved):
We should be able to delete all the vultr stuff, even the configuraiton, since we don't intend to keep using it and we can always recreate it when it becomes useful.

I can imagine a universe where we provide a Jitsi Videobridge furniture that lets people BYO Jitsi; but I am not attached to holding onto our current code in that regards.

anaulin:
Oops, my bad, forgot to delete the existing placements that had a `video_bridge` kind. Looks like there was a handful of them left, all in the Convene Demo space:
```
FurniturePlacement.where(furniture_kind: "video_bridge").map(&:space).map(&:name)
=> ["Convene Demo", "Convene Demo", "Convene Demo", "Convene Demo", "Convene Demo"]
```

I have now deleted all of them, and it is possible to visit https://convene-demo.zinc.coop/rooms/vivek-s-desk without having things explode.

PR #1052, opened 2023-01-19
Marketplace: Record Transfer Group in Stripe Checkout Session

zspencer:
- [x] Capture a Transfer Group in the Stripe Session


anaulin (approved):
✅

PR #1053, opened 2023-01-19
split ci pipeline so test job depends on build job

KellyAH:
## Problem:
CI pipeline currently runs as 1 job that executes everything sequentially (1. builds the app, 2. runs tests, 3. etc).

## Fix:
Break up the pipeline so after the app is built, unit, integration, and story tests run in parallel.

1. build app
2. --> run unit tests
    --> run int tests
    --> run browser tests

PR #1054, opened 2023-01-19
add names to clarify github actions

KellyAH:
Trying to make the CI yml easier to scan/understand so its easier to [break out tests into parallel jobs](https://github.com/zinc-collective/convene/pull/1053) in the near future. May be via [matrix](https://www.testmo.com/guides/github-actions-parallel-testing).

Using [this](https://boringrails.com/articles/building-a-rails-ci-pipeline-with-github-actions/) sample format yml which has handy names before each action.

It's still a bit unclear in the current yml if it includes enough steps to split out building the app in a separate job and building the test enviro to run tests in another job.

### DIFF
--- a/.github/workflows/test-convene-web.yml
+++ b/.github/workflows/test-convene-web.yml
@@ -40,74 +40,82 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+
     steps:
-    - uses: actions/checkout@v2
-    # Temporary mitigation for https://github.com/actions/runner-images/issues/6283
-    # Perhaps we should consider not using brew in CI (though it sure is nice!)
-    - uses: raviqqe/enable-homebrew@main
-    - uses: browser-actions/setup-firefox@latest
+      - name: Checkout code
+        uses: actions/checkout@v2
+      # Temporary mitigation for https://github.com/actions/runner-images/issues/6283
+      # Perhaps we should consider not using brew in CI (though it sure is nice!)
 
-    - name: Set up Ruby
-      uses: ruby/setup-ruby@v1
-      with:
-        bundler-cache: true
+      - name: Set up Homebrew environment
+        uses: Homebrew/actions/setup-homebrew@master
 
-    - name: Increase the amount of inotify watchers
-      run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+      - name: Install Firefox
+        uses: browser-actions/setup-firefox@latest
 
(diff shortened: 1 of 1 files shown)


KellyAH:
Sorry, @zspencer 😞 This isn't a huge improvement but I'm hoping it is a small step toward parallel CI test runs. Next, I'll need to deep dive more into what the yml is currently doing to see if I can break it up into separate build and test jobs.

.github/workflows/test-convene-web.yml
```diff
@@ -40,74 +40,82 @@ jobs:
...
-      with:
-        name: feature-test-failed-screenshot
-        path: features/test_reports/*.png
+      - name: Checkout code
+        uses: actions/checkout@v2
+      # Temporary mitigation for https://github.com/actions/runner-images/issues/6283
+      # Perhaps we should consider not using brew in CI (though it sure is nice!)
+
+      - name: Set up Homebrew environment
+        uses: Homebrew/actions/setup-homebrew@master
```
KellyAH, line 51: `raviqqe/enable-homebrew@main` doesn't seem safe to use anymore according to https://github.com/raviqqe/enable-homebrew#readme

.github/workflows/test-convene-web.yml
```diff
@@ -40,74 +40,82 @@ jobs:
...
+      - name: Install Firefox
+        uses: browser-actions/setup-firefox@latest
+
+      - name: Setup Ruby and install gems
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: Increase the amount of inotify watchers
+        run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
```
KellyAH, line 62: not really sure if this step is needed for the test environment or just for building the app.

.github/workflows/test-convene-web.yml
zspencer: I think the inclination to get rid of brew makes a lot of sense. It slows down the pipeline; and if we  *need* to test the `setup` script we could make that it's own job, rather than an integral part of the pipeline.

.github/workflows/test-convene-web.yml
zspencer: I don't remember either 🤦 What happens if we get rid of it?

zspencer (approved):
This is great! Thank you so much Kelly! All small steps towards a more comprehensible and parallelizable CI are ⭐️🌟⭐️!

We eat elephants a bite at a time; and this is a good bite.

PR #1057, opened 2023-01-20
Marketplace: Differentiate between Checkout and Order

zspencer:

Where `Checkout` is useful for handling a Checkout flow for the Payment Processor, it doesn't make a ton of sense from a Buyer perspective once the Cart has been Checked out.

I've sprouted an `Order` model, with corresponding Controllers, and moved the `checkouts/show` and `checkouts/_checkout` views to `orders/show` and `orders/_order`.

### DIFF
--- a/app/furniture/marketplace.rb
+++ b/app/furniture/marketplace.rb
@@ -1,15 +1,5 @@
 # @see features/furniture/marketplace.feature.md
 class Marketplace
-  def self.append_routes(router)
-    router.resources :marketplaces, only: [:show, :edit, :update], module: "marketplace" do
-      router.resources :products
-      router.resources :carts do
-        router.resources :cart_products
-      end
-      router.resources :checkouts, only: [:show, :create]
-    end
-  end
-
   def self.from_placement(placement)
     placement.becomes(Marketplace)
   end
--- a/app/furniture/marketplace/cart.rb
+++ b/app/furniture/marketplace/cart.rb
@@ -12,7 +12,7 @@ class Marketplace
     has_many :cart_products, dependent: :destroy, inverse_of: :cart
     has_many :products, through: :cart_products, inverse_of: :carts
     has_one :checkout, inverse_of: :cart
-    has_one :order, inverse_of: :cart, class_name: "Order"
+    has_one :order, inverse_of: :cart
 
     enum status: {
       pre_checkout: "pre_checkout",
--- a/app/furniture/marketplace/cart.rb
+++ b/app/furniture/marketplace/cart.rb
@@ -12,6 +12,7 @@ class Marketplace
     has_many :cart_products, dependent: :destroy, inverse_of: :cart
     has_many :products, through: :cart_products, inverse_of: :carts
     has_one :checkout, inverse_of: :cart
+    has_one :order, inverse_of: :cart, class_name: "Order"
 
     enum status: {
       pre_checkout: "pre_checkout",
--- a/app/furniture/marketplace/checkout.rb
+++ b/app/furniture/marketplace/checkout.rb
@@ -5,9 +5,10 @@ class Marketplace
     self.location_parent = :marketplace
 
     belongs_to :cart, inverse_of: :checkout
+    delegate :marketplace, to: :cart
 
     has_many :ordered_products, through: :cart, source: :cart_products, class_name: "Marketplace::OrderedProduct"
-    delegate :marketplace, to: :cart
+
     belongs_to :shopper, inverse_of: :checkouts
 
     # It would be nice to validate instead the presence of :ordered_products, but my attempts at this raise:
@@ -36,6 +37,11 @@ class Marketplace
       }, {
         api_key: marketplace.stripe_api_key
       })
+    end
+
+    def complete(stripe_session_id:)
+      update!(status: :paid, stripe_session_id: stripe_session_id)
+      cart.update!(status: :checked_out)
     end
 
     private
--- a/app/furniture/marketplace/checkouts_controller.rb
+++ b/app/furniture/marketplace/checkouts_controller.rb
@@ -4,7 +4,7 @@ class Marketplace
       authorize(checkout)
       if params[:stripe_session_id].present?
         checkout.complete(stripe_session_id: params[:stripe_session_id])
-        flash[:notice] = t('.success')
+        flash[:notice] = t(".success")
       end
       redirect_to checkout.becomes(Order).location
     end
--- a/app/furniture/marketplace/checkouts_controller.rb
+++ b/app/furniture/marketplace/checkouts_controller.rb
@@ -1,12 +1,12 @@
 class Marketplace
-  class CheckoutsController < FurnitureController
+  class CheckoutsController < Controller
     def show
       authorize(checkout)
       if params[:stripe_session_id].present?
-        checkout.update!(status: :paid, stripe_session_id: params[:stripe_session_id])
-        checkout.cart.update!(status: :checked_out)
-        flash[:notice] = t(".success")
+        checkout.complete(stripe_session_id: params[:stripe_session_id])
+        flash[:notice] = t('.success')
       end
+      redirect_to checkout.becomes(Order).location
     end
 
     def create
@@ -35,20 +35,8 @@ class Marketplace
       end
     end
 
-    helper_method def shopper
-      @shopper ||= if current_person.is_a?(Guest)
-        Shopper.find_or_create_by(id: session[:guest_shopper_id] ||= SecureRandom.uuid)
-      else
-        Shopper.find_or_create_by(person: current_person)
-      end
-    end
-
     helper_method def cart
       @cart ||= marketplace.carts.find_or_create_by(shopper: shopper, status: :pre_checkout)
-    end
-
-    helper_method def marketplace
-      @marketplace ||= policy_scope(Marketplace).find(params[:marketplace_id])
     end
   end
 end
--- a/app/furniture/marketplace/orders_controller.rb
+++ b/app/furniture/marketplace/orders_controller.rb
@@ -1,0 +1,11 @@
+class Marketplace
+  class OrdersController < Controller
+    def show
+      authorize(order)
+    end
+
+    helper_method def order
+      policy_scope(marketplace.orders).find(params[:id])
+    end
+  end
+end
--- a/app/furniture/marketplace/routes.rb
+++ b/app/furniture/marketplace/routes.rb
@@ -1,0 +1,14 @@
+class Marketplace
+  class Routes
+    def self.append_routes(router)
+      router.resources :marketplaces, only: [:show, :edit, :update], module: "marketplace" do
+        router.resources :orders, only: [:show]
+        router.resources :products
+        router.resources :carts do
+          router.resources :cart_products
+        end
+        router.resources :checkouts, only: [:show, :create]
+      end
+    end
+  end
+end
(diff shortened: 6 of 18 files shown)


app/furniture/marketplace.rb
```diff
@@ -1,15 +1,5 @@
...
 class Marketplace
-  def self.append_routes(router)
-    router.resources :marketplaces, only: [:show, :edit, :update], module: "marketplace" do
-      router.resources :products
-      router.resources :carts do
-        router.resources :cart_products
-      end
-      router.resources :checkouts, only: [:show, :create]
-    end
-  end
```
zspencer, line 11: I moved these to `furniture/marketplace/routes.rb` which I think will make it easier to find routes because you can use a fuzzy-finder and type `marketplace routes` and get the routes; instead of having to remember that they live in `furniture/marketplace.rb`

app/furniture/marketplace/checkout.rb
```diff
@@ -37,6 +38,11 @@ def create_stripe_session(success_url: , cancel_url: )
       })
     end
 
+    def complete(stripe_session_id:)
+      update!(status: :paid, stripe_session_id: stripe_session_id)
+      cart.update!(status: :checked_out)
+    end
```
zspencer, line 44: I figured shoveling this in here is probably better than leaving it in the controller; but maybe I'm wrong.

app/furniture/marketplace/checkouts_controller.rb
```diff
@@ -36,20 +35,8 @@ def create
       end
     end
 
-    helper_method def shopper
-      @shopper ||= if current_person.is_a?(Guest)
-        Shopper.find_or_create_by(id: session[:guest_shopper_id] ||= SecureRandom.uuid)
-      else
-        Shopper.find_or_create_by(person: current_person)
-      end
-    end
```
zspencer, line 45: I pulled these into `marketplace/controller.rb` since it was being used in many places.

app/furniture/marketplace/cart.rb
```diff
@@ -12,6 +12,7 @@ class Cart < ApplicationRecord
     has_many :cart_products, dependent: :destroy, inverse_of: :cart
     has_many :products, through: :cart_products, inverse_of: :carts
     has_one :checkout, inverse_of: :cart
+    has_one :order, inverse_of: :cart, class_name: "Order"
```
zspencer, line 15: I don't think the `class_name` is actually useful here 🤦

KellyAH: what happens if you remove it? what breaks?

zspencer: I don't think anything will, I just added it sillilly.

app/furniture/marketplace/checkout.rb
KellyAH: I think it's okay to leave the `complete` method here for now. It's just updating the cart record right? It's not doing anything in request/response flows -- which is what the checkouts_controller is for right?

spec/furniture/marketplace/checkouts_controller_request_spec.rb
```diff
@@ -10,9 +10,10 @@
     let(:checkout) { create(:marketplace_checkout, cart: cart, shopper: cart.shopper)}
     context "when a stripe_session_id is in the params" do
       it "marks the cart as checked out" do
-        get polymorphic_path([space, room, marketplace, checkout], { stripe_session_id: "12345" })
+        get polymorphic_path(checkout.location, { stripe_session_id: "12345" })
         expect(cart.reload).to be_checked_out
         expect(checkout.reload).to be_paid
+        expect(response).to redirect_to(checkout.becomes(Marketplace::Order).location)
```
KellyAH, line 16: The changes in this PR seem like a large enough refactor that I assumed there'd be more unit tests added/modified to provide test coverage for the new Order model, corresponding Controllers, and orders/show, orders/_order. But maybe it's just the over tester in me.

app/furniture/marketplace/checkout.rb
zspencer: Yep, it's doing the data transformations; so the controller just happens to know that showing a checkout when stripe gives it an id it also completes the checkout.

spec/furniture/marketplace/checkouts_controller_request_spec.rb
zspencer: Can you clarify where you're seeing risk here? At this point, we don't have significant deal-flow; so flexibility in mutating the design seems more valuable than enshrining the design via tests.

app/furniture/marketplace/orders_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+  class OrdersController < Controller
+    def show
+      authorize(order)
+    end
+
+    helper_method def order
+      policy_scope(marketplace.orders).find(params[:id])
+    end
+  end
+end
```
KellyAH, line 11: I am trying to visualize the way all these data models interact as the user goes thru the shopping flow is. Is it like this?

1. A customer adds items to their cart == a `Cart` of `pre_checkout` status contains `cart_products`.
2. The customer goes to the checkout page and is filling out credit card info to complete their purchase ==  a `Cart` of `pre_checkout` status is going thru the stripe checkout flow thus creating a `Checkout` record. 
3. if the checkout flow via stripe is successful, the `Cart` changes to `check_out` status, and an `Order` is created from the `cart_products`.

app/furniture/marketplace/routes.rb
```diff
@@ -0,0 +1,14 @@
...
+        router.resources :orders, only: [:show]
+        router.resources :products
+        router.resources :carts do
+          router.resources :cart_products
+        end
+        router.resources :checkouts, only: [:show, :create]
+      end
+    end
+  end
+end
```
KellyAH, line 14: why does `Marketplace` get it's own `routes.rb` file? Why not just put it in
https://github.com/zinc-collective/convene/blob/main/config/routes.rb

spec/furniture/marketplace/checkouts_controller_request_spec.rb
KellyAH: that's a good point. 👍

app/furniture/marketplace/cart.rb
KellyAH: Remove it! 🪄 ✨ 🔥 it! 😄

app/furniture/marketplace/orders_controller.rb
zspencer: So, let's decompose the datastructure a bit here:

1. at the *database* layer, a `Checkout` and  an `Order` are equivalent; they both store their attributes in the same table.
2. At the application layer, a `Checkout` and an `Order` have different routes, controllers, policies, and views.
3. At present, whether something is represented as a `Checkout` or an `Order` depends on what url is being used to retrieve the data. If the request path includes `marketplace/*/checkouts/*` then the data stored in the database will be treated as a `Checkout`. If it's `marketplace/*/orders` then the data stored in the database will be treated as an `Order`.

So, an `Order` is not 'created' in the sense that it's not actually creating any records in the database table; rather we leverage "polymorphism" (which can be loosely defined as 'the ability for one data structure to be treated as another data structure') so that we have more Domain objects for different use cases; while retaining a more normalized database.

It's entirely possible that we *do* want an `Order` to be a distinct database table from a `Checkout`, or we may not want a `Checkout` to not be a database-backed model at all! 

So, to reframe:

1. `Shopper` adds `Product`s to their `Cart`, creating a `CartProduct` record which keeps track of things like quantity. (note: I think a `Cart` is also a kind of `Order` or `Checkout`, which I think @anaulin may agree with)
2. `Shopper` creates a `Checkout`, which directs them to Stripe to collect payment  information.
3. `Shopper` completes `Checkout`, which directs them back from Stripe to Convene, and updates the `Checkout` object to indicate it had been succesfully checked-out.
4. `Shopper` is redirected to show the `Order`, which represents the `Cart` to `Shopper`s in a read-only manner.

app/furniture/marketplace/routes.rb
zspencer: A few reasons:

1. If we put in in `config/routes.rb` then every developer maintaining a piece of  `Furniture` would need to modify `routes.rb` in order to adjust the routes their `Furniture` users. Instead, we hook in to the `Furniture` class in `config/routes.rb` and tell it to iterate through every piece of furniture, and append the routes defined within the `Furniture` itself.

This allows `Furniture` pieces to "own" the code related to their furniture within the `app/furniture/furniture_name` folder; which reduces the amount of surface area a Furniture maintainer needs to cover in order to make significant changes to their `Furniture`'s behavior. 

TL/DR: Encapsulation! Every piece of Furniture owns it's entire implementation, from Routes to Models to Controllers to Policies to Views; within the `app/furniture/furniture_name` directory.

app/furniture/marketplace/checkout.rb
zspencer: I like to leave conversations open, rather than "resolving" them. Resolving feels like it dismisses the conversation! You had good questions! Other people may have them too!

spec/furniture/marketplace/checkouts_controller_request_spec.rb
zspencer: I think there are *definitely* risks here, I just don't know which ones are worth writing tests for yet! If you have ideas about which particular cases you'd like tested (either cross-request or individual request) I would be happy to implement them!

app/furniture/marketplace/cart.rb
zspencer: IT DEAD

app/furniture/marketplace/orders_controller.rb
KellyAH: wow this is great! once Marketplace is finished/finalized/stable we could make a seperate .MD doc readme with the above info and maybe a diagram ❤️

zspencer: I think it could be dropped directling into `app/furniture/marketplace/README.md` in it's own PR :D

anaulin (commented):
I don't see the need to have both a `Checkout` and an `Order` class. It seems a `Checkout` is simply an `Order` that is in the "initiated but not completed" state.

More broadly, personally I find that this style of code-reuse-via-inheritance makes codebases harder to understand and maintain. (And in general I am strongly pro composition-over-inheritance.)

All that said, I'm not going to stand in your way if you want to take things in this direction.

zspencer:
I've found in Rails codebases in particular, they become easier to maintain in a highly-rigorous domain-driven approach; which includes not having extraneous domain objects. Right now, I don't see `Checkout`, `Cart`, and `Order` as extraneous from a *domain* perspective, (as a `Cart` and an `Order` are things people reason about differently, and `Checkout` is just a ... weird abstraction) but I do think that they are extraneous from a *database* perspective. 

My hope is that:

1. I can merge the `marketplace_carts` and `marketplace_checkouts` table into a single `marketplace_orders` table; where the domain models do *not* inherit from each other, and
2. I can remove the `Checkout` class from being ActiveRecord backed; and leave it as a domain object whose only job is to handle the Stripey-bits.

Does this sound like something you can be on board with?

anaulin (approved):
My experience is that when folks insist on having a "highly-rigorous domain-driven approach" we end up with oddly convoluted structures, because of trying to shoe-horn the inevitable oddnesses of real-world needs and interactions into something that is "perfectly abstracted and modeled" instead of, say, using the occasional command pattern.

Not to talk about how highly debatable this approach can be: for example, in my opinion, "Checkout" makes no sense as a domain model for a marketplace, because it in fact represents a verb (the action of checking out) and not an noun, and thus has no business being a model.

All that said, I'm happy for you check in this or do whatever you feel is the best way to move this forward. I'm sure we'll shake out the kinks as we go along regardless.

zspencer:
I agree that we'll find the kinks and shake them out. I also agree that a Checkout is an absurd model; and I want it to be very tiny or not exist at all if we can get there.

PR #1058, opened 2023-01-20
Marketplace: Producer connects  Stripe account

zspencer:
- [x] Producer connects  a Stripe Account to Marketplace
  - [x] Add a field to store the linked stripe account id
  - [x] Request specs for StripeAccountsController#create


zspencer:
I was *not* paying attention and just force-squashed like a chump! SORRY @anaulin / @KellyAH !!!

PR #1059, opened 2023-01-20
Lint: Apparently `standard` and `standardrb` are different

zspencer:
So, I noticed a giant pile of red errors when rubocop was running; which was like... ok... and then it turned out that we were using the `standardrb` gem, which is not the same as the `standard` gem; so it was out of date...

So I updated *that* and then ran `bin/rubocop -a` which made a bajillion changes...

Computers are bad.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -109,8 +109,8 @@ group :development, :test do
 
   gem "rubocop-rails"
   gem "rubocop-rspec"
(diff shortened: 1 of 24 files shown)


KellyAH (approved):
⭐

anaulin:
Gah, thank you for catching this!

PR #1062, opened 2023-01-21
Marketplace: Refactor to turn Checkout into a Command

zspencer:

This "Demotes" `Checkout` from an Active Record and makes it serve more as a Command object than a Domain Model. It's still ActiveModel'y so it can be rendered using `render checkout` or routed to using `redirect_to checkout.location`, but otherwise it's not holding any data or what not.

This continues to allow us to leverage Rails' view-and-path-lookup ergonomics when rendering or redirecting to a `Marketplace::Cart` or a `Marketplace::Order`; but also adds a touch of duplication now that the domain models don't share a common ancestor.


app/furniture/marketplace/cart.rb
```diff
@@ -1,30 +1,27 @@
 # frozen_string_literal: true
 
 class Marketplace
-  class Cart < ApplicationRecord
+  class Cart < Record
     self.table_name = "marketplace_carts"
-    include WithinLocation
     self.location_parent = :marketplace
 
+    default_scope { where(status: :pre_checkout) }
```
zspencer, line 8: This ensures whenever we try and find a Cart it doesn't include anything that has been checked out.

app/furniture/marketplace/cart.rb
```diff
@@ -1,30 +1,27 @@
...
     belongs_to :marketplace, inverse_of: :carts
-    belongs_to :shopper, inverse_of: :carts
     delegate :space, :room, to: :marketplace
+
+    belongs_to :shopper, inverse_of: :carts
+
     has_many :cart_products, dependent: :destroy, inverse_of: :cart
     has_many :products, through: :cart_products, inverse_of: :carts
-    has_one :checkout, inverse_of: :cart
-    has_one :order, inverse_of: :cart
```
zspencer, line 15: Now we don't need these anymore, since the Checkout table is dead! DEAD!

app/furniture/marketplace/cart.rb
```diff
@@ -1,30 +1,27 @@
...
 
     enum status: {
       pre_checkout: "pre_checkout",
-      checked_out: "check_out"
+      paid: "paid"
     }
 
-    def self.model_name
-      @_model_name ||= ActiveModel::Name.new(self, ::Marketplace)
-    end
```
zspencer, line 24: I pulled this up into `Marketplace::Record` so all the `Marketplace` domain-objects don't need to define their own model-name.

app/furniture/marketplace/checkout.rb
```diff
@@ -1,53 +1,41 @@
...
         }
       }, {
         api_key: marketplace.stripe_api_key
       })
     end
 
     def complete(stripe_session_id:)
-      update!(status: :paid, stripe_session_id: stripe_session_id)
-      cart.update!(status: :checked_out)
+      cart.update!(status: :paid, stripe_session_id: stripe_session_id)
```
zspencer, line 28: No more double update!

app/furniture/marketplace/product.rb
```diff
@@ -14,6 +13,9 @@ class Product < ApplicationRecord
     has_many :cart_products, inverse_of: :product, dependent: :destroy
     has_many :carts, through: :cart_products, inverse_of: :products
 
+    has_many :ordered_products, inverse_of: :product, dependent: :destroy
```
zspencer, line 16: This makes me think that maybe we don't want to be able to destroy Products that have been Ordered!

zspencer (commented):
OK, I think this gets us closer to a more reasonable design; without the weird cart/checkout/order trifecta. 

I didn't inherit from `Cart` for `Order` and `OrderedProduct` from `CartProduct` (or vice versa ) but instead have them sharing the same database table...

zspencer:
Wooops, the tests aren't passing jus a sec.

anaulin (approved):

app/furniture/marketplace/order.rb
```diff
@@ -1,5 +1,24 @@
...
+    belongs_to :marketplace, inverse_of: :orders
+    delegate :space, :room, to: :marketplace
+
+    belongs_to :shopper, inverse_of: :orders
+
+    has_many :ordered_products, inverse_of: :order, foreign_key: :cart_id
+    has_many :products, through: :ordered_products, inverse_of: :orders
+
+    enum status: {
+      paid: "paid"
```
KellyAH, line 15: note: I can imagine more order statuses will need to be added once customers start using the feature. `refunded`, `canceled`, and maybe some of the ones listed [here](https://support.bigcommerce.com/s/article/Order-Statuses?language=en_US).

But yeah let's Keep It Simple for now.

app/furniture/marketplace/product.rb
KellyAH: so we'll always have to keep a record of any products in the database if anyone has ordered them even if the product is removed from the Marketplace/is no longer for sale?

Guess that's okay. Shouldn't take up too much database space right?

KellyAH (approved):

app/furniture/marketplace/product.rb
zspencer: Yea, it would be scary if a `Product` getting destroyed the `Order`s `OrderedProducts` because then how would `Vendor`s or `Shopper`s be able to see what they bought when?

zspencer: (Note: our hope is that Database costs are covered on a per-space basis; but I'm not entirely sure how we would go about doing that yet :D)

app/furniture/marketplace/order.rb
zspencer: Concur!

PR #1063, opened 2023-01-21
Docs: Fiddle with the ERD a bit more

zspencer:
This tightens things back up to @daltonrpruitt's initial recommendation, but also adds clustering so that objects in the same domain have the same models.


KellyAH (approved):

KellyAH:
clustering makes the ERD diagram alot easier to read. ty

Issue #1002, opened 2022-12-19
🥔🔌 `Utility`: `Stripe`


zspencer:
- https://github.com/zinc-collective/convene/issues/252

Stripe is a powerful Payment Processor that allows Furniture developers to accept payments on the behalf of themselves or space members.

## Used in Gizmos:
- https://github.com/zinc-collective/convene/issues/831

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1072, opened 2023-01-25
Stripe: Don't share the Stripe gem's namespace

zspencer:
https://github.com/zinc-collective/convene/issues/1002

This dumps a lot of complexity and brings the `UtilityHookup` a bit closer to how the `FurniturePlacement` stuff works; and opens up a pathway to rename from `UtilityHookup` to `Utility`


anaulin (approved):
Thank you for catching this!

Issue #709, opened 2022-06-09
🥔✨ `Gizmos`


zspencer:
- https://github.com/zinc-collective/convene/issues/709 are the Gain Achievers and Pain Relievers available to  https://github.com/zinc-collective/convene/issues/1475. They are placed in https://github.com/zinc-collective/convene/issues/1155

## Potatoes
_These work _just_ enough for most early adopters, but lack stability and polish..._
- [x] https://github.com/zinc-collective/convene/issues/831
- [x] https://github.com/zinc-collective/convene-journal/issues/1
- [x] https://github.com/zinc-collective/convene/issues/632
- [x] https://github.com/zinc-collective/convene/issues/200
- [x] zinc-collective/convene#1988

## Sprouts
_It's not ready for anyone but the most early of early adopters, but we've started building..._
- [ ] https://github.com/zinc-collective/tobias/issues/1

## Unplanted
_We have not even started to build..._
- [ ] https://github.com/zinc-collective/convene/issues/2150
- [ ] zinc-collective/convene#1927
- [ ] zinc-collective/convene#1928
- [x] https://github.com/zinc-collective/convene/issues/619
- [ ] https://github.com/zinc-collective/convene/issues/1094
- [ ] https://github.com/zinc-collective/convene/issues/990
- [ ] https://github.com/zinc-collective/convene/issues/989
- [ ] https://github.com/zinc-collective/convene/issues/973
- [ ] https://github.com/zinc-collective/convene/issues/881
- [ ] https://github.com/zinc-collective/convene/issues/847
- [x] https://github.com/zinc-collective/convene/issues/4
- [ ] https://github.com/zinc-collective/convene/issues/2036

## Use Cases
- [x] ✨ Adding a `Gizmo` to a `Section`
- [x] ✨ Removing a `Gizmo` from a `Section`
- [ ] ✨ Reordering a `Gizmo` in a `Section`

  
### For Developers 
- [ ] zinc-collective/convene#1735

## Bugs

- [x] https://github.com/zinc-collective/convene/issues/1565

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1074, opened 2023-01-26
Furniture: Destroying a Furniture respects `dependent:` settings

zspencer:
https://github.com/zinc-collective/convene/issues/709

When we were smoke-testing the Marketplace, I realized we had to destroy and re-create the Marketplace, but the way we delete them was... uhhh not respecting the (very prudent) foreign-keys we had added!

So now FurniturePlacements polymorph into their child implementation prior to destruction, so that all `dependent: :destroy`  etc. actions get carried out.


anaulin (approved):
Yay!

PR #1075, opened 2023-01-26
Marketplace: Collect Delivery Fee

zspencer:

We collect the delivery fee! Wooo!!!

### DIFF
--- a/app/furniture/marketplace/carts/_total.html.erb
+++ b/app/furniture/marketplace/carts/_total.html.erb
@@ -1,10 +1,10 @@
-<span id="cart-total-<%= cart.id%>" class="flex flex-col">
-  <span>
+<span id="cart-total-<%= cart.id%>" class="w-full flex flex-col">
+  <span class="text-right">
     Products:
-    <%= humanized_money_with_symbol(cart.product_total) %> +</span>
-  <span>Delivery:
-    <%= humanized_money_with_symbol(cart.delivery_fee) %> =</span>
-  <span>
+    <%= humanized_money_with_symbol(cart.product_total) %></span>
+  <span class="text-right">Delivery:
+    <%= humanized_money_with_symbol(cart.delivery_fee) %></span>
+  <span class="text-right font-bold">
     Total: <%= humanized_money_with_symbol(cart.price_total) %>
     </span>
 </span>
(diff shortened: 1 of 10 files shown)


app/furniture/marketplace/carts/_total.html.erb
```diff
@@ -1,3 +1,10 @@
...
+<span id="cart-total-<%= cart.id%>" class="flex flex-col">
+  <span>
+    Products:
+    <%= humanized_money_with_symbol(cart.product_total) %> +</span>
+  <span>Delivery:
+    <%= humanized_money_with_symbol(cart.delivery_fee) %> =</span>
+  <span>
+    Total: <%= humanized_money_with_symbol(cart.price_total) %>
+    </span>
+</span>
```
anaulin, line 10: Spantastic!

app/furniture/marketplace/checkout.rb
```diff
@@ -47,7 +47,14 @@ def stripe_line_items
...
         }
-      end
+      end + [
+        {quantity: 1,
+         price_data: {
+           currency: "USD",
+           unit_amount: marketplace.delivery_fee_cents,
+           product_data: {name: "Delivery"}
+         }}
+      ]
```
anaulin, line 57: I wonder if this would be a bit more readable with this one element first and then ` + <result of map` ?

app/models/furniture_placement.rb
```diff
@@ -42,4 +42,10 @@ def form_template
   def configurable?
     furniture.form_template != "furniture_placements/noop"
   end
+
+  def write_attribute(name, value)
+    super
+  rescue => e
```
anaulin, line 48: Hrm. Would it be possible to rescue a more specific error? ("missing attribute" or whatever?)

anaulin (approved):
Progress!

A couple of non-blocking comments with suggestions for tightening things up a bit.

app/furniture/marketplace/checkout.rb
zspencer: Good call, I think splitting it into two separate methods and joining them explicitely is even clearer!

app/models/furniture_placement.rb
zspencer: Yea, I was... YOLOing.

PR #1076, opened 2023-01-26
Marketplace: Refactor - Pull AccountLink up into Marketplace

zspencer:
The account link object was being made in the controller, and while that worked it was a mess.

This adjusts things so it's a little bit more order(ly) and places it in the model where it's "best."

### DIFF
--- a/app/furniture/marketplace/marketplace.rb
+++ b/app/furniture/marketplace/marketplace.rb
@@ -43,31 +43,14 @@ class Marketplace
         })
       end
 
-      account_link = if account.details_submitted?
-        Stripe::AccountLink.create(
-          {
-            account: stripe_account,
-            refresh_url: refresh_url,
-            return_url: return_url,
-            type: "account_update"
-          },
-          {
-            api_key: stripe_api_key
-          }
-        )
-      else
-        Stripe::AccountLink.create(
-          {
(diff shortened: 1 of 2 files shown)


app/furniture/marketplace/marketplace.rb
```diff
@@ -30,6 +30,46 @@ def delivery_fee_cents
...
+        }).tap do |account|
+          update(stripe_account: account.id)
+        end
+      else
+        Stripe::Account.retrieve(stripe_account, {
+          api_key: stripe_api_key
+        })
+      end
+
+      account_link = if account.details_submitted?
```
anaulin, line 46: Am I reading this right, that the only difference between the two if/else branches is the `type`? If so, maybe we can shorten this to something like:
```ruby
account_link = Stripe::AccountLink.create(
  {
    account: stripe_account,
    refresh_url: refresh_url,
    return_url: return_url,
    type: account.details_submitted? ? "account_update" : "account_onboarding"
  },
  {
    api_key: stripe_api_key
  }
)
```

anaulin (approved):

app/furniture/marketplace/marketplace.rb
zspencer: Good catch!

Issue #1078, opened 2023-01-30
Accepting a space invitation causes a 500


anaulin:
I went through the flow to accept a Space invitation into the demo Marketplace space, and once I entered my one-time code it threw an error:

![image](https://user-images.githubusercontent.com/6729309/215611130-6187e79c-bb18-4f96-be51-3bd885f8b564.png)

For debugging reference, the invitation was for ana@zinc.coop, invitation URL https://convene.zinc.coop/spaces/piikup-marketplace/invitations/eba2e5d2-5024-4d23-b3c6-26d3a8311d78/rsvp

zspencer:
Found it in Sentry, seems to be an intersection between [Space: Memberships](https://github.com/zinc-collective/convene/issues/117) and [BYO Domains](https://github.com/zinc-collective/convene/issues/74)

PR #1082, opened 2023-01-30
Memberships: Invitations work on BYO Domains

zspencer:
https://github.com/zinc-collective/convene/issues/1078

There's probably a bunch of other domain objects that want `parent_location` set; but this should be good enough to get the invitations flow unbroken.


anaulin (approved):
Thanks for the quick bugfix!

Looking at this change, I'm wondering how are we going to catch these "missing location" issues in the future, seems like currently this quite error prone.

zspencer:
Yea, I have a WIP PR trying to swat the bugs and would love for someone to figure out how we could run our end-to-end tests against a BYO-domain space!

PR #1084, opened 2023-01-31
`Neighborhood`: `Operators` may use Github Codespaces for contributing

daltonrpruitt:
To help facilitate easier development (and allow it online), this PR sets up devcontainer usage for convene. \
Started by following the Devcontainers extension's UI prompts, then added in other dependencies later.


zspencer:
This is very cool! How would someone go about testing or using this?

daltonrpruitt:
As far as I can tell, you should be able to click on my fork's branch, then click the `code` button, select the `codespaces` tab, then the `+` icon. I had a bunch of issues getting it started up, but reloading, restarting, reloading, ... eventually made it work. 

I just checked the one I thought I started earlier this week, and it seems to be broken. So that's cool... 🤷

daltonrpruitt:
It seems that I was really starting this wrong. Doing the + straight up just seems to start on `main`, which doesn't have devcontainer folder. Currently working on starting it up the right way.

.devcontainer/create-db-user.sql
```diff
@@ -0,0 +1,2 @@
+CREATE USER vscode CREATEDB;
+CREATE DATABASE vscode WITH OWNER vscode;
```
KellyAH, line 2: does this mean this setup only works with vscode? not other IDEs?

KellyAH: Seems like devcontainers is a pretty cool VSCode [extension](https://code.visualstudio.com/docs/devcontainers/create-dev-container). VSC just keeps getting cooler and cooler. 💎

KellyAH:
Will this eventually replace our current local dev environment setup with a better one?

.devcontainer/create-db-user.sql
daltonrpruitt: Yeah. As far as I can tell, this is a vscode-only thing. However, the dockerfile seems to not be vscode-specific, so it could probably be reused in other situations. (That said, it works pretty well with VS Code.)

daltonrpruitt:
This method of doing dev environments may work well (especially for new people to get started) because it: 

1. Is accessible online via codespaces and
2. Can be loaded/used locally in VS Code, bypassing the need to set up a local environment manually. 

However, I think that people will probably have their own ways to set up their environment in *just* the right way for them, and I don't know how persistence across local environments would work. I don't know if you can mount local files to the dev containers or if it is truly and completely isolated. If it is isolated, that could prevent people from setting things up in *just* the right way. I figured it was something that could be discussed.

zspencer:
> Will this eventually replace our current local dev environment setup with a better one?

Like Dalton, I see this as primarily a "Make it easy for folks who do not want to maintain a local dev environment"

For example, when I'm at a pub with my iPad, I currently have a Tailscale + SSH setup which allows me to hack on Convene on my home computer; but... uhhh... I would much rather be able to tap a button and create a branch with a fully operational IDE 🤣

zspencer:
Is it possible we're at a point where it's worth merging this and working out the kinks in future PRs? Or should we keep on letting it ride?

daltonrpruitt:
Good point. I think it's really close, but I'll need to find some time to do some final checks and see if it is at least usable. If I can get the whole "one-button-to-run-it-all" setup right yet, I at least want it to make sure it forwards ports correctly (which I've been led to believe it will automatically, but have yet to confirm). 

I should be able to test this tonight and tomorrow. Is that an okay timeframe?

zspencer:
I'm less concerned from a timing-of-completion perspective and asking more from a "Is this branch getting unwieldy?"  or "is this good enough that others could start to build on it if it was in `main`?" perspective; so no rush on my end as long as you're comfy with the contributor ergonomics!

daltonrpruitt:
Fair enough. It's very close to being somewhat useable, I'm just stuck on some kind of postgres issue at the moment. Once that's fixed, it should be ready for initial inclusion.

daltonrpruitt:
Geez. Well, it seems to be working now, but I'm sure we'll find some kinks. Let me know if you get to try it before merging! (You'd have to do it on the branch of my fork, though.)

zspencer (approved):
Thank you! I appreciate the breadcrumbs in the documentation and a new document specifically for how to deal with this! 

As we learn how these work, we may want to drop links to clarify bits and bobs!

PR #1091, opened 2023-02-01
Remove obsolete video_bridge placements from the Demo space

anaulin:
A left-over from https://github.com/zinc-collective/convene/pull/1051, which is causing non-working video-bridges to magically re-appear in the Demo space.


zspencer (approved):
🙈 Thank you!!!!

PR #1092, opened 2023-02-02
`Marketplace`: `Vendor` receives `Payment` upon `Order` completion

zspencer:

OK, this is the first shot across the bow for listening to stripe checkout completions and sending the money onwards to Vendors

Co-authored-by: Ana Ulin <anaulin@users.noreply.github.com>

**TODO**

* [x] add programmatic way to register a marketplace's webhook endpoint with Stripe
* [x] add signature verification (look at https://stripe.com/docs/webhooks/quickstart for an example code snippet)
* [x] add automated tests for new webhook controller

### DIFF
--- a/app/furniture/marketplace/stripe_events_controller.rb
+++ b/app/furniture/marketplace/stripe_events_controller.rb
@@ -7,7 +7,8 @@ class Marketplace
     def create
       skip_authorization
       payload = request.body.read
-      event = Stripe::Event.construct_from(params)
+      signature = request.env['HTTP_STRIPE_SIGNATURE'];
+      event = Stripe::Event.construct_from(params, signature, marketplace.stripe_webhook_endpoint_secret)
 
       case event.type
 
--- a/app/furniture/marketplace/stripe_events_controller.rb
+++ b/app/furniture/marketplace/stripe_events_controller.rb
@@ -7,8 +7,8 @@ class Marketplace
     def create
       skip_authorization
       payload = request.body.read
-      signature = request.env['HTTP_STRIPE_SIGNATURE'];
-      event = Stripe::Event.construct_from(params, signature, marketplace.stripe_webhook_endpoint_secret)
+      signature = request.env["HTTP_STRIPE_SIGNATURE"]
+      event = Stripe::Webhook.construct_event(params, signature, marketplace.stripe_webhook_endpoint_secret)
 
       case event.type
 
--- a/app/furniture/marketplace/stripe_events_controller.rb
+++ b/app/furniture/marketplace/stripe_events_controller.rb
@@ -8,7 +8,8 @@ class Marketplace
       skip_authorization
       payload = request.body.read
       signature = request.env["HTTP_STRIPE_SIGNATURE"]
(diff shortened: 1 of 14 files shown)


app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    skip_before_action :verify_authenticity_token
+
+    # @todo add signing check using the stripe webhook signing secret
+    # @see https://stripe.com/docs/webhooks/quickstart
+    def create
+      skip_authorization
+      payload = request.body.read
+      signature = request.env["HTTP_STRIPE_SIGNATURE"]
+
+      event = Stripe::Webhook.construct_event(payload, signature, marketplace.stripe_webhook_endpoint_secret)
```
zspencer, line 12: @anaulin - Whenever we are forwarding and replaying events, it hangs up on this area; even if we manually copy/paste the webhook signing secret directly during pry.

We haven't been able to figure out why it's not playing nicely, but if you have any insights it would be appreciated. We've:


1. Deleted and recreated the Webhook programmatically 
2. Manually copied in the signing key
3. Set the "Tolerance" to 999999999999 in an attempt to make it support however much clock-drift there may be between local dev and stripe servers

Thoughts?

zspencer: Turns out... the webhook_secret gets generated on a per `stripe listen` basis, so I was setting the secret to the wrong secret. WOMP WOMP WOMP.

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Marketplace
+  class StripeEventsController < Controller
+    skip_before_action :verify_authenticity_token
+
+    # @todo add signing check using the stripe webhook signing secret
```
anaulin, line 5: I think this todo is resolved now, no?

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+        payment_intent = Stripe::PaymentIntent.retrieve(event.data.object.payment_intent, {api_key: marketplace.stripe_api_key})
+        order = Order.find(payment_intent.transfer_group)
+
+        Stripe::Transfer.create({
+          amount: order.price_total.cents,
+          currency: "usd",
+          destination: marketplace.stripe_account,
+          transfer_group: order.id
+        }, {api_key: marketplace.stripe_api_key})
+      end
```
anaulin, line 25: It might be prudent to throw in here an `else` to log or raise an error, just in case we get any other types of events that we don't expect -- otherwise we'll have no way of knowing if we are ignoring a bunch of other incoming Stripe events.

app/furniture/marketplace/stripe_events_controller.rb
zspencer: It is!

anaulin (approved):
💸

Non-blocking comments ~~below~~ above 🚢

app/furniture/marketplace/stripe_events_controller.rb
zspencer: Prudent!

Issue #42, opened 2020-07-03
Rooms: Adding, Changing and Removing Rooms


user512:
## Scenarios:
- Andromeda
  - [ ] Adding Rooms
  - [ ] Changing Rooms
  - [ ] Removing Rooms
- B???+
  - [ ] Make the "Slug" Form Field show the Slug in context of the Space URL so it's clearer what a "Slug" is.

## Feature Definition

```features/rooms.feature
Feature: Rooms:
  Rooms serve to organize a Space; each Room can have Furniture within it tailored to the particular behaviors the Room is 
  intended to facilitate.

Scenario: Adding Rooms
  When a Space Member creates a new Room
  Then the Room is available to enter from the Room List

Scenario: Removing Rooms
  When a Space Member removes a Room
  Then the Room is not available to enter from the Room List
  
Scenario: Changing Rooms
  When a Space Member changes a Room
  Then the Room's information is updated in the Room List
```

## UI Mockup
![Convene - Creating a Room - UI Sketch](https://user-images.githubusercontent.com/50284/88490648-e769e980-cf51-11ea-864d-b42db22fe9a3.png)

Customer Feedback:
> Are there plans for the consumer to be able to add multiple web pages in the future?

PR #1099, opened 2023-02-05
Spaces: Rooms can be added to Spaces with an Entranc

zspencer:
https://github.com/zinc-collective/convene/issues/42

Apparently adding rooms to a Space with an entrance hasn't worked for a while. This adds a test to that.


anaulin (approved):
🔨🐛

PR #1104, opened 2023-02-07
`Journal`: `Entry`s are paginated

zspencer:
I am poking with using [Pagy](https://github.com/ddnexus/pagy) since it has better performance characteristics and more regular commits than either Kamanari or will_paginate.

That said, it doesn't seem to have great affordances for overriding classes on the nav and has had 6 major releses since it's inception a few years ago.

Anyway, it seems like not the worst gem for the job? Especially since it supports many different styles of pagination (date, cursor, etc)

Thoughts?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,8 @@ gem "redcarpet", "~> 3.6"
 gem "gretel", "~> 4.4"
 # Better UI components
 gem "view_component", "~> 2.82"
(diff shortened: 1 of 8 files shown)


anaulin (approved):
Looking at https://www.ruby-toolbox.com/categories/pagination, Pagy seems like a fine choice!

PR #1105, opened 2023-02-07
`Neighborhood`: Default to `postgres:latest` image for new databases

zspencer:
We had hardcoded in 12, but latest is probably more appropriate since postgresql's releases are generally top-notch and getting stuck in old postgres is a recipe for sad :'(

That said, it could also be a point of friction since it will mean we'll need to learn how to bump major versions of local / neighborhood databases... Anyway meh.

### DIFF
--- a/.github/workflows/test-convene-web.yml
+++ b/.github/workflows/test-convene-web.yml
@@ -45,7 +45,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:12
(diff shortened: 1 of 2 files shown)


anaulin (approved):
This is ok for now, but ideally we would be using in CI whatever we are using in production, and figure out a good way to keep those in sync.

zspencer:
Agreed.

PR #1123, opened 2023-02-16
Fix settings for marketplace delivery fee field

anaulin:

This fix makes it possible to set a delivery fee with two decimals.

### Before
image

### After
image


zspencer (approved):
Love it! Thank you!!!!

PR #1125, opened 2023-02-16
Show a link to add a Stripe API key if one isn't set

anaulin:

Instead of prompting to "connect stripe account", which fails if there is no Stripe API key, now we prompt to add a Space-level Stripe API key.

Initially I was linking directly to `utility_hookups/new`, but the problem with that is that it is possible to have already a utility hookup for stripe, but no API key, so going to `space/edit` seems better for now.

At some point we'll want to surround these buttons with policy checks, because maybe the vendor setting up the marketplace doesn't have permissions to set up the Stripe hookup for the space.

### Before
https://user-images.githubusercontent.com/6729309/219253804-57217199-06f2-49a0-9e37-034f3fd11158.mp4

### After
https://user-images.githubusercontent.com/6729309/219254250-5764d484-8d53-4e32-931f-fbaa1ad2ee59.mp4

### DIFF
--- a/app/furniture/marketplace/locales/en.yml
+++ b/app/furniture/marketplace/locales/en.yml
@@ -8,6 +8,7 @@ en:
       edit: "Configure Marketplace"
     marketplaces:
       edit:
+        add_stripe_hookup_key: "Add a Stripe API key to %{space_name}"
         connect_stripe: "Connect to Stripe"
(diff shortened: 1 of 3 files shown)


zspencer (approved):
This is much nicer! Thank you!!!

We may want to carry forward some design work on "how do we connect a Utility to a Furniture in a secure way?!?"

PR #1126, opened 2023-02-16
`Stripe`: Add a bit of help to the UI about finding the Stripe API keys

anaulin:
https://github.com/zinc-collective/convene/issues/1002

Because I need it every single time 😅

Also renamed "Utility Hookups" to "Utilities" in the UI.

### Before
image

### After
image

### DIFF
--- a/app/utilities/stripe_utilities/_form.html.erb
+++ b/app/utilities/stripe_utilities/_form.html.erb
@@ -1,1 +1,5 @@
 <%= render "password_field", form: form, attribute: :api_token %>
+<small>
+  <%= t('.api_key_help') %>
+  <%= link_to "https://dashboard.stripe.com/apikeys", "https://dashboard.stripe.com/apikeys" %>
+</small>
--- a/app/views/spaces/edit.html.erb
+++ b/app/views/spaces/edit.html.erb
@@ -45,7 +45,7 @@
 
 
 <fieldset>
-  <h3>Utility Hookups</h3>
+  <h3>Utilities</h3>
   <%- space.utility_hookups.each do |utility_hookup| %>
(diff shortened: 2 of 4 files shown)


app/utilities/stripe_utilities/_form.html.erb
```diff
@@ -1 +1,5 @@
 <%= render "password_field", form: form, attribute: :api_token %>
+<small>
+  <%= t('.api_key_help') %>
+  <%= link_to "https://dashboard.stripe.com/apikeys", "https://dashboard.stripe.com/apikeys" %>
+</small>
```
zspencer, line 5: It would be cool if our `password_field` was able to identify if there is help text for the attribute from our locales files and emit that; but we don't have to do that now.

app/views/spaces/edit.html.erb
```diff
@@ -45,7 +45,7 @@
 
 
 <fieldset>
-  <h3>Utility Hookups</h3>
+  <h3>Utilities</h3>
```
zspencer, line 48: 🎉🎉🎉

config/environments/development.rb
```diff
@@ -68,7 +68,7 @@
   config.assets.quiet = true
 
   # Raises error for missing translations.
-  # config.i18n.raise_on_missing_translations = true
+  config.i18n.raise_on_missing_translations = true
```
zspencer, line 71: Oooo, danger zone! (But good, I think?)

zspencer (approved):
I dig it! Thank you!!!

config/environments/development.rb
anaulin: I didn't mean to commit this. (Although in development it is kinda helpful?)

I reverted this.

PR #1138, opened 2023-02-25
`Marketplace`: Prevent race-condition in `Checkout`

zspencer:

So, it turns out *sometimes* Stripe will fire off that checkout session completed event faster than we can handle it!

Now that we listen for webhook events, it makes sense that we don't need to send people back to show their checkout; and can send them right on to their order.

Note: We may be able to get rid of the stripe checkout session id bits completely, since we use the transfer group as our key for looking up which order to apply the charge to.

### DIFF
--- a/app/furniture/marketplace/checkout.rb
+++ b/app/furniture/marketplace/checkout.rb
@@ -3,7 +3,7 @@ class Marketplace
     self.location_parent = :cart
     include ActiveModel::Validations
     attr_accessor :cart
-    delegate :shopper, :marketplace, to: :cart
+    delegate :shopper, :marketplace, :persisted?, to: :cart
 
     # It would be nice to validate instead the presence of :ordered_products, but my attempts at this raise:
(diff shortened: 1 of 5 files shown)


daltonrpruitt (approved):
I looked at this, but I don't think I know enough about the context to really say for sure that it's good or to even test it myself yet.
My first (and only) thought with the convene's side being too slow to handle the stripe event is to provide some block for an async-await type structure (my only experience with web-dev before now was some basic javascript years ago). I assumed something similar was already being done due to Ruby's native support for passing lamdas around (essentially), but I could be wrong. 
I'll be the first to note my review here is shallow, so please do wait for further approval if desired.

zspencer:
This was an inter-process race-condition caused by the Stripe Checkout Success event firing before the Stripe Checkout redirected back to Convene; rather than an in-process race-condition.

And a shallow review is all I need! Thanks @daltonrpruitt !

PR #1140, opened 2023-02-27
`Dev`: Install Solargraph  on DevContainer

zspencer:
This also formats the JSON file with Prettier, so it follows the rest of our codebases conventions

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,46 +1,47 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the
 // README at: https://github.com/devcontainers/templates/tree/main/src/ruby-rails-postgres
 {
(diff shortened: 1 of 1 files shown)


daltonrpruitt (approved):
I didn't think this could possibly cause errors (since the only logic change was adding a VS Code extension), but I wanted to rebuild the container to see if anything funky would happen. 
Lo and behold, I am currently getting an error when building in codespaces. 
However, I can startup the Dev Container fine locally. 
I'm chalking this up to a fluke, since I was able to build correctly a second time. 
Thank you for the improvements!

zspencer:
Thanks!!! I was able to start a Github CodeSpace with Solargraph and it worked; so it may have been a fluke

I've noticed a lot of builds failing trying to instal Overmind lately; looks like the Github CDN is struggling a bit?

Issue #890, opened 2022-10-21
🥔✨ `Neighborhood`: `Operator` First Run Experience


zspencer:
There's a number of pain points around setting up Convene for new contributors. This is a catch-all for tracking the pain-points and polish for setting up a new neighborhood as a Contributor.

- [x] https://github.com/zinc-collective/convene/issues/755

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1149, opened 2023-02-28
`Neighborhood`: Delete dead `Client`-based `Blueprint`.

zspencer:
- https://github.com/zinc-collective/convene/issues/476
- https://github.com/zinc-collective/convene/issues/890

This was put in place to reduce cost-of-setup, but is no longer necessary.


KellyAH (approved):
LGTM

Issue #1154, opened 2023-03-02
🥔⛰️ `Space`


zspencer:
`Space` organize `Client`'s data within a `Neighborhood`

## Use Cases
- [ ] 🌱 Registering a Space
  - [ ] https://github.com/zinc-collective/convene/issues/1585
- [x] https://github.com/zinc-collective/convene/issues/117
- [x] #1364
- [x] #1514
- [x] https://github.com/zinc-collective/convene/issues/74
- [x] https://github.com/zinc-collective/convene/issues/1473
- [x] https://github.com/zinc-collective/convene/issues/833
- [x] 🥔 Configuring a Space
- [x] https://github.com/zinc-collective/convene/issues/621

## Bugs
- [x] https://github.com/zinc-collective/convene/issues/1658
- [x] https://github.com/zinc-collective/convene/issues/1659

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1158, opened 2023-03-02
`Space`: `Entrance` may be set via UI

zspencer:
- https://github.com/zinc-collective/convene/issues/1154

Turns out, we had never exposed an affordance for setting the `Space` entrance! Whoopsie-doodle!

### DIFF
--- a/app/views/spaces/edit.html.erb
+++ b/app/views/spaces/edit.html.erb
@@ -39,6 +39,7 @@
                include_blank: false,
                options: Space::THEME_OPTIONS.map { |t| [t.titleize, t] },
                form: space_form %>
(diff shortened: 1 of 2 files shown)


app/views/spaces/edit.html.erb
```diff
@@ -39,6 +39,7 @@
                include_blank: false,
                options: Space::THEME_OPTIONS.map { |t| [t.titleize, t] },
                form: space_form %>
+    <%= render "select", attribute: :entrance_id, include_blank: true, options: space.rooms.map { |r| [r.name, r.id] }, form: space_form %>
```
KellyAH, line 42: I assume `r` is `room`...
```suggestion
    <%= render "select", attribute: :entrance_id, include_blank: true, options: space.rooms.map { |room| [room.name, room.id] }, form: space_form %>
```

KellyAH (approved):

PR #1160, opened 2023-03-02
`Dev`: Fix bash completion in codespaces

daltonrpruitt:
Trying to get git (and possibly other bash) completion to work in the devcontainer.
Refer to the [container repo page](https://mcr.microsoft.com/en-us/product/devcontainers/universal/about) and the [actual Dockerfile](https://github.com/devcontainers/images/blob/main/src/universal/.devcontainer/Dockerfile) for details on the `universal` container image.


daltonrpruitt:
Tested on the branch. Able to fetch and switch to upstream branches using completion. 
There may be other added bonuses to using the `universal` image (such as not having to update the devcontainer setup again if necessary).

PR #1161, opened 2023-03-02
`Client`: Seems unnecessary and is getting in the way of things

zspencer:
At some point, we may need to group `Space`s within `Client`, but today is not that day.


KellyAH (approved):
💎

PR #1165, opened 2023-03-02
🧹 `Invitation`: Follow `rspec-rails` conventions for Request spec

zspencer:
- https://github.com/zinc-collective/convene/issues/117

`rspec-rails` wants request specs to:
1. Be nested in folders by namespace, rather than route
2. Be named for their `Controller`
3. Use the `describe "#action"` syntax, rather than the path
4. Start `context` block descriptions with `when`, `with`, and `without`

### DIFF
--- a/spec/requests/rsvps_controller_request_spec.rb
+++ b/spec/requests/rsvps_controller_request_spec.rb
@@ -2,13 +2,13 @@
 
 require "rails_helper"
 
-RSpec.describe "/spaces/:space_id/invitations/:invitation_id/rsvp", type: :request do
+RSpec.describe RsvpsController do
   let(:space) { invitation.space }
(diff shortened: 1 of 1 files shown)


KellyAH (approved):
❤️  this! makes it easier to scan tests and map which tests are for which controller action.

KellyAH:
Does this change need to be made to other controller specs too?

Screenshot 2023-03-02 at 11 21 30 PM

zspencer:
The `furniture` folder is a bit special; in that `spec/furniture/marketplace` has all the specs for the marketplace. I could see a world where `furniture` becomes a top-level folder (sibling to `app`) and then each `furniture` has it's own `spec` folder; but I don't quite know how to make that go!

PR #1166, opened 2023-03-02
🏎️ `Membership`: Find `current_membership` through `Person`, not `Space` 

zspencer:
- https://github.com/zinc-collective/convene/issues/117

This is likely a premature optimization, but my inclination is that a `Person` will likely have fewer `Membership`s than a `Space` does; which means the index on `Membership#space_id` will likely require more compute to locate the particular `Membership` than the `Membership#person_id` index.


KellyAH (approved):
Makes sense to me.

Seems more efficient and faster to find the `current_membership` via the `current_person`, rather than the `current_space` which could have a ton of memberships that were or never were accepted.

Issue #1155, opened 2023-03-02
⛰️✨ `Sections`


zspencer:
- Rooms are Convene's sub-organizational unit below  https://github.com/zinc-collective/convene/issues/1154.
   https://github.com/zinc-collective/convene/issues/709 are placed in https://github.com/zinc-collective/convene/issues/1155

- From an Interaction Design perspective, https://github.com/zinc-collective/convene/issues/1155 are like a screen on a Mobile App or a landing page on a Website. Every https://github.com/zinc-collective/convene/issues/1154 has a default room, which we call the `Entrance`

## Use Cases
- [x] #1965
- [x] ✨ `Section`: Adding `Sections` to `Spaces`
- [x] ✨ `Section`: Removing `Sections` from `Spaces`
- [x] ✨ `Section`: Limit `Sections` to `Spaces`

## Data Model
- [x] #1961
- [x] #2228

## Design

- [ ] #1472

## Bugs

- [x] https://github.com/zinc-collective/convene/issues/1841

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

Issue #926, opened 2022-11-10
🔐🤔`ApplicationController` makes `authorize` calls -- should they be policy-scoped calls instead?


anaulin:
`ApplicationController` makes `authorize` calls when it fetches the current space and room. Should these instead be policy-scoped lookups?

- [x] https://github.com/zinc-collective/convene/pull/929
- [ ] ~~Don't use policy_scope either, since that bypasses the scope checks in downstream Controllers!~~

anaulin:
@zspencer yeah, this seems fine. Let's close.

PR #1167, opened 2023-03-02
🧹🔐 `Room`: Don't always provide a `current_room`

zspencer:
- https://github.com/zinc-collective/convene/issues/926

There are many pages that are not in a `room`, and the `policy_scope` of `current_room` could result in a false safety.

That said, we probably should *also* be writing `request` specs to confirm that folks get 404 on endpoints outside of their scope...

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -128,19 +128,6 @@ class ApplicationController < ActionController::Base
     policy_scope(Space.includes(:rooms, entrance: [:furniture_placements]))
   end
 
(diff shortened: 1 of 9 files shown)


zspencer:
I think this is a mite dangerous, in particular because we do have the idea of an `entrance`, so I think it's probably best if we don't apply this change.

anaulin:
Yeah, I don't think it is worth removing this. The `policy_scope` here feels fine, and it feels right to have a `current_room` at the `ApplicationController` level. I say let's drop this for now. (We can always revisit if we run into issues again.)

Issue #1136, opened 2023-02-24
🗺️ `Marketplace`: `DeliveryArea`


zspencer:
- https://github.com/zinc-collective/convene/issues/831
- https://github.com/zinc-collective/convene/issues/1325

In the beginning, the [Marketplace](https://github.com/zinc-collective/convene/issues/831) only supported per-Marketplace delivery pricing. 

When chatting with [Piikup](https://github.com/zinc-collective/convene/issues/832), we learned that they prefer to price  on a per-`Marketplace` basis based on which `DeliveryArea` the `Shopper` collects

### Fauxsonas
See [Marketplace Fauxsonas](https://github.com/zinc-collective/convene/issues/831)

### Use Cases

- [x] 🥔 `Distributor` adds `DeliveryArea`s
- [ ] #2014
- [x] 🥔 `Shopper` sets `DeliveryArea` for `Order`
- [x] 🥔`Shopper` pays `DeliveryFee` based upon `DeliveryArea`
- [x] 🥔 `Distributor` sets `OrderBy` and `DeliveryWindow` for `DeliveryArea`
- [x] 🥔`Shopper` sees `DeliveryConstraints` based upon `DeliveryArea`
- [x] 🥔 `Shopper` may not see `Marketplaces` without a `DeliveryArea`

zspencer:
From April:
> We're going for Catering, so Restaurants and Food Makers and the pricing is based on City.
> So if we are delivering from Cafe Gabriele or Understory those are Catering orders, and the prices are:
> - Oakland, Alameda, Berkeley and Emeryville is $50 for those cities
> - San Francisco, $75

zspencer:
I think this is good enough for us to close! We can keep refactoring and tidying if we want, but I'm not sure what other use cases it's missing.

zspencer:
Re-opening because it does seem like we have more work to do to make this tight.

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1169, opened 2023-03-02
🔐 `Marketplace`: Encrypt `Order#delivery_address`!

zspencer:

So, I totally didn't think about how delivery addresses are PII and probably should not be stored in plaintext! Womp. Womp. Womp.

Now they ain't!

We'll want to delete the `release:after_build` bits after a prod deploy.

### DIFF
--- a/app/furniture/marketplace/order.rb
+++ b/app/furniture/marketplace/order.rb
@@ -11,7 +11,7 @@ class Marketplace
     has_many :ordered_products, inverse_of: :order, foreign_key: :cart_id
     has_many :products, through: :ordered_products, inverse_of: :orders
 
-    has_encrypted :delivery_address
+    has_encrypted :delivery_address, migrating: true
 
     enum status: {
       pre_checkout: "pre_checkout",
--- a/app/furniture/marketplace/order.rb
+++ b/app/furniture/marketplace/order.rb
@@ -11,7 +11,7 @@ class Marketplace
     has_many :ordered_products, inverse_of: :order, foreign_key: :cart_id
     has_many :products, through: :ordered_products, inverse_of: :orders
 
-    attribute :delivery_address, :string
+    has_encrypted :delivery_address
 
     enum status: {
       pre_checkout: "pre_checkout",
--- a/db/migrate/20230302202459_marketplace_encrypt_order_delivery_address.rb
+++ b/db/migrate/20230302202459_marketplace_encrypt_order_delivery_address.rb
@@ -1,6 +1,5 @@
 class MarketplaceEncryptOrderDeliveryAddress < ActiveRecord::Migration[7.0]
   def change
-    rename_column :marketplace_orders, :delivery_address, :deprecated_delivery_address
     add_column :marketplace_orders, :delivery_address_ciphertext, :text
   end
 end
--- a/db/migrate/20230302202459_marketplace_encrypt_order_delivery_address.rb
+++ b/db/migrate/20230302202459_marketplace_encrypt_order_delivery_address.rb
@@ -1,0 +1,6 @@
+class MarketplaceEncryptOrderDeliveryAddress < ActiveRecord::Migration[7.0]
+  def change
+    rename_column :marketplace_orders, :delivery_address, :deprecated_delivery_address
+    add_column :marketplace_orders, :delivery_address_ciphertext, :text
+  end
+end
--- a/lib/tasks/release.rake
+++ b/lib/tasks/release.rake
@@ -3,6 +3,12 @@
 namespace :release do
   desc "Ensures any post-release / pre-deploy behavior has occurred"
   task after_build: [:environment, "db:prepare"] do
+    # @todo Delete after running in prod
+    Marketplace::Order.all.find_each do |order|
+      next unless order.deprecated_delivery_address.present?
+
+      order.update(delivery_address: order.deprecated_delivery_address, deprecated_delivery_address: nil)
+    end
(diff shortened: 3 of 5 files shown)


KellyAH:
> We'll want to delete the release:after_build bits after a prod deploy.

why? because they contain the plain text delivery addresses PII?

KellyAH:
Aren't we also storing the contact person's email address too after every stripe order... is that PII (personal identifiable information) too that needs to be encrypted?

db/migrate/20230302202459_marketplace_encrypt_order_delivery_address.rb
```diff
@@ -0,0 +1,6 @@
+class MarketplaceEncryptOrderDeliveryAddress < ActiveRecord::Migration[7.0]
+  def change
+    rename_column :marketplace_orders, :delivery_address, :deprecated_delivery_address
+    add_column :marketplace_orders, :delivery_address_ciphertext, :text
```
KellyAH, line 4: why not encrypt the existing column like lockbox gem says is possible [here](https://github.com/ankane/lockbox#migrating-existing-data)?

lib/tasks/release.rake
```diff
@@ -3,6 +3,12 @@
 namespace :release do
   desc "Ensures any post-release / pre-deploy behavior has occurred"
   task after_build: [:environment, "db:prepare"] do
+    # @todo Delete after running in prod
+    Marketplace::Order.all.find_each do |order|
+      next unless order.deprecated_delivery_address.present?
+
+      order.update(delivery_address: order.deprecated_delivery_address, deprecated_delivery_address: nil)
+    end
```
KellyAH, line 11: I see.. this is the prod data migration.....

So.. I assume this rake task runs BEFORE the `delivery_address` column is renamed in `20230302202459_marketplace_encrypt_order_delivery_address.rb`

app/furniture/marketplace/order.rb
```diff
@@ -11,7 +11,7 @@ class Order < Record
     has_many :ordered_products, inverse_of: :order, foreign_key: :cart_id
     has_many :products, through: :ordered_products, inverse_of: :orders
 
-    attribute :delivery_address, :string
+    has_encrypted :delivery_address
```
KellyAH, line 14: What's the order of operations for everything this code change does?

Seems like in this change...
1. `:delivery_address` is being renamed to `deprecated_delivery_address`
2. data in `deprecated_delivery_address` is migrated to `delivery_address_ciphertext`
3.  data in `deprecated_delivery_address` is replaced with `nil`

And in later PR... I assume this happens?:
4. `deprecated_delivery_address` is dropped.
5. `delivery_address_ciphertext` is renamed `delivery_address` so `Order` model still works.

db/migrate/20230302202459_marketplace_encrypt_order_delivery_address.rb
zspencer: Ah, the migrating syntax is neat! I didn't realize it existed; but yea, that would save us the `rename_column` step and having to write the `find_each`!

I'll definitely do it next time!

lib/tasks/release.rake
zspencer: Ah, let me step ya through the Heroku deploy process:

1. Heroku checks out the git repository, and starts a build which creates a new `image` and installs gems, node packages, etc.
2.  Heroku launches a `release` container using the new `image`, and runs the `release` command defined in the `Procfile`, which runs the `rails db:migrate` and release:after_build` tasks in sequence
4. Launches new `web` and `worker` containers using the new `image`.
5. Begins routing traffic to the new `web` containers
6. Stops routing traffic to the old `web` containers
7. Shuts down the old `web` and `worker` containers.

So the database has the new columns prior to running the rake task, *and* before deploying new `web` and `worker` containers. this means that the existing `web` and `worker` containers will still be running the code which references `delivery_address`, which could result in sql query errors. We could [install `strong_migrations`](https://github.com/ankane/strong_migrations) to catch risky operations (like `rename`s); but in this particular case I felt like it was safe enough given our traffic.

That said, there's no reason *not* to use the `Lockbox.migrate` strategy; which eliminates the risk and seems pretty dang cool!

app/furniture/marketplace/order.rb
zspencer: Yes *except* step 5 never happens; because the `Lockbox` infers that `delivery_address_ciphertext` is the database name for the encrypted `delivery_address` attribute.

zspencer:
> Aren't we also storing the contact person's email address too after every stripe order... is that PII (personal identifiable information) too that needs to be encrypted?

Good catch. This is part of why we have the `AuthenticationMethod` table, which *should* have encrypt the `contact_location` field and give it a `blind_index`. I don't think I want to bundle that into this PR; but I may wind up making a second one!

zspencer:
@KellyAH - I swapped to `Lockbox.migrate` since it was pretty straightforward, and gives us an example of doing it "the safer way" for future.

lib/tasks/release.rake
KellyAH: I made issue https://github.com/zinc-collective/convene/issues/1177 to install [strong_migrations](https://github.com/ankane/strong_migrations). 

My work team uses it and it is really handy.

KellyAH (approved):
🚢

lib/tasks/release.rake
zspencer: Yah!!!!

PR #1141, opened 2023-02-27
`Neighborhood`: Assign an `Operator` and create a `Space` on first run

zspencer:
* https://github.com/zinc-collective/convene/issues/890

- [x] Pull out:
  - [x] Extracting the `CardComponent` 
  - [x] Removing the `DemoSpace`
- [x] Get tests passing (or delete them)
- [x] Assign the `Operator` somehow
- [x] Only let Operators create a `Space` or Update a `Neighborhood`

https://user-images.githubusercontent.com/50284/222006383-2dc0f1c8-aaa6-465a-8f8c-49a30dc13f4f.mp4

PR #1171, opened 2023-03-02
🐞 `Space`: Fix broken `SpacesController#new` action

zspencer:
- https://github.com/zinc-collective/convene/issues/1154
- https://github.com/zinc-collective/convene/pull/1161
- https://github.com/zinc-collective/convene/pull/1141

Deleting the `Client` was great! But when we merged the `Neighborhood` first-run PR, we still had the `Client#name` field; which no longer exists.

So now there's two tests for the `SpacesController#new` action:
- 1 to catch any errors in the happy-path request/response cycle
- 1 to make sure `Guest`s cant create a Space

### DIFF
--- a/spec/requests/spaces_controller_request_spec.rb
+++ b/spec/requests/spaces_controller_request_spec.rb
@@ -5,39 +5,6 @@ require "swagger_helper"
 RSpec.describe SpacesController do
   include ActiveJob::TestHelper
 
-  describe "#show" do
-    subject(:perform_request) do
-      get url
-      test_response
-    end
-
-    let(:space) { create(:space) }
-    let(:url) { polymorphic_url(space) }
-
-    it { is_expected.to be_ok }
-    specify { perform_request && assert_select("##{dom_id(space)}") }
-
-    context "with a branded domain" do
-      let(:space) { create(:space, branded_domain: "beta.example.com") }
-
-      context "when accessing via the neighborhood url" do
-        it { is_expected.to redirect_to "http://beta.example.com" }
-      end
-
-      context "when accessing via domain" do
-        before do
-          space
-          host! "beta.example.com"
-        end
-
-        let(:url) { "http://beta.example.com" }
-
-        it { is_expected.to be_ok }
-        specify { perform_request && assert_select("##{dom_id(space)}") }
-      end
-    end
-  end
-
   path "/spaces" do
     include ApiHelpers::Path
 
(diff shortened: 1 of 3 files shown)


spec/support/auth_helpers.rb
```diff
@@ -8,7 +8,7 @@ def sign_in(space, person)
 
     authentication_method.bump_one_time_password!
 
-    post(space_authenticated_session_path(space),
```
KellyAH, line 11: was `space_authenticated_session_path` deleted with the deletion of client?

spec/requests/spaces_controller_request_spec.rb
```diff
@@ -67,8 +55,21 @@
...
+        get polymorphic_path(space)
+
+        expect(response).to redirect_to "http://beta.example.com"
+      end
+    end
+  end
+
   describe "#destroy" do
-    context "as an Operator using the API" do
+    context "when an API-Operator" do
```
KellyAH, line 72: Do we now have 2 different types of operators?
`API-Operator` and `Operator`?

`as an Operator using the API` was fine as is!

spec/requests/spaces_controller_request_spec.rb
```diff
@@ -110,4 +111,24 @@
...
+    context "when not logged in" do
+      let(:user) { nil }
+
+      it { is_expected.to be_not_found }
+    end
+
+    context "when an Operator" do
+      let(:user) { create(:person, operator: true) }
+
+      it { is_expected.to be_ok }
```
KellyAH, line 131: so only `operators` can make spaces but not space members?

spec/requests/spaces_controller_request_spec.rb
zspencer: Right now, we have an API key that has `Operator` level privileges. but I don't think this rephrase is useful either.

spec/requests/spaces_controller_request_spec.rb
zspencer: Yes, for now. We may want to set up `Neighorhood#accepting_new_spaces?` or something so that `Operator`s can turn self-registration on and off; similar to how Mastodon allows their moderators to enable and disable public registrations.

spec/support/auth_helpers.rb
zspencer: No, but it needs to use the `polymorphic_path` so that it works when on the `NeighborhoodsController#show` action.

KellyAH (approved):

PR #1173, opened 2023-03-02
✨ `Marketplace`: Collect `Shopper`s Delivery Address before `Checkout`

zspencer:

While collecting the address via Stripe was reasonable at the time, if we want to support dynamic pricing, we'll need the address ahead of time.

This will also (theoretically) allow us to let people know when they are outside of the `Distributor` delivery area.

## After!

https://user-images.githubusercontent.com/50284/222832192-12b8f715-b0d5-4d1e-bd09-9dfdcf8f9d9a.mp4

## Before

https://user-images.githubusercontent.com/50284/222587195-9b7b01e9-5fd1-4732-824f-663516c5c83e.mp4

### DIFF
--- a/app/furniture/marketplace/cart.rb
+++ b/app/furniture/marketplace/cart.rb
@@ -28,7 +28,11 @@ class Marketplace
       end
     end
 
-    delegate :delivery_fee, to: :marketplace
+    def delivery_fee
+      return marketplace.delivery_fee if delivery_address.present?
+
+      0
+    end
 
     def price_total
       product_total + delivery_fee
--- a/app/furniture/marketplace/carts/_cart.html.erb
+++ b/app/furniture/marketplace/carts/_cart.html.erb
@@ -12,7 +12,7 @@
         Delivering to <%= cart.delivery_address %>
 
       </p>
-      <%= button_to "Change Address", cart.location, params: { cart: { delivery_address: nil } }%>
+      <%= button_to "Update Address", cart.location, params: { cart: { delivery_address: nil } }%>
     </div>
 
   <%- end %>
--- a/app/furniture/marketplace/carts/_cart.html.erb
+++ b/app/furniture/marketplace/carts/_cart.html.erb
@@ -15,30 +15,29 @@
       <%= button_to "Update Address", cart.location, params: { cart: { delivery_address: nil } }%>
     </div>
 
+    <div class="-mx-4 mt-8 overflow-hidden shadow ring-1 ring-black ring-opacity-5 sm:-mx-6 md:mx-0 md:rounded-lg">
+      <table class="min-w-full divide-y divide-gray-300 table-fixed">
+        <thead class="bg-gray-50">
+          <tr>
+            <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6">
+              <%= Marketplace::Product.human_attribute_name(:name) %>
+            </th>
+            <th scope="col" class="hidden px-3 py-3.5 text-left text-sm font-semibold text-gray-900 lg:table-cell">
+              <%= Marketplace::Product.human_attribute_name(:description) %>
+            </th>
+            <th scope="col" class="hidden px-3 py-3.5 text-left text-sm font-semibold text-gray-900 sm:table-cell">
+              <%= Marketplace::Product.human_attribute_name(:price) %>
+            </th>
+            <th  scope="col" class="w-48">&nbsp;</th>
+          </tr>
+        </thead>
+        <tbody class="divide-y divide-gray-200 bg-white">
+          <%- cart.marketplace.products.each do |product| %>
+            <%= render cart.cart_products.find_or_initialize_by(product: product) %>
+          <%- end %>
+        </tbody>
+        <%= render "marketplace/carts/footer", cart: cart %>
+      </table>
+    </div>
   <%- end %>
-
-  <div class="-mx-4 mt-8 overflow-hidden shadow ring-1 ring-black ring-opacity-5 sm:-mx-6 md:mx-0 md:rounded-lg">
-    <table class="min-w-full divide-y divide-gray-300 table-fixed">
-      <thead class="bg-gray-50">
-        <tr>
-          <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6">
-            <%= Marketplace::Product.human_attribute_name(:name) %>
-          </th>
-          <th scope="col" class="hidden px-3 py-3.5 text-left text-sm font-semibold text-gray-900 lg:table-cell">
-            <%= Marketplace::Product.human_attribute_name(:description) %>
-          </th>
-          <th scope="col" class="hidden px-3 py-3.5 text-left text-sm font-semibold text-gray-900 sm:table-cell">
-            <%= Marketplace::Product.human_attribute_name(:price) %>
-          </th>
-          <th  scope="col" class="w-48">&nbsp;</th>
-        </tr>
-      </thead>
-      <tbody class="divide-y divide-gray-200 bg-white">
-        <%- cart.marketplace.products.each do |product| %>
-          <%= render cart.cart_products.find_or_initialize_by(product: product) %>
-        <%- end %>
-      </tbody>
-      <%= render "marketplace/carts/footer", cart: cart %>
-    </table>
-  </div>
 </div>
--- a/app/furniture/marketplace/carts/_cart.html.erb
+++ b/app/furniture/marketplace/carts/_cart.html.erb
@@ -1,24 +1,44 @@
-<div class="-mx-4 mt-8 overflow-hidden shadow ring-1 ring-black ring-opacity-5 sm:-mx-6 md:mx-0 md:rounded-lg">
-  <table class="min-w-full divide-y divide-gray-300 table-fixed">
-    <thead class="bg-gray-50">
-      <tr>
-        <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6">
-          <%= Marketplace::Product.human_attribute_name(:name) %>
-        </th>
-        <th scope="col" class="hidden px-3 py-3.5 text-left text-sm font-semibold text-gray-900 lg:table-cell">
-          <%= Marketplace::Product.human_attribute_name(:description) %>
-        </th>
-        <th scope="col" class="hidden px-3 py-3.5 text-left text-sm font-semibold text-gray-900 sm:table-cell">
-          <%= Marketplace::Product.human_attribute_name(:price) %>
-        </th>
-        <th  scope="col" class="w-48">&nbsp;</th>
-      </tr>
-    </thead>
-    <tbody class="divide-y divide-gray-200 bg-white">
-      <%- cart.marketplace.products.each do |product| %>
-        <%= render cart.cart_products.find_or_initialize_by(product: product) %>
-      <%- end %>
-    </tbody>
-    <%= render "marketplace/carts/footer", cart: cart %>
-  </table>
+<div id="<%= dom_id(cart) %>">
+
+  <%- if cart.delivery_address.blank? %>
+    <%= form_with model: cart.location do |cart_form| %>
+      <%= render "text_field", attribute: :delivery_address, form: cart_form %>
+
+      <%= cart_form.submit %>
+    <%- end %>
+  <%- else %>
+    <div class="flex flex-wrap">
+      <p class="p-3">
+        Delivering to <%= cart.delivery_address %>
+
+      </p>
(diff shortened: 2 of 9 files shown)


app/furniture/marketplace/carts/_cart.html.erb
```diff
@@ -1,24 +1,44 @@
...
-      </tr>
-    </thead>
-    <tbody class="divide-y divide-gray-200 bg-white">
-      <%- cart.marketplace.products.each do |product| %>
-        <%= render cart.cart_products.find_or_initialize_by(product: product) %>
-      <%- end %>
-    </tbody>
-    <%= render "marketplace/carts/footer", cart: cart %>
-  </table>
+<div id="<%= dom_id(cart) %>">
```
zspencer, line 1: Sorry for the indentation change, but I needed to put in an id for the whole shebang so we could replace it all at once :X.

app/furniture/marketplace/checkout.rb
```diff
@@ -16,9 +16,6 @@ def create_stripe_session(success_url:, cancel_url:)
         mode: "payment",
         success_url: success_url,
         cancel_url: cancel_url,
-        shipping_address_collection: {
-          allowed_countries: ["US"],
-        },
```
zspencer, line 21: We don't want to get it from Stripe anymore, since we get it before hand.

app/furniture/marketplace/carts/_total.html.erb
```diff
@@ -1,10 +1,12 @@
-<span id="cart-total-<%= cart.id%>" class="w-full flex flex-col">
-  <span class="text-right">
+<span id="cart-total-<%= cart.id%>" class="w-full flex flex-col sm:text-right">
```
zspencer, line 1: We were setting the text-right in every span, which seemed excessive.

KellyAH:
The `After!` video shows the saved address not auto-populating the field when the `change address` button is clicked. Was that fixed to not show a blank field when changing a saved address?

app/furniture/marketplace/carts/_footer.html.erb
```diff
@@ -8,8 +8,9 @@
   <tr>
     <td class="text-left px-3 py-3.5 text-right" colspan="8">
       <%= render "marketplace/carts/total", cart: cart %>
-      <%= button_to("Checkout", cart.location(child: :checkout), data: { turbo: false }) %>
+      <%- if cart.delivery_address.present? %>
+        <%= button_to("Checkout", cart.location(child: :checkout), data: { turbo: false }) %>
+      <%- end %>
```
KellyAH, line 13: this seems to make the checkout button hidden if the user never inputted a `cart.delivery_address`. Customers might worry when they don't see the checkout button.

Why not just have the `Checkout` disabled if `cart.delivery_address` does not exist or is empty string? Or always keep `Checkout` enabled,  but if they submit the form without the `cart.delivery_address`, do not let the form submit and inform them that `cart.delivery_address` is required.

app/furniture/marketplace/cart.rb
```diff
@@ -26,7 +26,11 @@ def product_total
       end
     end
 
-    delegate :delivery_fee, to: :marketplace
+    def delivery_fee
+      return marketplace.delivery_fee if delivery_address.present?
```
KellyAH, line 30: I assume this logic is covering the fact that the `cart.delivery_fee` cannot be calculated without a `delivery_address` so it sets `delivery_fee` to zero.

app/furniture/marketplace/carts/_cart.html.erb
```diff
@@ -1,24 +1,44 @@
...
+
+      <%= cart_form.submit %>
+    <%- end %>
+  <%- else %>
+    <div class="flex flex-wrap">
+      <p class="p-3">
+        Delivering to <%= cart.delivery_address %>
+
+      </p>
+      <%= button_to "Change Address", cart.location, params: { cart: { delivery_address: nil } }%>
```
KellyAH, line 15:
```suggestion
      <%= button_to "Update Address", cart.location, params: { cart: { delivery_address: nil } }%>
```

app/furniture/marketplace/carts/_total.html.erb
```diff
@@ -1,10 +1,12 @@
...
+  <span>
     Products:
     <%= humanized_money_with_symbol(cart.product_total) %></span>
-  <span class="text-right">Delivery:
-    <%= humanized_money_with_symbol(cart.delivery_fee) %></span>
-  <span class="text-right font-bold">
+  <%- if cart.delivery_address.present? %>
+    <span>Delivery:
```
KellyAH, line 6: Should this be?
```suggestion
    <span>Delivery Fee:
```

app/furniture/marketplace/carts/_cart.html.erb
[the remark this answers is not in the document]
KellyAH: nevermind. found it in https://apidock.com/rails/ActionView/RecordIdentifier/dom_id

app/furniture/marketplace/cart.rb
zspencer: Exactly.

app/furniture/marketplace/carts/_footer.html.erb
zspencer: `disabled` is better than hidden for sure! I considered making it so if it wasn't set, the `CheckoutsController#show` action gave them an affordance for adding the `Cart#delivery_address` but I decided against it because it was more code.

I also considered putting the entire cart *except* for the `delivery_address` behind a `if/else` block; but that seemed excessive.

app/furniture/marketplace/carts/_cart.html.erb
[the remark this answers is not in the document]
zspencer: Yea, it's part of the Rails Model/View/Controller coupling; where a unique `id` attribute is generated from the model element. It's pretty useful in `turbo` land!

zspencer:
> The `After!` video shows the saved address not auto-populating the field when the `change address` button is clicked. Was that fixed to not show a blank field when changing a saved address? It would sux as a user being forced to retype the whole address if they only needed to fix a typo.

So if I were to make it that smart, I would likely need to promote the `delivery_address` from a `String` to a full-on `Model`, which I'm not opposed to but felt like it was too big of a step for this PR. I had started down that path, going so far as sprouting a new `Record` for `Marketplace::DeliveryAddress`, which `belongs_to :shopper` and `has_many :carts` and `has_many :orders`.

But that would have wound up adding a few hundred lines of code; and I didn't want to do that just yet.

app/furniture/marketplace/carts/_footer.html.erb
zspencer: @KellyAH - Do you have a preference between:

1. Disabling the `checkout` button unless the `delivery_address` is set, _or_
2. Hiding the `cart` until it has a `delivery_address` 

Part of me thinks option 2 is actually worth doing; since it serves as a [forcing function](https://www.interaction-design.org/literature/book/the-glossary-of-human-computer-interaction/forcing-functions) and reduces the amount of elements on the page significantly...

spec/furniture/marketplace/carts_controller_request_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  let(:marketplace) { create(:marketplace) }
+  let(:space) { marketplace.space }
+  let(:room) { marketplace.room }
+  let(:person) { create(:person) }
+
+  let(:shopper) { create(:marketplace_shopper, person: person) }
+
+  describe "#update" do
+    it "changes the delivery address" do
+      cart = create(:marketplace_cart, marketplace: marketplace, shopper: shopper)
```
KellyAH, line 13: why can't the `cart` be a `let` outside the `it` block? Is it because the `cart` instance must be created fresh for every `it` block? And making `cart` a `let` would make it a shared object accessed between `it` blocks and that would make tests too coupled or possibly cause tests to be flaky if they use/modify the same `cart` object?

zspencer: Nah, it's because I didn't have multiple `it` statements; but I could pull this into a `let`/`before`/`subject` structure if ya want.

spec/furniture/marketplace/cart_spec.rb
```diff
@@ -22,7 +22,13 @@
       cart.cart_products.create!(product: product_b, quantity: 2)
     end
 
-    it { is_expected.to eql(product_a.price + product_b.price + product_b.price + marketplace.delivery_fee) }
+    it { is_expected.to eql(product_a.price + product_b.price * 2) }
+
+    context "when the #delivery_address is present" do
+      let(:cart) { create(:marketplace_cart, delivery_address: "123", marketplace: marketplace) }
+
+      it { is_expected.to eql(product_a.price + product_b.price + product_b.price + marketplace.delivery_fee) }
```
KellyAH, line 30: why is `product_b.price` listed twice? Is that a typo or is the test increasing `product_b` quantity in the cart just by adding a 2nd `product_b.price`?

zspencer: Yep, the `cart_product` that references `product_b` has a quantity of `2`; I could do `product_b.price * 2` if that is more clear?

app/furniture/marketplace/checkout.rb
KellyAH: I assume: "get it before hand." from the user via this code change that allows the user to input their delivery address in a field.

app/furniture/marketplace/carts_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  class CartsController < Controller
+    def update
+      authorize(cart)
+      cart.update(cart_params)
+      respond_to do |format|
+        format.html do
+          redirect_to cart.room.location
+        end
+        format.turbo_stream do
+          render turbo_stream: [turbo_stream.replace(dom_id(cart), cart)]
```
KellyAH, line 11: Is this logic using turbo stream to AJAX-like update/rerender only the cart item elements on the webpage instead of rerendering the whole page?

app/furniture/marketplace/checkout.rb
zspencer: Yep

app/furniture/marketplace/carts_controller.rb
zspencer: Yep!

app/furniture/marketplace/carts/_footer.html.erb
KellyAH: wouldn't 2️⃣  `hiding the cart`, freak the customer out since they can't see their cart items. thus they don't know what is in their cart.

IMHO 1️⃣ makes it clear that some other required action is needed to be allowed to proceed to the checkout flow.

zspencer: I think we can *presume* a lot about what would or would not freak a customer out. It's not terribly unusual for clicking a link to navigate to a different page; so a customer freaking out because when they clicked a link they went somewhere else would surprise me.

spec/furniture/marketplace/carts_controller_request_spec.rb
KellyAH: leave it. it can be pulled out if more `it` statements are ever added.

spec/furniture/marketplace/cart_spec.rb
KellyAH: IMHO and nitpick: `product_b.price * 2` looks more clear and intentional.

app/furniture/marketplace/carts/_footer.html.erb
KellyAH: My hesitation on 2️⃣  is that I just couldn't visualize how it would look in the UI/UX. 

If the flow looks fine with the 2️⃣ change and is something you as a customer would be comfy with. I say do 2️⃣ 👍 😄

KellyAH (changes requested):
waiting for https://github.com/zinc-collective/convene/pull/1173/files#r1124122135 to be resolved, then I will approve this PR. The rest of the code looks great 💎 😄

app/furniture/marketplace/carts/_footer.html.erb
zspencer: Here is 2️⃣'s flow! I think it's pretty decent (but it needs better text! lol)

https://user-images.githubusercontent.com/50284/222829834-ff257c74-35e8-488c-a5cf-65e0e4e2fce1.mp4

spec/furniture/marketplace/carts_controller_request_spec.rb
zspencer: Wound up finding a bug during my testing, which made me want to write a second test case! So `let` it is!

spec/furniture/marketplace/cart_spec.rb
zspencer: Done!

KellyAH (approved):

PR #1178, opened 2023-03-03
🐞 `Space`: `SpacesController#show` works with `BYODomain`

zspencer:
- See also: https://github.com/zinc-collective/convene/pull/1171

I introduced this bug in b65a233c4d98c1304a744deb522aa75c8b451a13, basically the `SpaceController#show` was not finding by domain, so it would give an empty page.
    
This also adds support for using [`assert_select`](http://api.rubyonrails.org/classes/ActionDispatch/IntegrationTest.html) in request specs, so long as the `subject` is a `TestResponse`.


zspencer:
I am going to merge this because currently, no sites work :X :X :X

PR #1179, opened 2023-03-03
🧹 `Utility`: Use `UtilityHookup#location` in forms

zspencer:
- Extracted from htps://github.com/zinc-collective/convene/pull/1083

This is mostly for consistency and to start to close out a piece of work I started a while back but never finished.

### DIFF
--- a/app/models/utility_hookup.rb
+++ b/app/models/utility_hookup.rb
@@ -4,6 +4,7 @@
 class UtilityHookup < ApplicationRecord
   # @return [Space]
   belongs_to :space
+  self.location_parent = :space
 
   # Human-friendly name for disambiguating when a particular kind of {Hookup}
   # has multiple {UtilityHookup}s.
--- a/app/views/utility_hookups/_form.html.erb
+++ b/app/views/utility_hookups/_form.html.erb
@@ -1,4 +1,4 @@
-<%= form_with(model: [space, utility_hookup]) do |form| %>
+<%= form_with(model: utility_hookup.location) do |form| %>
   <%- form.object = form.object.utility %>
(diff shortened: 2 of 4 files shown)


app/models/utility_hookup.rb
```diff
@@ -4,6 +4,7 @@
 class UtilityHookup < ApplicationRecord
   # @return [Space]
   belongs_to :space
+  self.location_parent = :space
```
KellyAH, line 7: I see that`location_parent` was created for rails routing... but why do we need it?

https://github.com/zinc-collective/convene/blob/53130397f50b27c73cdcfacc0ce53133dc20db3b/app/models/within_location.rb#L6

zspencer: It allows us to provide a url hierarchy when leveraging polymorphic routes. For example: `/spaces/:space_id/utility_hookups` vs `/utility_hookups/`

KellyAH: ok within_location.rb  seems like a custom implementation and not something that come with rails.

I'll learn more about polymorphic routes on my own.

KellyAH (approved):

app/models/utility_hookup.rb
zspencer: Yea, rails does not include it by default because most `Rails` applications are flat; and it's [often an anti-pattern to have too deeply nested of resources](https://www.theodinproject.com/lessons/ruby-on-rails-advanced-topics#nested-routes)!

However, we need to keep models within the `Room`  or `Space`; so we almost always have at least those two levels of nesting; plus any additional domain-specific nesting that `Furniture` may want.

PR #1181, opened 2023-03-04
🧹 `Space`: Use `Space#location` in forms

zspencer:
- Extracted from: https://github.com/zinc-collective/convene/pull/1083

### DIFF
--- a/app/views/spaces/edit.html.erb
+++ b/app/views/spaces/edit.html.erb
@@ -33,7 +33,7 @@
 </fieldset>
 
 <fieldset>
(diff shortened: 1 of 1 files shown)


KellyAH (approved):
LGTM.

tho I'm still trying to figure out what [WithinLocation#location](https://github.com/zinc-collective/convene/blob/32877cdbdc2fa1ed7e3de566ec6d0926163d5c3b/app/models/within_location.rb#L18) does and why it's needed.

zspencer:
We may not "need" it, but it makes it a heck of a lot easier to adjust the url hierarchy.

PR #1182, opened 2023-03-04
🧹 `Room`: Use `Room#location` in form

zspencer:
- Extracted from https://github.com/zinc-collective/convene/pull/1083

### DIFF
--- a/app/views/rooms/_form.html.erb
+++ b/app/views/rooms/_form.html.erb
@@ -1,4 +1,4 @@
-<%= form_with model: [room.space, room], data: { controller: "room-form" }, local: true do |room_form| %>
+<%= form_with model: room.location, data: { controller: "room-form" } do |room_form| %>
   <fieldset>
     <div>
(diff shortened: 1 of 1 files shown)


app/views/rooms/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_with model: [room.space, room], data: { controller: "room-form" }, local: true do |room_form| %>
+<%= form_with model: room.location, data: { controller: "room-form" } do |room_form| %>
```
KellyAH, line 1: why `local: true` no longer needed?

KellyAH (approved):
I trust you Zee. as long as tests pass. The code refactor DOES look cleaner.

app/views/rooms/_form.html.erb
zspencer: It was needed back when we had both `turbo` and `rails-ujs` running in the project.

Issue #1151, opened 2023-03-01
🌫️`Gizmo`: `Usher`


zspencer:
In order to simplify `Room`s, I'd like us to get rid of the `locked` / `internal` etc. distinctions and instead leverage a piece of Furniture. 

## Fauxsonas

- Guest - Unauthenticated `Person`
- ??? - Authenticated `Person` who is not a `Member` of any `Space` int he `Neighborhood` (Perhaps this is `Guest` and `Visitor` is someone we know nothing about?
- Neighbor - Authenticated `Person` who is a `Member` of any `Space` in the `Neighborhood`
- Member - Authenticated `Person` who is a `Member` of this `Space`

## Use Cases

- [ ] Adding `Bouncer` to `Room`
- [ ] `Bouncer` prevents access to `Room`
  - [ ] By redirecting to a different `Room` when the `Room` is being opened by someone who is less than a...
    - [ ] A `Visitor`
    - [ ] A `Guest`
    - [ ] A `Neighbor`
    - [ ] A `Member`
    - [ ] A `Member` of a particular `Group`  - https://github.com/zinc-collective/convene/issues/811
  - [ ] By requiring an `Access Secret`

anaulin:
Quibble: a name that feels kinder than "bouncer" would be nice. Some ideas: usher, doorbot, doorkeeper, roomkeeper

zspencer:
I like Usher!

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1183, opened 2023-03-04
🧹 `Rooms`: Get rid of `locked`/`unlocked`/`internal` Rooms in favor of `public`/`internal`

zspencer:
- https://github.com/zinc-collective/convene/issues/1151

It's not really serving us, and is a hold-over from when Convene was
more video-forward.

Instead, we may want to create a `Furniture` that can "bounce" people
from a `Room` unless they know the secret code / are `Member`s of the
`Space` or in an appropriate `Group`.


app/models/room.rb
```diff
@@ -17,29 +17,19 @@ class Room < ApplicationRecord
...
-
-  def locked?
-    access_level&.to_sym == :locked
-  end
-
-  def unlocked?
-    access_level&.to_sym == :unlocked
-  end
+  # `internal` only Members may access the Room
+  attribute :access_level, :string, default: :public
```
anaulin, line 23: Not from this PR, and not blocking, but I'm surprised this is not an enum. It would save us from defining the `internal?` and `public?` methods, and also help make sure we don't accidentally set it to something that is not in the enum (like, say, during a refactor...).

```ruby
enum :access_level, {
  internal: "internal",
  public: "public"
}, default: :public
```

Then our one use of `Room::ACCESS_LEVELS.map(&:to_s)` becomes instead `Room.access_levels.values`.

spec/policies/room_policy_spec.rb
```diff
@@ -55,16 +42,14 @@
     it "includes all the rooms" do
       space = create(:space)
       internal_room = create(:room, :internal, space: space)
-      locked_room = create(:room, :locked, space: space)
-      unlocked_room = create(:room, :unlocked, space: space)
+      public_rom = create(:room, :public, space: space)
```
anaulin, line 45: `public_roOm`, here and on line 52

app/models/blueprint.rb
```diff
@@ -101,38 +101,27 @@ def space_attributes
...
           publicity_level: :listed,
-          access_level: :unlocked,
-          access_code: nil,
+          access_level: :public,
           furniture_placements: {
             markdown_text_block: {content: "# Welcome!"}
           }
         },
         {
           name: "Listed Room 2",
```
anaulin, line 110: Should we rename all of these to "Public", instead of "Listed"?

anaulin (approved):
Yay, great simplification!

app/models/room.rb
zspencer: I considered piling that into this PR, but dissuaded myself because of the weirdness I ran into with the ENUMs on Heroku last time.

app/models/blueprint.rb
zspencer: Oo, I remember why I didn't!!! If I had, it would have broken feature tests which have those names hardcoded... And technically they are *listed* they're just also *public*, since *listed* indicates whether or not they are enumerated in the navigation menu.

Issue #1137, opened 2023-02-24
🥔✨ `Marketplace`: `Tax` Collection and Compliance


zspencer:
- https://github.com/zinc-collective/convene/issues/831
- https://github.com/zinc-collective/convene/issues/1324
- https://github.com/zinc-collective/convene/issues/1326

## Use Cases

- [x] 🥔 `Seller` sets `TaxRate`s
  - [x] 🥔 As a %age of `Product#price`
  - [ ] 🌱 As a fixed amount per unit
  - [x] 🥔 `TaxRate`s are shared across a `Space`s `Marketplace`s
 - [x] 🥔 `Seller` associates `TaxRate`s to `Product`s 
- [x] #1469
- [x] 🥔 `Shopper` pays `Taxes`
- [x] #1655

zspencer:
I got a chance to speak with Adrionna Fike, the Board member at NoBAWC  who worked at Mandela up until last year. The lowdown was:

1. There are city and state level sales taxes
2. Most sales taxes are collected from the customer, but *some* sales taxes are collected in part from the customer *and* the seller (i.e. the sugary foods tax)
3. Some products may be covered under multiple sales-taxes (i.e. bottled beverages > 16oz have a tax *and* a deposit fee)

For now, we'll likely want to allow a Space to define the `SalesTax` they collect across all `Marketplaces`, for example:

- Sales Tax (City of Oakland): 6%
- Bottle Deposit (State of California): $0.10
- Tobacco Tax (State of California): 13%

And each `Product` can then be associated to multiple taxes on the `Marketplace::ProductsController#edit`

zspencer:
Do we want to group the taxes with the products when showing them? Or show them on a per-tax-rate basis? Right now we do:

```
Product A
Product B
Taxes Total
Delivery Fee
Total
```

but maybe it would make sense to organize it like:

```
Product A
  (Product A Tax)
Product B
Product C
  (Product C Tax)
Sub Total
Delivery Fee
Taxes 
Total
```

or

```
Product A
Product B
Product C
Product Total

Tax Rate A
Tax Rate B
Tax Rate Total

Delivery Fee

Total
```

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1186, opened 2023-03-05
✨ `Marketplace`: Create and Update `TaxRate`

zspencer:
  
Slaps some `update` and `create` actions together for setting `TaxRate`s.

That said, I  don't know that an integer or a float is better, since I suppose there could be a partial tax? 

Also, I tried to set it up so tax rates applied across all spaces; but then I ran into some weirdness... if y'all have ideas happy to hear.

### DIFF
--- a/app/furniture/marketplace/marketplaces/edit.html.erb
+++ b/app/furniture/marketplace/marketplaces/edit.html.erb
@@ -17,3 +17,11 @@
   <%= button_to t('marketplace.product.index'),
       marketplace.location(child: :products), method: :get, data: { turbo: false } %>
 <%- end %>
+
+<h3>Tax Rates</h3>
+<%= render policy_scope(marketplace.tax_rates) %>
+
+
+<%- if policy(marketplace.tax_rates.new).create? %>
+  <%= link_to("Add Tax Rate", marketplace.location(:new, child: :tax_rate)) %>
+<%- end %>
--- a/app/furniture/marketplace/marketplaces/edit.html.erb
+++ b/app/furniture/marketplace/marketplaces/edit.html.erb
@@ -21,7 +21,6 @@
 <h3>Tax Rates</h3>
 <%= render policy_scope(marketplace.tax_rates) %>
 
-
 <%- if policy(marketplace.tax_rates.new).create? %>
   <%= link_to("Add Tax Rate", marketplace.location(:new, child: :tax_rate)) %>
 <%- end %>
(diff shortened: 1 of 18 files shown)


app/furniture/marketplace/routes.rb
```diff
@@ -9,8 +9,12 @@ def self.append_routes(router)
           router.resource :checkout, only: [:show, :create]
         end
 
+        router.resources :tax_rates
+
         router.resources :orders, only: [:show]
-        router.resources :products
+        router.resources :products do
+          router.resources :tax_rates, only: [:create, :destroy, :index, :show]
```
anaulin, line 16: I only see code to edit and display the Marketplace Tax Rates, but there is this `products/tax_rates` set of routes, and also a reference from the marketplace products table to tax rates, although the `TaxRate` model merely `belongs_to :marketplace` -- maybe you are missing the `has_many :marketplace_products` on `TaxRate` and the `belongs_to :tax_rate, optional: true` on `MarketplaceProduct`? 

I'm not sure if the Tax Rates on Products are left over from a previous version, or if you intend to use those somehow differently?

db/migrate/20230305005417_create_marketplace_tax_rates.rb
```diff
@@ -0,0 +1,14 @@
+class CreateMarketplaceTaxRates < ActiveRecord::Migration[7.0]
+  def change
+    create_table :marketplace_tax_rates, id: :uuid do |t|
+      t.float :tax_rate
```
anaulin, line 4: Float feels correct to me, since I agree that it could be a percent, or a fractional amount, or an integer fixed amount. (Maybe eventually we'll end up with an enum here to say if it is %age or "fixed".)

anaulin (approved):
🚕

Looks good, but see below for my question re: taxes on products.

app/furniture/marketplace/marketplaces/edit.html.erb
```diff
@@ -17,3 +17,11 @@
   <%= button_to t('marketplace.product.index'),
       marketplace.location(child: :products), method: :get, data: { turbo: false } %>
 <%- end %>
+
+<h3>Tax Rates</h3>
+<%= render policy_scope(marketplace.tax_rates) %>
+
+
```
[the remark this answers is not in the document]
zspencer: Allegedly, `prettier` has an `erb` linter... *allegedly*

app/furniture/marketplace/routes.rb
zspencer: I should get rid of this since I haven't started the associating `TaxRate`s to `Product`s yet.

PR #1189, opened 2023-03-05
✨🌸 `Marketplace`: Tidier `Cart::DeliveryAddress` flow

zspencer:

We noticed that the `Update Address` and `Checkout` buttons were of equivalent size/important in the visual hierarchy; which seemed a bit wrong.

This gets rid of the bright call-to-action of the `Change Address` button, shrinks it in size, and aligns it with the label `Deliver to:` in the header.

### After!
Screenshot 2023-03-05 at 12 01 29 PM

### Before
Screenshot 2023-03-05 at 11 49 59 AM


anaulin (approved):
Nice!

Issue #1187, opened 2023-03-05
🌸 `DesignLibrary`


zspencer:
- https://github.com/zinc-collective/convene/issues/1520 use the https://github.com/zinc-collective/convene/issues/1187 to ease the implementation of https://github.com/zinc-collective/convene/issues/709 and provide a consistent user experience.

## Use Cases
- [x] 🥔 Browsable Component Previews at `/components`!
  - [ ] 🌱 That works on Production

## Components
- [x] 🥔`Button`
- [x] #2110
- [ ] #2216
- [ ] 🌱 Forms
  - [ ] https://github.com/zinc-collective/convene/issues/2082
  - [ ] 🌱 Text Field
  - [ ] 🌱 Number Field
  - [ ] 🌱 Percentage Field
  - [ ] 🌱 Submit
- [ ] #1381

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #943, opened 2022-11-17
Add View Component and create our first component: IconButtonComponent

anaulin:
This PR:
* adds View Component
* extracts a new component "Icon Button", based on our existing plus/minus/remove/etc buttons
* replace our existing `buttons/remove` partial with our new component

If this looks good, I can work on follow-up PRs to replacing our other "icon buttons" with the new component.

PR #1190, opened 2023-03-06
🛠️ `Component`: Set up `view_component` `Preview`s

zspencer:

This is an intermediary step before we look into `LookBook` or similar which:

1. Makes it possible to define component previews in `spec/components/previews`
2. Creates a basic layout for showing Components

### DIFF
--- a/app/views/layouts/component_preview.html.erb
+++ b/app/views/layouts/component_preview.html.erb
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title></title>
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <%= csrf_meta_tags %>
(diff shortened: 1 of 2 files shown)


KellyAH (approved):
looks like just setup changes to prep for future Lookbook/component changes/refactoring. 🤷‍♀️ Tests are passing so this doesn't seem to break anything. 

I look forward to seeing how this will be used for LookBook or something like it.

anaulin:
FWIW, I tried adding Lookbook previously, when I added the first ViewComponent, but we ran into some issues with live reloading and ended up removing it. You can see some of the partial work (and the removal) in the commits in https://github.com/zinc-collective/convene/pull/943 (there were other things we tried that did not get pushed up and are not in the commits, sadly).

PR #1191, opened 2023-03-06
🛠️🧹 `Component`: Sprout `CardComponentPreview`

zspencer:

This:

1. Adds a `CardComponentPreview`
2. Sprouts an `ApplicationComponent` which we can start to massage into playing well with tailwind's class syntax

(Originally was in https://github.com/zinc-collective/convene/pull/1190)

Screenshot 2023-03-05 at 6 14 16 PM Screenshot 2023-03-05 at 6 14 08 PM

### DIFF
--- a/app/components/application_component.rb
+++ b/app/components/application_component.rb
@@ -1,0 +1,19 @@
+class ApplicationComponent < ViewComponent::Base
+  attr_accessor :data
+  attr_writer :classes
+
+  def initialize(data: {}, classes: "")
+    self.data = data
+    self.classes = classes
+  end
+
+  # @todo this should gracefully merge left passed in data
+  def attributes(classes: "")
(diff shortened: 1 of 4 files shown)


app/components/application_component.rb
```diff
@@ -0,0 +1,19 @@
...
+  attr_writer :classes
+
+  def initialize(data: {}, classes: "")
+    self.data = data
+    self.classes = classes
+  end
+
+  # @todo this should gracefully merge left passed in data
+  def attributes(classes: "")
+    tag_builder.tag_options(data: data, class: self.classes(classes))
```
anaulin, line 12: Would

```suggestion
    tag_builder.tag_options(data: data, class: self.classes.merge(classes))
```

address the `@todo`?

(Assuming that we don't want the passed-in classes to be persisted in @classes, not sure if we do.)

anaulin (approved):

PR #1194, opened 2023-03-07
🛠️🌸 `Components`: Install `Lookbook` for Component Preview

zspencer:

Now we get even *shinier* components! Such shiny! Very browse! Much wow!

I went with the 2.0 beta because I figured it's close to prod.
Screenshot 2023-03-06 at 4 24 40 PM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,7 +41,9 @@ gem "redcarpet", "~> 3.6"
 # Breadcrumbs!
 gem "gretel", "~> 4.4"
 # Better UI components
+gem "lookbook", ">= 2.0.0.beta.4"
 gem "view_component", "~> 2.82"
+
 # Pagination!
 gem "pagy", "~> 6.0"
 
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -41,7 +41,9 @@ gem "redcarpet", "~> 3.6"
 # Breadcrumbs!
 gem "gretel", "~> 4.4"
 # Better UI components
+gem "lookbook", ">= 2.0.0.beta.4"
```
KellyAH, line 44: beta? is that stable?

anaulin (approved):
Hopefully this newer Lookbook version works better with Zeitwerk!

zspencer:
Yea, I'm worried about the autoloading; but 2.0 seems to do a bit better than the 1.7 I've used in other projects./

Gemfile
zspencer: I went with "stable enough" since it's a 2.0 beta.

PR #1196, opened 2023-03-07
✨🥔 `Marketplace`: Adding `TaxRate`s to `Product`s

zspencer:

There's a bit of low-hanging 🌸, but I don't want to block on them. In particular:

- It would be nice if we had a well-styled `Component` or partial for the `collection_check_boxes` form helper, since right now the `input[type="checkbox"]` and the `label` for each option are stacked on top of each other. Nasteh!
- There's probably a way to show the tax rate amount with a humanized percentage when selecting options, but I didn't want to figure it out

https://user-images.githubusercontent.com/50284/223316859-7f5fdc90-ace7-43f1-a576-ca7d9a651f10.mp4

### DIFF
--- a/app/furniture/marketplace/product.rb
+++ b/app/furniture/marketplace/product.rb
@@ -16,7 +16,8 @@ class Marketplace
     has_many :ordered_products, inverse_of: :product, dependent: :destroy
     has_many :orders, through: :ordered_products, inverse_of: :products
 
-    has_many :tax_rates, inverse_of: :products
+    has_many :product_tax_rates, inverse_of: :product
+    has_many :tax_rates, through: :product_tax_rates, inverse_of: :products
 
     attribute :name, :string
     validates :name, presence: true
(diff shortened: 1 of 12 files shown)


anaulin (approved):
This is ok, but I wonder if we really need the complexity of having another model and table for ProductTaxRates. It seems that we could:
* have only a set of Marketplace TaxRates
* some of those TaxRates could be attached to Products, so the tax can be applied to that particular Product if it is present in an order
* some of those TaxRates can have a type of "per order", so they are applied to each order (regardless of products in the order)

So we would end up with `Markeplace has_many :tax_rates` and also `Marketplace has_many :tax_rates`, those being the same model/table.

I'm finding it hard to reason about what makes sense here without some examples or a specification for the flows that we need to support with these new models.

zspencer:
Ah, there is only one kind of `TaxRate`, but the `ProductTaxRate` class is the Join table, because I get skeered of `has_and_belongs_to_many`

PR #1201, opened 2023-03-09
✨ `Marketplace`: Collect `Order#contact_phone_number` prior to `Checkout`

KellyAH:

Shopper can add a contact Phone Number to their order during the Checkout flow and it displays in the order_received_mailer.

added phone field to checkout page

phone field and updated button text

**mobile mode**
fields stack in mobile mode

**mailer**
add phone to mailer

Co-authored-by: Ana <anaulin@users.noreply.github.com>
Co-authored-by: Dalton Pruitt <daltonrpruitt@users.noreply.github.com>

### DIFF
--- a/app/furniture/marketplace/cart.rb
+++ b/app/furniture/marketplace/cart.rb
@@ -40,7 +40,7 @@ class Marketplace
     end
 
     def ready_for_delivery?
-      self.delivery_address.present? && self.contact_phone_number.present? ? true: false
+      (delivery_address.present? && contact_phone_number.present?)
     end
   end
 end
--- a/app/furniture/marketplace/cart.rb
+++ b/app/furniture/marketplace/cart.rb
@@ -38,5 +38,9 @@ class Marketplace
     def price_total
       product_total + delivery_fee
     end
+
+    def ready_for_delivery?
+      self.delivery_address.present? && self.contact_phone_number.present? ? true: false
+    end
   end
 end
--- a/app/furniture/marketplace/cart.rb
+++ b/app/furniture/marketplace/cart.rb
@@ -16,6 +16,7 @@ class Marketplace
     has_many :products, through: :cart_products, inverse_of: :carts
 
     has_encrypted :delivery_address
+    has_encrypted :contact_phone_number
 
     enum status: {
       pre_checkout: "pre_checkout",
--- a/app/furniture/marketplace/carts/_cart.html.erb
+++ b/app/furniture/marketplace/carts/_cart.html.erb
@@ -5,7 +5,7 @@
       <%= render "text_field", attribute: :delivery_address, form: cart_form %>
       <%= render "text_field", attribute: :contact_phone_number, form: cart_form %>
 
-      <%= cart_form.submit t('marketplace.edit_delivery_info.edit') %>
+      <%= cart_form.submit t('marketplace.delivery_info.edit') %>
     <%- end %>
   <%- else %>
     <div class="flex justify-between flex-wrap">
--- a/app/furniture/marketplace/carts/_cart.html.erb
+++ b/app/furniture/marketplace/carts/_cart.html.erb
@@ -9,11 +9,13 @@
     <%- end %>
   <%- else %>
     <div class="flex justify-between flex-wrap">
-      <div class="w-full flex items-center justify-between text-sm">
+      <div class="sm:w-1/2 w-full flex items-center justify-between text-sm">
         <span>Delivering to:</span>
         <%= cart.delivery_address %>
         <%= render ButtonComponent.new(href: cart.location.concat([params: { cart: { delivery_address: nil } }]), label: t('marketplace.delivery_address.edit'), title: nil, classes: "shrink font-light text-xs m-0 bg-primary-200 underline") %>
+      </div>
 
+      <div class="sm:w-1/2 w-full flex items-center justify-between text-sm">
         <span>Contact Phone Number:</span>
         <%= cart.contact_phone_number %>
         <%= render ButtonComponent.new(href: cart.location.concat([params: { cart: { contact_phone_number: nil } }]), label: t('marketplace.contact_phone_number.edit'), title: nil, classes: "shrink font-light text-xs m-0 bg-primary-200 underline") %>
--- a/app/furniture/marketplace/carts/_cart.html.erb
+++ b/app/furniture/marketplace/carts/_cart.html.erb
@@ -1,20 +1,23 @@
 <div id="<%= dom_id(cart) %>">
 
-  <%- if cart.delivery_address.blank? %>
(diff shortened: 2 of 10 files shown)


app/furniture/marketplace/carts/_cart.html.erb
```diff
@@ -1,20 +1,23 @@
...
   <%- else %>
     <div class="flex justify-between flex-wrap">
       <div class="w-full flex items-center justify-between text-sm">
         <span>Delivering to:</span>
+        <%= cart.delivery_address %>
         <%= render ButtonComponent.new(href: cart.location.concat([params: { cart: { delivery_address: nil } }]), label: t('marketplace.delivery_address.edit'), title: nil, classes: "shrink font-light text-xs m-0 bg-primary-200 underline") %>
-      </div>
-
 
-      <p class="grow"><%= cart.delivery_address %></p>
```
zspencer, line 17: If this was a `div` that was a sibling of the `div` that contains the "Delivering To" text, and had the same classes it would likely stack on smaller screens. 

We may need to add `sm:w-1/2` or `sm:w-auto` to make it take up half the space on larger screens

app/furniture/marketplace/carts/_footer.html.erb
```diff
@@ -8,7 +8,7 @@
   <tr>
     <td class="text-left px-3 py-3.5 text-right" colspan="8">
       <%= render "marketplace/carts/total", cart: cart %>
-      <%- if cart.delivery_address.present? %>
+      <%- if cart.delivery_address.present? && cart.contact_phone_number.present?%>
```
zspencer, line 11: It may be nice to have a `cart.ready_for_delivery?` or something that contains these checks

zspencer (approved):
Very nice!

One thing I would like to see is the notification email being duplicated!

spec/furniture/marketplace/carts_controller_request_spec.rb
```diff
@@ -7,7 +7,7 @@
 
   describe "#update" do
     subject(:perform_request) do
-      put polymorphic_path(cart.location), params: {cart: {delivery_address: "123 N West St"}}
+      put polymorphic_path(cart.location), params: {cart: {delivery_address: "123 N West St", contact_phone_number: "(415)-123-4567"}}
```
zspencer, line 10: It would be nice if there was a `specify` block that asserted this actually changed.

app/furniture/marketplace/order_received_mailer/notification.html.erb
```diff
@@ -10,3 +10,4 @@
 <p>Total: <%= humanized_money_with_symbol(order.price_total) %></p>
 <p>Deliver To: <%= order.delivery_address %>
 <p>Contact Email: <%= order.contact_email %>
+<p>Contact Phone Number: <%= order.contact_phone_number %>
```
KellyAH, line 13: BUG: `Contact Phone Number:` fails to render in the mailer. `order.contact_phone_number` is always nil.

Is the `cart.contact_phone_number` supposed to be passed to the `order.contact_phone_number`?

zspencer: I would recommend tracing:
- Down through the application layer to the storage layer to determine where both `Cart#contact_phone_number` and `Order#contact_phone_number` is stored , and
- Along the interaction layer to what `Controller#action` would be responsible for passing data from `Cart` to `Order` 

My current guess is that `Order.contact_phone_number` may not have the lockbox settings?

app/furniture/marketplace/cart.rb
```diff
@@ -37,5 +38,9 @@ def delivery_fee
     def price_total
       product_total + delivery_fee
     end
+
+    def ready_for_delivery?
+      self.delivery_address.present? && self.contact_phone_number.present? ? true: false
```
zspencer, line 43: Technically `delivery_address.present? && contact_phone_number.present?` is equivalent, and will return `true` when both are present and `false` when one is not.

app/furniture/marketplace/carts/_cart.html.erb
```diff
@@ -1,20 +1,25 @@
 <div id="<%= dom_id(cart) %>">
 
-  <%- if cart.delivery_address.blank? %>
+  <%- if cart.delivery_address.blank? || cart.contact_phone_number.blank? %>
     <%= form_with model: cart.location do |cart_form| %>
       <%= render "text_field", attribute: :delivery_address, form: cart_form %>
+      <%= render "text_field", attribute: :contact_phone_number, form: cart_form %>
 
-      <%= cart_form.submit t('marketplace.delivery_address.edit') %>
+      <%= cart_form.submit t('marketplace.edit_delivery_info.edit') %>
```
zspencer, line 8: Nit: `marketplace.delivery_info.edit` is less duplicaive than `marketplace.edit_delivery_info.edit`

app/furniture/marketplace/carts/_cart.html.erb
zspencer: Nice!

app/furniture/marketplace/order_received_mailer/notification.html.erb
zspencer: Nah, looks like lockbox is set correctly; Have you updated the `OrderReceivedMailerPreview` to include a contact_phone_number? It is likely that the `Preview` calls the `factory`, so you'll want to search for where we define the `marketplace_order` factory and add a `Faker` call for the phone_number...

spec/furniture/marketplace/carts_controller_request_spec.rb
zspencer: So good!

zspencer (approved):
Looking good! Just gotta run the lintah and confirm that the `MailerPreview` has a phone number!

app/furniture/marketplace/cart.rb
KellyAH: 1. Is this condition coded for the behavior we want? 
2. we want `ready_for_delivery?` to evaluate to `true` when both `delivery_address` and `contact_phone_number` exists right? And if either `delivery_address` or  `contact_phone_number` are missing, it evaluates to `false`, right?

app/furniture/marketplace/carts/_cart.html.erb
KellyAH: good catch, I missed that. thanks. fixed in 2e304dfe65982b217c7aea7889a44d984214f9c3

app/furniture/marketplace/cart.rb
KellyAH: nevermind I think I understand now. and Rubocop caught the issue too. my ternary was redundant since `.present` already evaluates to a boolean.

KellyAH: fixed in 21395a5729b4858daab4cb3d524a1f6ef0f32e5d

PR #1207, opened 2023-03-10
✨🥔 `Marketplace`: `Shopper` and `Distributor` `Order` History

zspencer:

`Neighbor`s can see all their past `Order`s! And `Member`s can see all the `Order`s for a given `Marketplace`
Screenshot 2023-03-10 at 3 36 32 PM
Screenshot 2023-03-10 at 3 36 38 PM

It would be lovely if someone would take the `_order` partial and turn it into a `ViewComponent` that looks v pretty, instead of this hideous table monster I spawned.


anaulin (approved):
🚢

PR #1209, opened 2023-03-11
🧹 `Furniture`: Complete the rename!

zspencer:
- Requires: https://github.com/zinc-collective/convene/pull/1208

OK, so I think this is a straight-up-rename; but there may be stuff that wants fixing...


.rubocop_todo.yml
```diff
@@ -41,11 +41,11 @@ RSpec/ExampleLength:
     - 'spec/furniture/journal/entries_request_spec.rb'
     - 'spec/furniture/markdown_text_block_spec.rb'
     - 'spec/furniture/marketplace/products_controller_request_spec.rb'
-    - 'spec/models/furniture_placement_spec.rb'
+    - 'spec/models/furniture_spec.rb'
     - 'spec/models/membership_spec.rb'
     - 'spec/models/room_spec.rb'
     - 'spec/policies/room_policy_spec.rb'
-    - 'spec/requests/furniture_placements_controller_request_spec.rb'
+    - 'spec/requests/furnitures_controller_request_spec.rb'
```
anaulin, line 48: You could just un-lint these 😉

anaulin (approved):
Yes!!!!

.rubocop_todo.yml
zspencer: I was afeared of trying to do even more in this giganormoustrosity

anaulin: If you put it in a separate commit I won't object!

But up to you, fine either way.

PR #1210, opened 2023-03-11
🧹 `Utility`: Eliminate the `Utilities` module!

zspencer:
- https://github.com/zinc-collective/convene/issues/252

OK, this should get us close to renaming `UtilityHookup` to `Utility`!


anaulin (approved):
Ah, so much better!

You should never listen to my naming suggestions. They are terrible.

PR #1211, opened 2023-03-11
🧹 `Furniture`: Appease the Fuzz

zspencer:
- https://github.com/zinc-collective/convene/pull/1209#discussion_r1133016759

I don't really like the direction this pushes specs, but I can live with it.

In particular, a single assert per request/response seems... excruciatingly slow.

But there are other mechanisms we can explore to mitigate that as it gets out of hand.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -23,7 +23,6 @@ RSpec/AnyInstance:
 # Prefixes: when, with, without
 RSpec/ContextWording:
   Exclude:
-    - 'spec/furniture/journal/entry_policy_spec.rb'
     - 'spec/models/authenticated_session_spec.rb'
     - 'spec/support/shared_examples/a_space_member_only_route.rb'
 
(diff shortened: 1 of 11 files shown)


anaulin (approved):
Thank you!

I am also not fond of the "one assert per case" rule, and would be in favor of simply disabling it in our settings (which I know is a little bit against the whole point of using standard, but I think for Rspec-specific rules like this is justified).

zspencer:
I would be ecstatic if I can have my multi-expect it's again. `standard` doesn't cover rubocop-rspec, AFAIK so I don't think it's unreasonable for us to tweak that.

anaulin:
I went to disable that particular cop and discovered that it is already disabled -- you disabled it 4 months ago: https://github.com/zinc-collective/convene/blame/main/.rubocop.yml#L19-L20

I reverted your commit locally and I didn't see any warnings for multiple expectations. 🤷🏼‍♀️

zspencer:
Yea, I think it was the "Number of Lines per Spec" cop... I just didn't read.

Issue #1185, opened 2023-03-05
🥔✨ `Marketplace`: `DeliveryConstraint`s


zspencer:
- https://github.com/zinc-collective/convene/issues/831
- https://github.com/zinc-collective/convene/issues/1325

`Marketplace` `Vendor`s often offer multiple options for `Delivery`, for example a catering option where `Shopper`s can set the date and time and a pre-scheduled option where the `Distributor` picks up a set of `Order`s from from the `Vendor` and brings them to the `Shopper`s at a set date and time.

## Use Cases
- [x] 🥔 `Vendor` or `Distributor` sets Soft Timing Constraints with words
- [ ] 🌱 `Vendor` or `Distributor` sets hard Timing Constraints 
  - [ ] Delivery during Specific Hours (i.e. Mon-Fri 10 to 12)
  - [ ] Order By (i.e. 48 hours before delivery)
- [x] 🥔 `Vendor` specific Date and Time ("Sellers Choice")
- [x] 🥔 `Shopper` chooses Date and Time ("Shoppers Choice?")
  - [x] 🥔 `Vendor` or `Distributor` enables `Shoppers Choice` Delivery
      - Right now, this is implict by clearing the `Marketplace#delivery_window`, we probably should make it explicit...
  - [x] 🥔 `Shopper` sets Delivery Time
    - [ ] 🌱 ... As enforced by the Timing Constraints

zspencer:
Since Piikup is going for Catering first, For now, Piikup and Caterers need a 48+ hour notice. We'll also want to let them select the date and time for delivery.

zspencer:
Hard timing constraints (in particular "place orders 48 hours before delivery") seems pretty important to Piikup's client, Cafe Gabriela

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1218, opened 2023-03-16
✨ `Marketplace`: `Distributor` may set `Marketplace#delivery_window`

zspencer:

Added a `Marketplace#delivery_window` attribute, which is a plain-old-string for now. It's going to quickly become insufficient, because we *may* need to do date and time crap at one point.

But "quickly" could mean "months from now!" sooooooo.


KellyAH (approved):
LGTM

PR #1220, opened 2023-03-16
✨ `Neighborhood`: Make `money_field` look more like `text_field`

zspencer:

A super cool person may want to actually make this a Component, which would be super cool.

### After
Screenshot 2023-03-15 at 7 06 21 PM

### Before
Screenshot 2023-03-15 at 7 06 47 PM


anaulin (approved):
🌷

PR #1221, opened 2023-03-16
`Furniture`: Add a way for furniture to signal they have their own controller, and add link to kind-specific `edit` from Room edit form

anaulin:
I noticed that the Marketplace was not showing the `edit` gear in the Room edit form, so this PR fixes that. This is not the most elegant code I've ever written, but it communicates the intent and works. If we like this approach, we can polish it to have better method naming, etc.

Also took the chance to improve the Room `edit` page look a bit.

### Before
image

### After
image

### DIFF
--- a/app/furniture/markdown_text_block.rb
+++ b/app/furniture/markdown_text_block.rb
@@ -4,6 +4,8 @@
 class MarkdownTextBlock < Furniture
   include RendersMarkdown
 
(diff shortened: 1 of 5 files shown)


zspencer (approved):
I love it!

That said, we probably should get rid of the magical pony for furniture form generation, since it does not appear particularly useful and it is very confusing.

PR #1222, opened 2023-03-16
🌸 `Marketplace`: Tidy the `edit` form quite a bit

zspencer:


anaulin (approved):
💄

app/furniture/marketplace/marketplaces/edit.html.erb
```diff
@@ -1,26 +1,25 @@
 <%- breadcrumb :edit_marketplace, marketplace %>
 
-<%= render "form", marketplace: marketplace %>
+<div class="max-w-2xl self-stretch mx-auto">
```
anaulin, line 3: `max-w-prose` instead of `max-w-2xl` might be neat as well (we can play with that in a separate PR tho)

PR #1232, opened 2023-03-19
Set `Order` datetime based off actual order creation

KellyAH:

Currently the `Order` datetime is based off when the `Cart` was created and not when the `Order` was created.

This fixes that by setting `placed_at` for `Order` after the order is successfully created in Stripe.

note: The `Cart` is created when the Marketplace page is loaded.


zspencer (approved):
Yes! Can't wait to see the follow up on `Marketplace::Orders#index`!!!!

PR #1233, opened 2023-03-20
✨🥗 `Marketplace`: Show `Tax` on `OrderReceivedMailer`

zspencer:

I took some time to:

- Tidy up the layout
- Write a subject line

So this should be pretty functional!

Stuff that's missing (because we don't have it):

1. Contact Name
2. Pickup Time
3. Delivery Time

I still feel dissatisfied with how we present the tax collected (Like I think we should break it out by type mayhaps?); but I think it's probably OK for now?

## After

Screenshot 2023-03-20 at 11 16 07 AM

## Before
Screenshot 2023-03-20 at 11 16 47 AM

### DIFF
--- a/app/furniture/marketplace/locales/en.yml
+++ b/app/furniture/marketplace/locales/en.yml
@@ -19,6 +19,8 @@ en:
         failure: "Marketplace could not be updated."
     order:
       index: "Order History"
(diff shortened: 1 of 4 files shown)


anaulin (approved):
💌

PR #1240, opened 2023-03-20
🐞 `Marketplace`: `StripeEvent#create` was failing

zspencer:
- https://github.com/zinc-collective/convene/issues/1236

It looks like the `PaymentIntent` shape changes depending on what features we request. In particular, `charges` is sometimes absent and since we're not collecting the address from them on `Checkout` it's not included.

It may be worth taking time to write a shot-across-the-bow feature test that:

1. Sets up Stripe in a Space
2. Sets up the Marketplace in a Space
3. Places an Order
4. Confirms the Order was received and emails went out

But that also sounds hard.

### DIFF
--- a/app/furniture/marketplace/stripe_events_controller.rb
+++ b/app/furniture/marketplace/stripe_events_controller.rb
@@ -17,6 +17,7 @@ class Marketplace
         order = Order.find(payment_intent.transfer_group)
 
         return if order.paid?
+
         latest_charge = Stripe::Charge.retrieve(payment_intent.latest_charge, api_key: marketplace.stripe_api_key)
         balance_transaction = Stripe::BalanceTransaction.retrieve(latest_charge.balance_transaction, api_key: marketplace.stripe_api_key)
 
--- a/app/furniture/marketplace/stripe_events_controller.rb
+++ b/app/furniture/marketplace/stripe_events_controller.rb
@@ -17,18 +17,10 @@ class Marketplace
         order = Order.find(payment_intent.transfer_group)
 
         return if order.paid?
+        latest_charge = Stripe::Charge.retrieve(payment_intent.latest_charge, api_key: marketplace.stripe_api_key)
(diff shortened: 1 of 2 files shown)


spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -9,25 +9,27 @@
...
-          address: double(line1: "123 N West", line2: "apt 1", city: "Oakland", state: "CA", postal_code: "94609")))))
+      data: double(object: double(payment_intent: "pi_1234", customer_details: double(email: "test@example.com"))))
   end
 
   let(:balance_transaction) { double(Stripe::BalanceTransaction, fee: 100) }
 
   let(:payment_intent) do
     double(Stripe::PaymentIntent, transfer_group: order.id,
-      charges: [double(Stripe::Charge, balance_transaction: "btx_2234")])
+      latest_charge: "ch_1234")
```
anaulin, line 19: This probably all fits on the previous line.

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -17,18 +17,10 @@ def create
         order = Order.find(payment_intent.transfer_group)
 
         return if order.paid?
+        latest_charge = Stripe::Charge.retrieve(payment_intent.latest_charge, api_key: marketplace.stripe_api_key)
```
anaulin, line 20: Nit: a blank line after the guard clause would be nice

anaulin (approved):

app/furniture/marketplace/stripe_events_controller.rb
zspencer: Good call!

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
zspencer: It do!

PR #1250, opened 2023-03-23
🔐 `Marketplace`: Step 3 for Encrypting `Order#contact_email`

zspencer:

I skipped Step 2 Because I forgot that you're supposed to do the `ignore_columns` *prior* top dropping the columns but eff it.


anaulin (approved):
yassss

PR #1251, opened 2023-03-23
🥗 `Marketplace`: Capture `Cart#delivery_window` as a DateTime

zspencer:
We probably want to do something a bit smerter with like, an `ActiveRecord::Attribute`, but for now this lets us handle nil or unparsed values more effectively

### After:
Screenshot 2023-03-22 at 5 46 04 PM
Screenshot 2023-03-22 at 5 45 45 PM

### Before:

Screenshot 2023-03-22 at 5 47 27 PM
Screenshot 2023-03-22 at 5 47 18 PM

### DIFF
--- a/app/furniture/marketplace/cart.rb
+++ b/app/furniture/marketplace/cart.rb
@@ -29,7 +29,12 @@ class Marketplace
     end
 
     def delivery_window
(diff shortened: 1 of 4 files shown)


anaulin (approved):
🕐

PR #1253, opened 2023-03-23
🌸 `Neighborhood`: Tidy up "New Utility" form and tweak global form styles

anaulin:
Another small UI tweak, including some tweaks to global form styles (based on a Tailwind UI component)

### Before
New Utility form:
image

Example other form:
image

### After
New Utility form:
image

Example other form:
image

### DIFF
--- a/app/assets/stylesheets/components/form.scss
+++ b/app/assets/stylesheets/components/form.scss
@@ -10,7 +10,7 @@
   form {
     @apply inline;
     div {
-      @apply mb-2;
+      @apply mb-3;
     }
     label,
     input,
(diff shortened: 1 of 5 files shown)


zspencer (approved):
Nice! Is there any chance you'd be available to inline the style definitions so we can delete the `form.scss` file?

anaulin:
@zspencer I keep going back and forth on that. Do we want to inline the style definitions? It seems kinda nice to not have to define them on every single call, and creating separate partials or components for every label, field, etc fields like a lot, but maybe that's a better way to go?

I'm not sure what's the current "better practices" thinking on all that. If you think we should be inlining all the styles and creating shareable components, I'm happy to go down that route.

zspencer:
I think moving everything into partials and components and ditching the webpack build is a nice way to clean things up and make it a bit easier while we do prototyping. I can imagine a world where we package some default styling as we add support for themes, but that's not likely to happen anytime soon.

PR #1254, opened 2023-03-23
🧹 `Furniture`: Add a facility to add attribute-like fields backed by `settings`

anaulin:
This is a proposal for tightening our manually-added "settings-backed attribute" methods. What do y'all think?

Also maybe this is partially duplicative with the [existing `write_attribute` method](https://github.com/zinc-collective/convene/blob/main/app/models/furniture.rb#L49-L53)? Maybe a better approach is to add a `read_attribute`? (if that exists?)

### DIFF
--- a/app/furniture/marketplace/marketplace.rb
+++ b/app/furniture/marketplace/marketplace.rb
@@ -11,6 +11,9 @@ class Marketplace
     has_many :tax_rates, inverse_of: :marketplace
 
     setting :notify_emails
+    setting :order_by
+    setting :stripe_webhook_endpoint
+    setting :stripe_webhook_endpoint_secret
 
     def has_controller_edit?
       true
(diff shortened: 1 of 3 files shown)


app/furniture/marketplace/marketplace.rb
```diff
@@ -10,6 +10,8 @@ class Marketplace < Furniture
 
     has_many :tax_rates, inverse_of: :marketplace
 
+    settings_attribute :notify_emails
```
zspencer, line 13: I feel like `setting :notify_emails` is sufficient and shorter; but I feel like `settings_attribute` is already a big win.

zspencer (approved):
Re: `read_attribute`: I think that `read_attribute` is intended to support some Railsy stuff that may not be necessary anymore. 

If we delete it, what happens?

Re: This change 

I like it! It gives us the syntactical sugar to make `settings` a bit less of a train-wreck so we can make more improvements over time.

app/furniture/marketplace/marketplace.rb
anaulin: Renamed!

anaulin:
Well, clearly `write_attribute` is still needed -- removing it breaks some stuff: https://github.com/zinc-collective/convene/actions/runs/4496512048/jobs/7911257480

app/models/furniture.rb
```diff
@@ -83,4 +83,15 @@ def self.from_placement(placement)
   def to_kind_class
     becomes(Furniture.registry.fetch(furniture_kind.to_sym))
   end
+
+  # Adds a writer and a reader for a value backed by `settings`
+  def self.setting(setting_name)
```
zspencer, line 88: A V minor nit that I should have brought up yesterday and is more related to dealing with high-allocations in non-Convene Rails applications; but the unboxing of the symbol to a string could happen once at `self.setting` time rather than 2x when calling `setting` time and Nx when calling the setting methods. I'm ashamed that I even think about this now 🤣

app/furniture/marketplace/marketplace.rb
```diff
@@ -10,6 +10,11 @@ class Marketplace < Furniture
 
     has_many :tax_rates, inverse_of: :marketplace
 
+    setting :notify_emails
+    setting :order_by
+    setting :stripe_webhook_endpoint
+    setting :stripe_webhook_endpoint_secret
```
zspencer, line 16: SO GOOD THANK YOU!

zspencer (commented):
Yahhhh!!!

PR #1257, opened 2023-03-23
🌸 `Neighborhood`: Apply standard form input styling to date-time input

anaulin:
We forgot to style date-time inputs in the same way as all the other inputs. This PR fixes that.

### Before
image

### After
image


zspencer (approved):
Thank you! Nice catch!

PR #1259, opened 2023-03-24
🧹 `Furniture`: Improve `Furniture#setting` helper, and migrate all Furniture kinds to use it

anaulin:
As a follow-up to https://github.com/zinc-collective/convene/pull/1254, this PR:
* updates the `Furniture#setting` method to only convert the setting name to string once
* extends `Furniture#setting` to allow setting a default value
* updates all existing Furniture to use the `Furniture#setting` method to define their settings

### DIFF
--- a/app/furniture/marketplace/marketplace.rb
+++ b/app/furniture/marketplace/marketplace.rb
@@ -10,8 +10,13 @@ class Marketplace
 
     has_many :tax_rates, inverse_of: :marketplace
 
+    setting :delivery_fee_cents, default: 0
+    monetize :delivery_fee_cents
+    setting :delivery_window
     setting :notify_emails
     setting :order_by
+    setting :stripe_account
+    alias_method :vendor_stripe_account, :stripe_account
     setting :stripe_webhook_endpoint
     setting :stripe_webhook_endpoint_secret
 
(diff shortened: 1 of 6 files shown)


app/furniture/marketplace/marketplace.rb
```diff
@@ -30,53 +34,26 @@ def stripe_api_key?
       false
     end
 
-    # @todo Probably want to rename this for-reals but lazy
-    def stripe_account
-      settings["stripe_account"]
-    end
-    alias_method :vendor_stripe_account, :stripe_account
```
zspencer, line 37: If we do this, we'll need to migrate from the old setting to this new one somehow, since right now the stripe accounts are stored in the `stripe_account` key of the `settings` hash.

Instead, we may want to do `setting :stripe_account` and keep the alias.

app/furniture/marketplace/vendor_stripe_accounts_controller.rb
```diff
@@ -1,5 +1,5 @@
 class Marketplace
-  class StripeAccountsController < FurnitureController
+  class VendorStripeAccountsController < FurnitureController
```
zspencer, line 2: I'm not sold on this rename, since I feel like `StripeAccount` is the actual domain model here; and the `vendor_stripe_account` is an indicator of who the `StripeAccount` in the `Marketplace` is connected to. We could wind up with `distributor_stripe_account` for use cases where Arizmendi works with Piikup and the local bicycle courier coop.

zspencer (approved):
I'm a little worried about losing data, and about the class-level rename for `StripeAccount` but I don't think either of those are blocking so long as we do a data migration.

spec/models/furniture_spec.rb
```diff
@@ -43,7 +43,7 @@
 
   describe "#setting" do
     class TestFurniture < described_class # rubocop:disable Lint/ConstantDefinitionInBlock, RSpec/LeakyConstantDeclaration
-      setting :name
+      setting :name, default: "some default value"
```
zspencer, line 46: Very nice!

anaulin:
@zspencer ok, I reverted the rename to `stripe_account`, and instead I'm just leaving the method alias in place. Do you want to take another look?

app/furniture/marketplace/marketplace.rb
anaulin: I thought we weren't yet at the point where we cared about migrating data for Marketplaces, since it is not used by anyone?

In any case, I reverted this renaming.

zspencer (approved):
Looks great! Thank you!

PR #1262, opened 2023-03-25
Update Webpack to 5.76.0 to address CVE-2023-28154

anaulin:
Addresses CVE-2023-28154 , resolves Dependabot alert https://github.com/zinc-collective/convene/security/dependabot/57

PR #1264, opened 2023-03-25
Update yarn.lock file to fix broken Heroku build

anaulin:
Latest Heroku auto-deploy failed with the errors:
```
----> Installing dependencies
       Installing node modules (yarn.lock)
       yarn install v1.22.19
       [1/4] Resolving packages...
       error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
       info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.

-----> Build failed
 !     Outdated Yarn lockfile
       Your application contains a Yarn lockfile (yarn.lock) which does not
       match the dependencies in package.json. This can happen if you use npm
       to install or update a dependency instead of Yarn.
       Please run the following command in your application directory and check
       in the new yarn.lock file:
       $ yarn install
       $ git add yarn.lock
       $ git commit -m "Updated Yarn lockfile"
       $ git push heroku main
```

I think I merged https://github.com/zinc-collective/convene/pull/1262 after tweaking `package.json`, but forgot to run `yarn install` to update the `.lock` file. This fixes that.


anaulin:
Merging to verify that this unbreaks deploys

zspencer:
Weird, but glad it was easy to fix.

PR #1266, opened 2023-03-25
🐞 `Marketplace`: fix `delivery_window` type issues by giving more structure (and types!) to `Marketplace#settings`

anaulin:
Fixes https://github.com/zinc-collective/convene/issues/1265 by using `StoreModel` to model `Marketplace#settings` with a bit more structure, including types.

### DIFF
--- a/spec/furniture/marketplace/marketplaces_controller_request_spec.rb
+++ b/spec/furniture/marketplace/marketplaces_controller_request_spec.rb
@@ -23,13 +23,16 @@ RSpec.describe Marketplace::MarketplacesController, type: :request do
   end
 
   describe "#update" do
-    before { sign_in(space, member) }
+    before do
+      sign_in(space, member)
+      freeze_time
+    end
 
     it "updates the attributes" do
-      put polymorphic_path(marketplace.location), params: {marketplace: {delivery_fee: 50.00, delivery_window: "Tomorrow, mahhhhn...", order_by: "10AM"}}
+      put polymorphic_path(marketplace.location), params: {marketplace: {delivery_fee: 50.00, delivery_window: 2.hours.from_now, order_by: "10AM"}}
 
       expect(marketplace.reload.delivery_fee_cents).to eq(50_00)
-      expect(marketplace.reload.delivery_window).to eq("Tomorrow, mahhhhn...")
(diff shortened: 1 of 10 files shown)


zspencer (approved):
I really like this, I am struggling to come up with a coherent way to express that we're probably going to need to make the `DeliveryWindow` a real Domain Object soon because there's a lot of complexity we'll need to think about.

For example:

1. Acta Non Verba delivers *only* on Wednesday from 1 to 6
2. Table Catering delivers *any day of the week* between 8??am and 9??pm

The way I'm kind of imagining modeling this is that the `DeliveryWindow` would have:

- The Days Of Week (i.e. Monday, Thursday, Friday)
- The Start Hour and End Hour (i.e. 2pm and 5pm)

And for now, we can work from the assumption that a Marketplace can have only one, but I can also see a world where we have to let them fine tune further; so we may want to be prepared to promote it from a 1:1  `Marketplace`:`DeliveryWindow` relationship to a 1:n.

spec/furniture/marketplace/marketplaces_controller_request_spec.rb
```diff
@@ -23,13 +23,16 @@
   end
 
   describe "#update" do
-    before { sign_in(space, member) }
+    before do
+      sign_in(space, member)
+      freeze_time
```
zspencer, line 28: Does this mean time is frozen for the remainder of the thread? Should we be using `around` instead of `before`?

anaulin: My understanding is that the default behavior is that it automatically calls `travel_back` after the tests, so time gets unfrozen at that point: https://api.rubyonrails.org/v5.2.4/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-after_teardown

anaulin:
I agree that we will need to make `DeliveryWindow` (or whatever name) its own object. The intent of this PR is mainly to stop throwing 500s.

spec/furniture/marketplace/marketplaces_controller_request_spec.rb
zspencer: :the-more-you-know.gif:

PR #1268, opened 2023-03-27
🧹 `Marketplace`: Consolidate Order Emails

zspencer:
- https://viewcomponent.org/guide/slots.html
- https://thoughtbot.com/blog/building-custom-rspec-matchers-with-regular-objects

OK, so there's a few things:

1. I'm grouping the `Placed` and `Received` Mailers under the `Marketplace::Order` namespace
2. I pulled out a `EmailReceiptComponent` that takes responsibility for the shared HTML
3. I defined a factory for building a full `Order` with products, taxes, and all delivery info
4. I defined a number of custom matchers for use in `Mailer` specs, one of which may be useful in `Component` and `Request` specs.

## `Placed`
Screenshot 2023-03-27 at 11 34 53 AM

## `Received`

Screenshot 2023-03-27 at 11 35 01 AM


spec/mailers/previews/marketplace/order/mailer_preview.rb
```diff
@@ -0,0 +1,19 @@
...
+          notify_emails: "distributor@example.com,vendor@example.com")
+
+        tax_rate = FactoryBot.build(:marketplace_tax_rate, marketplace: marketplace)
+        order = FactoryBot.build(:marketplace_order, :with_products, marketplace: marketplace,
+          delivery_window: 1.hour.from_now, placed_at: 5.minutes.ago,
+          product_count: (1..5).to_a.sample, delivery_address: Faker::Address.full_address,
+          contact_email: Faker::Internet.safe_email, contact_phone_number: Faker::PhoneNumber.cell_phone)
+
+        order.ordered_products.first.product.tax_rates << tax_rate
+        order
```
zspencer, line 15: I feel like this could move to the factory rather than sprouting this weird middle-child class.

spec/support/mailer.rb
```diff
@@ -0,0 +1,129 @@
...
+        def failure_message
+          "Expected mail subject to be '#{expected_subject}', was '#{actual_subject}'"
+        end
+
+        def description
+          "have subject '#{expected_subject}'"
+        end
+      end
+
+      class RenderComponentMatcher
```
zspencer, line 58: I could see myself pulling this up to `Spec::Support::RenderComponentMatcher` and getting rid of the mail-specific bits

anaulin: This definitely seems like it could be very useful outside of Mailers as well.

spec/support/mailer.rb
```diff
@@ -0,0 +1,129 @@
...
+module Spec
+  module Support
+    module Mailer
+      def be_to(email_addresses)
+        BeToMatcher.new(email_addresses)
+      end
+
+      def have_subject(subject)
+        HaveSubjectMatcher.new(subject)
+      end
```
anaulin, line 11: I personally don't see a ton of value in these two matchers -- it seems like testing the contents of `mail.to` and `mail.subject` is simple enough that we don't need to wrap them in helpers. 🤷🏼‍♀️

app/furniture/marketplace/order/mailer.rb
anaulin, line 1: Personally I'd rather have one `Marketplace::Order::Mailer` with a `placed_notification` and a `received_notification` (and whatever we add in the future), than a class hierarchy just so that we can have different `to` lines. Seems overly complicated.

app/furniture/marketplace/order/mailer.rb
```diff
@@ -0,0 +1,15 @@
+class Marketplace
+  class Order
+    class Mailer < ApplicationMailer
+      attr_reader :order
+      helper_method :order
```
anaulin, line 5: I don't understand the purpose of these two lines. Is this just so that the template can refer to `order` instead of `@order`, or am I missing something else?

anaulin (approved):

spec/support/mailer.rb
zspencer: Yea, there's not really a *great* reason for them other than "I like them and think they're neat" which is something I need more of in my life right now.

In particular, I like how they make tests feel, because you can do cool stuff like `it { is_expected.to be_to('person@example.com').exclusively }` on the subject of the example group. I was pretty averse to this kind of custom matchers back when it took their god-awful DSL to define them; but now that they're POROs I'm pretty excited to explore how expressive we can make our tests.

app/furniture/marketplace/order/mailer.rb
zspencer: You're not wrong, but this is definitely something that I want to keep exploring for a little bit. I dislike having multiple emails per `Mailer` but I won't object to them if people put them in.

app/furniture/marketplace/order/mailer.rb
zspencer: Yes, that's exactly it. I am allergic to instance variables in my templates to the point where I will do absurd things to avoid them.

PR #1280, opened 2023-03-30
🐞 `Member`s: You can double destroy it's fine

zspencer:
- https://github.com/zinc-collective/convene/issues/1267

We had written some logic to be like "Don't Revoke!" but that is causing exceptions in cases where we double-tap buttons and such


daltonrpruitt (approved):

rosschapman (approved):
Seems reasonable! 🚀

PR #1281, opened 2023-03-30
WIP: `Marketplace`: Attempt to validate Delivery info

zspencer:

I started down the path by adding to the top level cart, but that's not great because it means I had to do a find_or_initialize_by which is making everything squicky

I think next time I will try the "Extract a Domain Object!" path and switch from the form being a `Cart` form and make it a `DeliveryDetails` form (or something)


zspencer:
Superceded by https://github.com/zinc-collective/convene/pull/1292

PR #1282, opened 2023-03-30
Adds step to machine setup instructions to copy `.env.sample` to `.env`

rosschapman:
I'm assuming this is what you're supposed to do. Should we also include some guidance on how to request missing credentials (S3, database, etc..) from other project maintainers?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,7 +93,8 @@ Then, run `bin/setup` to install Ruby and Node dependencies and set up the
 database.
 
 Once setup completes, configure your local development environment.
-1. Open `.env` and make any changes.
+1. Copy `.env.sample` to `.env`
+1. Update variables in `.env` as needed
 
 Once you have completed configuration; run `bin/run`. You now should be able to open
 http://localhost:3000 and see Convene.


CONTRIBUTING.md
```diff
@@ -93,7 +93,8 @@ Then, run `bin/setup` to install Ruby and Node dependencies and set up the
 database.
 
 Once setup completes, configure your local development environment.
-1. Open `.env` and make any changes.
+1. Copy `.env.sample` to `.env`
```
zspencer, line 96: The `bin/setup` script should copy the file from `.env.sample` to `.env` but it's reasonable to break it into two steps.

zspencer (approved):
I think this is useful!

Re: Get real data - We could add links to docs that show them how to set up their own accounts for the services; but we will only be granting access to Zinc resources (Heroku, Stripe, etc) to employees or maintainers who eventually apply for Member-Ownership.

So, it could be worth adding links to the `.env.sample` file?

CONTRIBUTING.md
rosschapman: Ohh it's possible I didn't run the bin/setup step yet when I noticed this.

zspencer: Ahhh maybe we're doing it in both places for some silly reason.

anaulin: Doesn't `bin/setup` call `bin/setup-rails` at some point? Is the line in both scripts?

anaulin: (Do these need to be separate scripts at all? Does anyone use the standalone script?)

zspencer: I think it's because we renamed the `bin/setup` that comes with `rails` to an independent script so we could write our own script for installing other dependencies... but we could inline them all into a ruby script!

anaulin: To answer my own questions:
* yes, the copying is in both `bin/setup` and `bin/setup-rails`
* one of the scripts in `bash` and the other `ruby`, which makes combining them more work than just copy-pasting one into the other, which to me makes combining the scripts not worth it (but i'm happy for someone who is excited to do it to go for it)

So I'd say, if the spirit moves you, delete lines 41-46 in `bin/setup` to remove the redundant copying (and just assuming that we want to keep as much of the logic as possible in the Ruby script, because it seems nicer than bash). But it doesn't seem a big deal either way.

:shipit:

anaulin:
@rosschapman do you need anything else before either merging or closing this PR? (decision up to you)

zspencer:
Closing this out since @rosschapman  mentioned this doesn't feel necessary to merge.

PR #1292, opened 2023-04-02
✨🥔🥗 `Marketplace`: Validate delivery independently

zspencer:
- [x] Add meaningful tests on the `Cart::DeliveriesController#edit` and `Cart::DeliveriesController#update` endpoints

OK, so this is a biggun, but the exciting part is it's now pretty easy to make assertions about turbo streams!

I didn't do any prettifying as part of this, just putting request tests together and making sure the validations work...

## Before
Screenshot 2023-04-02 at 10 19 45 PM

## After

Screenshot 2023-04-02 at 10 19 53 PM

Screenshot 2023-04-02 at 10 20 47 PM
Screenshot 2023-04-02 at 10 20 42 PM


anaulin (approved):
Thanks for all the additional testing!

It really feels like we want a separate model for Delivery here (ETA: a "real" one, backed by its own table and whatnot).

Issue #1279, opened 2023-03-30
🛠️ `Neighborhood`: Consider temporarily using a session recorder to learn how Space Owners are using Convene


anaulin:
As we start having Real Users, it would be nice to use some "session recording" software to help us understand how Convene is being used, so that we can improve the app to serve them better. Some options that I'm familiar with are:
https://www.fullstory.com/
https://www.hotjar.com/

These tools are a bit creepy, and we probably don't want to really depend on them in the mid-term. But maybe we can use them:
- for a limited time (say, collect data for ~1 month)
- only for space owners

We probably only need use the free tier of these tools.

zspencer:
I agree that it would be helpful for us to gather this as we improve the product, and agree that limitin this to Space Owners within their own Space is probably the "SafeWay"

anaulin:
I added a basic Hotjar installation (no identification or users or any other extra configuration) and it is now deployed. The Hotjar "installation verification" thing is happy.

I tried it on myself, and I see this in the console:
![image](https://user-images.githubusercontent.com/6729309/229385524-2e76bdd0-6473-4763-b561-99479199c866.png)

So we'll see how much information we actually end up receiving -- if our user base has ad blockers and other such tools, we might get very little info.

anaulin:
I have verified that our recording is working.

PR #1293, opened 2023-04-02
🛠️ `Neighborhood`: Add Hotjar session tracking and enable it for all users

anaulin:
Addresses https://github.com/zinc-collective/convene/issues/1279

We had discussed enabling this only for space owners, but thinking more about it, I feel like it would be nice to be able to see guests trying to place a marketplace order as well, so for now I enabled this for all users.


zspencer (approved):
Sounds good. We'll want to take it cautious-like and turn it down/off if it starts being a lot.

anaulin:
Yeah, for sure. We're also on the free plan, so they will never let us record more than 35 sessions a day anyway.

Issue #1278, opened 2023-03-30
🛠️ `Neighborhood`: Capture and report JavaScript errors


anaulin:
Configure Sentry to capture and report JavaScript errors

https://docs.sentry.io/platforms/javascript/

PR #1296, opened 2023-04-02
🛠️ `Neighborhood`: Report browser JavaScritp errors to Sentry

anaulin:
Resolves https://github.com/zinc-collective/convene/issues/1278

I have not bothered setting up the whole webpack sourcemap integration, because it seems like a pain and also because maybe we'll move away from webpack to import-maps or js-bundling soon-ish. This makes the error report less detailed, so maybe it is not worth it adding this in that case, but I think it might still be helpful to at least see that we have an error.

Looks like this in Sentry:
image
https://zinc-collective.sentry.io/issues/4057031271/?project=6052513&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,5 +1,8 @@
 // Entry point for the build script in your package.json
 
+import { initializeSentry } from "./sentry.js";
+initializeSentry();
+
 import * as ActiveStorage from "@rails/activestorage";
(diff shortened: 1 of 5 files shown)


zspencer (approved):
THANK YOU!!! Huge  win!

PR #1297, opened 2023-04-02
🛠️ `Component`: `ButtonComponent` supports `GET` Turbo Streams

zspencer:
- https://turbo.hotwired.dev/handbook/streams#streaming-from-http-responses

It turns out that `Turbo` only does turbo-streams when `PUT`/`POST`/`DELETE` are the request methods.

If we want to make `GET` requests use the `turbo-stream` request type, we need to specify it in the HTML output, i.e. `<a href="..." data-turbo-stream=true>`

I wasn't sure if this should be default behavior or not, so I leaned towards making it default to not 🤷

### DIFF
--- a/app/components/button_component.rb
+++ b/app/components/button_component.rb
@@ -8,7 +8,8 @@ class ButtonComponent < ViewComponent::Base
     method: :put,
     confirm: nil,
     disabled: false,
-    classes: nil
+    classes: nil,
+    turbo_stream: nil
   )
     @label = label
     @title = title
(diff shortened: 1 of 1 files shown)


app/components/button_component.rb
```diff
@@ -8,7 +8,8 @@ def initialize(
     method: :put,
     confirm: nil,
     disabled: false,
-    classes: nil
+    classes: nil,
+    turbo_stream: nil
```
anaulin, line 12: Should this be `false` instead of `nil`? Doesn't make a functional difference right now, but feels more correct.

anaulin (approved):
This default seems right.

app/components/button_component.rb
zspencer: `false` feels better for sure.

PR #1300, opened 2023-04-03
🐛🔨 `Neighborhood`: Update Room policy scope to not show inaccessible rooms to users

anaulin:
Fixes https://github.com/zinc-collective/convene/issues/1295

### Before
image

### After
image


anaulin:
I notice that a whole bunch of our scopes are just `scope.all`, so maybe we want to review them and start being a little bit more rigorous with these now that we are less alpha-ish and more beta-ish.

zspencer (approved):
Love it thank you!

PR #1301, opened 2023-04-03
🌸 `Neighborhood`: Improve clarity of Room publicity and access level options

anaulin:
As a follow-up on https://github.com/zinc-collective/convene/pull/1300, this PR replaces the visibility and access level select drop-downs in the Room edit form with radio groups.

My goal is to make the available options and current setting clearer.

I would like to soon add an "explanation text" to each of these options. I was experimenting with using [Tailwind's `peer/...` and `peer/...checked` classes](https://tailwindcss.com/docs/hover-focus-and-other-states#differentiating-peers) and I couldn't get them working reliably, so I've given up on that for now. But I'll be back.

I also:
* added a database-level default for the `publicity_level`
* converted the access and publicity levels to Rails enums, so that we can get the scopes and predicate methods and validations for free

#### Before
image

#### After
image

### DIFF
--- a/app/views/application/_radio_group.html.erb
+++ b/app/views/application/_radio_group.html.erb
@@ -1,0 +1,11 @@
+<%= form.label attribute %>
+<div class="space-y-4 flex items-center sm:flex sm:items-center sm:space-x-10 sm:space-y-0">
+  <%- options.each do |option| %>
+    <div class="flex items-center mb-0">
+      <%= form.radio_button attribute, option,
+          class: "h-4 w-4 border-gray-300 text-purple-900 focus:ring-purple-600" %>
+      <%= form.label attribute, option, value: option,
+          class: "ml-3 block text-sm leading-6 text-gray-900" %>
+    </div>
+  <%- end %>
+</div>
--- a/db/migrate/20230403013014_add_default_publicity.rb
+++ b/db/migrate/20230403013014_add_default_publicity.rb
@@ -1,0 +1,9 @@
+class AddDefaultPublicity < ActiveRecord::Migration[7.0]
+  def up
+    change_column :rooms, :publicity_level, :string, default: :listed, null: false
+  end
+
+  def down
+    change_column :rooms, :publicity_level, :string, default: nil, null: true
+  end
+end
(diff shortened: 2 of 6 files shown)


db/migrate/20230403013014_add_default_publicity.rb
```diff
@@ -0,0 +1,9 @@
+class AddDefaultPublicity < ActiveRecord::Migration[7.0]
+  def up
+    change_column :rooms, :publicity_level, :string, default: :listed, null: false
```
zspencer, line 3: I'm trying to decide whether `:listed` or `:unlisted` is the better default, but I am landing on 🤷

zspencer (commented):
Thank you! I really like where this is going!

db/migrate/20230403013014_add_default_publicity.rb
anaulin: Yeah, I wasn't sure either, but if feels like a `public` default sort of implies a `listed` default?

anaulin:
Just realized that there is some weirdness in the mobile version:
image

Not sure if that's ok to merge and iterate or not, y'all tell me.

Either way, I'll look into fixing that later this week.

app/views/application/_radio_group.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= form.label attribute %>
+<div class="space-y-4 flex items-center sm:flex sm:items-center sm:space-x-10 sm:space-y-0">
+  <%- options.each do |option| %>
+    <div class="flex items-center mb-0">
+      <%= form.radio_button attribute, option,
+          class: "h-4 w-4 border-gray-300 text-purple-900 focus:ring-purple-600" %>
+      <%= form.label attribute, option, value: option,
+          class: "ml-3 block text-sm leading-6 text-gray-900" %>
+    </div>
+  <%- end %>
```
zspencer, line 10: Nice! I'm glad we have a radio component now!

db/migrate/20230403013014_add_default_publicity.rb
zspencer: That makes sense to me!

zspencer (approved):
v pretty

PR #1255, opened 2023-03-23
✨🥔 `Marketplace`: `Distributor` creates `DeliveryArea`

KellyAH:
This is for https://github.com/zinc-collective/convene/issues/1136

Create `marketplace_delivery_areas table` for storing DeliveryArea `label` and `price` so Shoppers can be charged extra rates based on where catered food is being delivered to.

Create #new, #create, and #index actions for `Marketplace::DeliveryAreasController` 

**Index for DeliveryAreas**
Screenshot 2023-03-29 at 7 18 49 PM

**Create new DeliveryArea Price**
Screenshot 2023-03-29 at 7 19 10 PM

note: `Marketplace::DeliveryAreasController` `#edit` and `#destory` actions in a new PR.

PR #1302, opened 2023-04-03
modify `DeliveryArea` controller tests

KellyAH:
A continuation of https://github.com/zinc-collective/convene/pull/1255

- add `assert_select` to tests to make assertions about what the page will display. 
-  add assertions about how `DeliveryArea` controller #create changes the underlying data

### DIFF
--- a/spec/furniture/marketplace/delivery_areas_controller_request_spec.rb
+++ b/spec/furniture/marketplace/delivery_areas_controller_request_spec.rb
@@ -12,10 +12,18 @@ RSpec.describe Marketplace::DeliveryAreasController, type: :request do
   let(:marketplace) { create(:marketplace) }
   let(:person) { create(:membership, space: space).member }
 
+  describe "#index" do
+    let(:perform_request) { get polymorphic_path(marketplace.location(child: :delivery_areas)) }
+
+    it { is_expected.to be_ok }
+    specify { perform_request && assert_select("a", "Add Delivery Area") }
+  end
+
   describe "#new" do
     let(:perform_request) { get polymorphic_path(marketplace.location(:new, child: :delivery_area)) }
 
     it { is_expected.to be_ok }
+    specify { perform_request && assert_select("form input", 3) && assert_select("#delivery_area_label") && assert_select("#delivery_area_price") }
   end
 
   describe "#create" do
@@ -24,5 +32,6 @@ RSpec.describe Marketplace::DeliveryAreasController, type: :request do
     end
 
     it { is_expected.to redirect_to(marketplace.location(child: :delivery_areas)) }
+    specify { expect { perform_request }.to change { marketplace.delivery_areas.reload.count }.by(1) }
   end
 end


spec/furniture/marketplace/delivery_areas_controller_request_spec.rb
```diff
@@ -24,5 +32,6 @@
     end
 
     it { is_expected.to redirect_to(marketplace.location(child: :delivery_areas)) }
+    specify { expect { perform_request }.to change { marketplace.delivery_areas.reload.count }.by(1) }
```
KellyAH, line 35: I was hoping to assert `label` and `price_cents` in the record. But 
`marketplace.delivery_areas.reload` returns an [ActiveRecord CollectionProxy](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html) which has no method that lets me dig into the record data in the collection.

So below attempts doesn't work and throw the expected `NoMethodError`.
```
specify { expect { perform_request }.to change { marketplace.delivery_areas.reload.label }.from(nil) }
```
When debugging w/ a `pry` breakpoint I noticed `marketplace.delivery_areas.reload.first.label` returns the data I want. But it doesn't work in the test code, throwing the same NoMethodError on the `ActiveRecord::Associations::CollectionProxy`

spec/furniture/marketplace/delivery_areas_controller_request_spec.rb
```diff
@@ -12,10 +12,18 @@
...
+
+    it { is_expected.to be_ok }
+    specify { perform_request && assert_select("a", "Add Delivery Area") }
+  end
+
   describe "#new" do
     let(:perform_request) { get polymorphic_path(marketplace.location(:new, child: :delivery_area)) }
 
     it { is_expected.to be_ok }
+    specify { perform_request && assert_select("form input", 3) && assert_select("#delivery_area_label") && assert_select("#delivery_area_price") }
```
KellyAH, line 26: If this is too long, I can break into separate `specify` blocks.

zspencer: I think this is fine for now. I tend to prefer asserting that the request renders a particular partial or component rather than specifying the view details in the request specs; but until we get good at writing component specs I think this is a fine option!

spec/furniture/marketplace/delivery_areas_controller_request_spec.rb
zspencer: You could use the [`exists?`](http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-exists-3F) predicate method with the `label` and `tax_rate` (i.e. `expect { perform_request }.to change { marketplace.delivery_areas.exists?(label: ..., tax_rate:...) }.from(false).to(true)` or something like that)

zspencer (approved):
I think this is a substantative improvement!

anaulin (approved):
Yay, thank you for improving our tests! 🥗

anaulin:
@KellyAH I notice that this PR is in `draft` still, but also we already did reviews. Let me/us know if you want more review passes if you make more changes.

KellyAH:
merging for now.

spec/furniture/marketplace/delivery_areas_controller_request_spec.rb
KellyAH: If this is going to be a data comparison, I'd need to be able to pass the factory made data to the assertion. Maybe I'll  add `let`s for the `label` and `price_cents` in the `describe "#create"` block so I can use it in the `expect` block.

zspencer: Yea I tend to add a `let(:domain_model_attributes) { attributes_for(:domain_model) }` when I want to use the factory data elsewhere.

PR #1312, opened 2023-04-05
`Authentication`: Downcase email addresses

zspencer:
- https://github.com/zinc-collective/convene/issues/1310
- https://github.com/zinc-collective/convene/issues/1311

So, the `AuthenticationMethodcontact_location` field is case insensitive from a *validation* perspective, but not from an assignment or lookup perspective.

So when someone had signed in with `Bob@example.com`, then tried to sign in as `bob@example.com` it had a sad day.

Similarly, `Invitation#email`

### DIFF
--- a/app/models/authentication_method.rb
+++ b/app/models/authentication_method.rb
@@ -32,6 +32,10 @@ class AuthenticationMethod < ApplicationRecord
     self.person ||= Person.find_or_create_by(email: contact_location)
   end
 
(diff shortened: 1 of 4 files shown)


zspencer:
I also went through and downcased everything in production

anaulin (commented):
Thanks for the quick debugging and fixing! 👍🏼

Issue #892, opened 2022-10-21
🥔⛰️`Neighborhoods`


zspencer:
- https://github.com/zinc-collective/convene/issues/892 are managed by https://github.com/zinc-collective/convene/issues/103 and provide https://github.com/zinc-collective/convene/issues/1154 to https://github.com/zinc-collective/convene/issues/1475

## Use Cases
- [ ] #1450
- [ ] #1449
- [ ] #1503
- [x] https://github.com/zinc-collective/convene/issues/891

### What kind of Neighborhood are there?
While right now, there is only the [Zinc `Neighborhood`](https://convene.zinc.coop) `Neighborhoods` may  specialize in a particular industry, region, or cultural touchpoint. Examples:

- A regional solidarity economy network's Neighborhood provides fundraising, logistics, and community and donor engagement services to their mutual aid and mutual benefit programs.
- A Healthcare Cooperative Network maintains a HIPPA compliant Neighborhood and streamlines appointment booking, medical record storage and transfer, and patient communications.
- A Sci-Fi and Fantasy Writing Club's Neighbohood provides Authors spaces to sell physical and digital editions of their books directly to their fans.
- A network of artisans hosts a Neighborhood that offers a shared online marketplace with individual sites for each of their members.
- A University provides a Neighborhood for students and alumni to keep their social media, email and web-hosting.

If you are hosting a Neighborhood, get in touch so we can add you to our list!

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1317, opened 2023-04-06
🛠️ `Neighborhood`: Expose `sidekiq-web` ui to Operators

zspencer:

We have some sad background jobs in the queue, this will let me purge the heretics.


anaulin:
🤖 is mad

anaulin (approved):
Good idea! 📈

PR #1319, opened 2023-04-06
🧹🔨🥗`Marketplace`: Show `Order#delivery_window` in `Order::EmailReceiptComponent`

zspencer:

So, the `delivery_window` field which can either be a plain string or a particular date or even a date range is a much more sophisticated object.

I took some time to explore what it could look like to use `ActiveModel::Type`s  to create a value object which can take responsibility for representing the data more appropriately.

I *think* the next step would be to back the `Delivery` class with a `marketplace_deliveries` table and move  `Order#contact_email`, etc. into it.

In the meantime, I also added some tests to where the `delivery_window` is being used, so that when / if we make that change we have some guard rails in place.

No UI changes, but this does add `capybara` and configure our `view_component` library for testing.


app/furniture/marketplace/order.rb
```diff
@@ -12,6 +12,7 @@ class Order < Record
     has_many :products, through: :ordered_products, inverse_of: :orders
 
     has_encrypted :delivery_address
+    attribute :delivery_window, ::Marketplace::Delivery::WindowType.new
```
zspencer, line 15: I feel like this should work with `delegate :window, to: :delivery, prefix: true`

app/furniture/marketplace/cart.rb
```diff
@@ -33,15 +34,6 @@ def delivery
...
     end
 
-    def delivery_window
-      return marketplace.delivery_window if marketplace&.delivery_window.present?
-      return nil if super.blank?
-
-      DateTime.parse(super) || 48.hours.from_now
-    rescue Date::Error => _e
-      48.hours.from_now
-    end
```
zspencer, line 43: But hey at least I didn't copy and paste this code everywhere 🤣

zspencer (commented):
This is a weird PR. Apologies in advance.

I noticed that the `Order#delivery_window` was not showing up in the `EmailReceiptComponent` in production, but it *was* in the `mailer_preview`. That was because I was using a `FactoryBot.build(:marketplace_order)` to populate the values, so the delivery window had never needed to be retrieved from the database! And I had done the same in the `Marketplace::OrderPlacedMailer` specs... So at no time was `Order#delivery_window` being called on an `Order` that had actually been pulled out of the database. 

Womp. Womp.

And when I was testing it via the UI, I was using the `Cart` object... Which did have a  [`Cart#delivery_window`](https://github.com/zinc-collective/convene/pull/1319/files#diff-5bf6c5d1fea91de62611487e6cae008c63f51d3cf46da82dd1ead1cf7d35bc9fL36-L43).

I started by just copy and pasting the `delivery_window` method into `Order`, but that offended my oh-so-delicate sensibilities and I revolted. And since I ... reallllllly want.... a `Delivery` class..... I thought "Well, what if I sprout that `Delivery` model?! It sounds like the time!"

And I wanted to relearn the `ActiveRecord::Type` bits since I haven't had much time to do hands-on-keys coding in a bit; and I want to have a better feel for how the pieces work together...

So anyway, that's all the justification I have for why I didn't create a `marketplace_deliveries` table and  store the data that I've been copy-pasting between tables.

I could see wanting to go down that path  sooner rather than later tho. I was really hoping the `Type` stuff would feel like a solid win...

I'm going to add some more precise tests at the `Type` and `Model` level.

app/furniture/marketplace/cart.rb
```diff
@@ -7,7 +7,7 @@ class Cart < Record
 
     default_scope { where(status: :pre_checkout) }
 
-    belongs_to :marketplace, inverse_of: :carts
+    belongs_to :marketplace, inverse_of: :carts, class_name: "Marketplace::Marketplace"
```
zspencer, line 10: I don't know why I had to make this change and it annoys me

app/views/application/_datetime_field.html.erb
```diff
@@ -1,5 +1,6 @@
+<%- disabled = local_assigns.fetch(:disabled?, false)%>
 <div>
   <%= form.label attribute %>
-  <%= form.datetime_field attribute %>
+  <%= form.datetime_field attribute, disabled: disabled %>
```
zspencer, line 4: I feel like this could be merged independently, and maybe wombo-comboed with a `FieldComponent` so more of our `:x_field` partials have standardized inputs.

app/furniture/marketplace/order.rb
zspencer: Note! It does *not* because:

1. `becomes` does not re-initialize the data from the database, so the `attribute` needs to be pulled initially!

app/furniture/marketplace/order/email_receipt_component.html.erb
```diff
@@ -49,7 +49,7 @@
 <tr><th colspan="3">Delivery Details</th></tr>
 <tr>
   <td class="text-right">Delivery Schedule:</td>
-  <td><%= order.delivery_window %></td>
+  <td><%= render(order.delivery.window) %></td>
```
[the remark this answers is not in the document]
zspencer: It is now!

app/furniture/marketplace/order.rb
zspencer: This *will* work when we make a `marketplace_deliveries` table instead of using the `becomes` hack.

app/views/application/_datetime_field.html.erb
anaulin: Wombo-comboed! That's a new one for me.

app/furniture/marketplace/delivery/window.rb
```diff
@@ -0,0 +1,29 @@
...
+          value.eql?(other.value)
+        else
+          eql?(WindowType.new.cast(other))
+        end
+      end
+
+      alias_method :==, :eql?
+
+      def time_like?
+        value.respond_to?(:strftime)
```
anaulin, line 25: Nit: maybe inlining this method would make the `strftime` method clearer -- you are calling strftime if it responds to it -- saving the extra indirection of going "is `time_like?` a built-in thing? oh, no, it is defined down here, let's see... oh, i see".

app/furniture/marketplace/delivery/window.rb
```diff
@@ -0,0 +1,29 @@
+class Marketplace
+  class Delivery
+    class Window < Model
+      attr_accessor :value
+      delegate(:nil?, :blank?, :empty?, to: :value)
+
+      def strftime(*args, **kwargs)
+        return value.strftime(*args, **kwargs) if time_like?
+        value
```
anaulin, line 9: Callers of `strftime` will probably expect a string coming back from this call. Is `value` a string? Or can it be something else that should be cast to a string?

anaulin (approved):
It's not _that_ weird. But perhaps I'm not the best judge of that.

Thanks for all the extra testing goodness!

app/furniture/marketplace/delivery/window.rb
zspencer: Yea, there's an indirection bump but `time_like?` is used in a few other places (in particular figuring out which field to render for the form and how to present the window on the receipt)

app/furniture/marketplace/delivery/window.rb
zspencer: Oo that's a good point... I could just call `to_s`? For now, `value` is either a string or a `DateTime`; but if we start wanting it to hold a date range or other less primitive data structures it will be weird to return them.

app/furniture/marketplace/delivery/window.rb
anaulin: Ah, fair enough.

PR #1323, opened 2023-04-08
🧹🥗🥔 `Marketplace`: Use `DeliveryArea#price` to calculate `Cart#price_total`

zspencer:

Now the pricing reflects the delivery area when the Marketplace and Cart have them! Fancy!

I feel like this is a reasonable amount of test coverage for this change, but if we want to add `request` or `component` specs or something I could be persuaded...

### DIFF
--- a/app/furniture/marketplace/delivery.rb
+++ b/app/furniture/marketplace/delivery.rb
@@ -1,12 +1,20 @@
 class Marketplace
   class Delivery < Record
     self.table_name = "marketplace_orders"
+
     belongs_to :marketplace
     belongs_to :shopper
+    belongs_to :delivery_area
+    validates :delivery_area_id, presence: true # rubocop:disable Rails/RedundantPresenceValidationOnBelongsTo
+
     attribute :delivery_window, WindowType.new
 
     def window
       delivery_window
+    end
+
+    def fee
+      delivery_area&.price.presence || marketplace&.delivery_fee.presence
     end
   end
 end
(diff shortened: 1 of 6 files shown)


app/furniture/marketplace/delivery.rb
```diff
@@ -1,12 +1,23 @@
 class Marketplace
   class Delivery < Record
     self.table_name = "marketplace_orders"
+
     belongs_to :marketplace
     belongs_to :shopper
+    belongs_to :delivery_area
+    validates :delivery_area_id, presence: true # rubocop:disable Rails/RedundantPresenceValidationOnBelongsTo
```
anaulin, line 8: How come we need to validate this separately? Is it because it is somehow not a "real" association, or some other not-quite-standardness? (Asking for my own enlightenment.)

spec/furniture/marketplace/delivery_spec.rb
```diff
@@ -1,7 +1,10 @@
...
 
 RSpec.describe Marketplace::Delivery, type: :model do
-  subject(:delivery) { described_class.new }
+  subject(:delivery) { build(:marketplace_delivery, marketplace: marketplace, delivery_area: delivery_area) }
+
+  let(:marketplace) { build(:marketplace) }
+  let(:delivery_area) { nil }
 
   it { is_expected.to belong_to(:marketplace) }
   it { is_expected.to belong_to(:shopper) }
```
anaulin, line 10: Might be worth adding expectations for the new `belongs_to`, and the new validation you added.

anaulin (approved):
Very nice, thank you!

I think if we want to add more request/controller tests, that can be done in a separate PR, but this seems sufficient to me for now. 🥦

app/furniture/marketplace/delivery.rb
zspencer: I was annoyed and confused by this to, but because we set the `delivery_area_id` in the form if I don't have it then there's no feedback about why it's invalid.

anaulin: Ugh. Fair enough.

PR #1329, opened 2023-04-09
🌸✨ `Marketplace`: Add shared Marketplace Management component

anaulin:
My main goal here is to provide a shared element that links to all the configuration and management bits of a marketplace. This is far from perfect or "done", but it gives us something that we can hopefully iterate on alongside adding more areas to the Marketplace app.

Things I plan to do in a follow-up PR:
* add "active" menu states to the side menu, so that you can see where you are
* do something about the "Add Stripe API Key" and "Connect Stripe" buttons in the "Configuration" area
* figure out if I like or don't like the "Marketplace <room name>" header, and what might be a better alternative (I do feel like we need to be better at telling the user where they are, I don't think the breadcrumbs are enough, especially since they don't show "entrance hall" if that's where you are)
* figure out how we want to treat the Order History -- I haven't enclosed it in the Marketplace Management component yet, since I assume we want it to work both for the marketplace managers and for a shopper

#### Marketplace
image

#### Marketplace management, showing general configuration area
image

#### Marketplace management on mobile, showing product management area
image


app/components/button_component.rb
```diff
@@ -1,38 +1,25 @@
 # frozen_string_literal: true
 
-class ButtonComponent < ViewComponent::Base
-  attr_accessor :spacing, :shape, :typography, :color, :animation
```
zspencer, line 4: Was this an intentional removal? I put them there so I could do some follow-on clean-up of the UI for different buttons without having to pass in every single class every time.

app/furniture/marketplace/management_component.html.erb
```diff
@@ -0,0 +1,67 @@
...
+            <% menu_item.with_title(t('marketplace.marketplace.configuration_menu')) %>
+          <% end %>
+        <% end %>
+
+        <%- if policy(marketplace).edit? %>
+          <%= render Marketplace::ManagementMenuItemComponent.new(
+            location: marketplace.location(child: :products)) do |menu_item| %>
+            <% menu_item.with_icon do %>
+              <path stroke-linecap="round" stroke-linejoin="round" d="M9.568 3H5.25A2.25 2.25 0 003 5.25v4.318c0 .597.237 1.17.659 1.591l9.581 9.581c.699.699 1.78.872 2.607.33a18.095 18.095 0 005.223-5.223c.542-.827.369-1.908-.33-2.607L11.16 3.66A2.25 2.25 0 009.568 3z" />
+              <path stroke-linecap="round" stroke-linejoin="round" d="M6 6h.008v.008H6V6z" />
```
zspencer, line 26: I would appreciate it if these were named somehow, rather than a blob of svg.

app/furniture/marketplace/management_component.html.erb
```diff
@@ -0,0 +1,67 @@
+<div class="flex flex-wrap items-baseline">
+  <h2 class="ml-2 mt-1">
+    <%= t('marketplace.marketplace.manage_header') %>
```
zspencer, line 3: Thoughts on using a relative path? That way if we delete the component we know what locales keys we can get rid of too?

app/furniture/marketplace/marketplaces/edit.html.erb
```diff
@@ -1,20 +1,8 @@
...
-      <%= link_to t('marketplace.products.index.link_to'), marketplace.location(child: :products) %>
-    <%- end %>
-
-    <%- if policy(marketplace.delivery_areas.new).create? %>
-      <%= link_to t('marketplace.delivery_areas.index.link_to'), marketplace.location(child: :delivery_areas) %>
-    <%- end %>
-
-    <%- if policy(marketplace.tax_rates.new).create? %>
-      <%= link_to(t("marketplace.tax_rates.index.link_to"), marketplace.location(child: :tax_rates)) %>
-    <%- end %>
```
zspencer, line 16: Wooo!!

app/furniture/marketplace/orders/_order.html.erb
```diff
@@ -21,7 +21,7 @@
     <%- end %>
   <%- end %>
   <hr class="col-span-6 mb-4" />
-  <div class="col-span-4 text-sm">Delivery to <%= order.delivery_area.label %></div>
+  <div class="col-span-4 text-sm">Delivery to <%= order.delivery_area&.label %></div>
```
zspencer, line 24: Nice

app/components/buttons/secondary_component.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Buttons
+  class SecondaryComponent < ButtonComponent
```
zspencer, line 4: What do you think about using custom constructors rather than custom classes? I.e. `ButtonComponent.new_secondary_button` or `Component#remove_button`

app/furniture/marketplace/cart/deliveries/_delivery.html.erb
```diff
@@ -4,5 +4,10 @@
     <%= render(delivery.delivery_area) if delivery.delivery_area.present? %>
   </span>
 
-  <%= render Marketplace::Cart::Delivery::EditButtonComponent.new(delivery) %>
+  <%= render Buttons::SecondaryComponent.new(
```
zspencer, line 7: Ya! I was hoping to ditch the overly-specific `EditButtonComponent`

app/furniture/marketplace/delivery_areas/index.html.erb
```diff
@@ -1,18 +1,20 @@
...
-  </main>
+    </main>
 
-  <footer class="text-center">
-    <%= link_to(t("marketplace.delivery_areas.new"), marketplace.location(:new, child: :delivery_area)) %>
-  </footer>
-</section>
\ No newline at end of file
+    <footer class="text-center">
+      <%= link_to(t("marketplace.delivery_areas.new"), marketplace.location(:new, child: :delivery_area)) %>
```
zspencer, line 17: We've been using the convention `controller.action.link_to` for our lookups, i.e. "marketplace.delivery_areas.new.link_to`

This way they live next to the `marketplace.delivery_areas.new.success` and `marketplace.delivery_areas.new.failure` items

app/furniture/marketplace/management_component.html.erb
```diff
@@ -0,0 +1,67 @@
...
+              location: marketplace.location(child: :delivery_areas)) do |menu_item| %>
+            <% menu_item.with_icon do %>
+              <path stroke-linecap="round" stroke-linejoin="round" d="M9 6.75V15m6-6v8.25m.503 3.498l4.875-2.437c.381-.19.622-.58.622-1.006V4.82c0-.836-.88-1.38-1.628-1.006l-3.869 1.934c-.317.159-.69.159-1.006 0L9.503 3.252a1.125 1.125 0 00-1.006 0L3.622 5.689C3.24 5.88 3 6.27 3 6.695V19.18c0 .836.88 1.38 1.628 1.006l3.869-1.934c.317-.159.69-.159 1.006 0l4.994 2.497c.317.158.69.158 1.006 0z" />
+            <% end%>
+            <% menu_item.with_title(t('marketplace.delivery_areas.index.link_to')) %>
+          <% end %>
+        <% end %>
+
+        <%- if policy(marketplace.tax_rates.new).create? %>
+          <%= render Marketplace::ManagementMenuItemComponent.new(
```
zspencer, line 43: Sometimes I like to pull these into little component helpers. `Marketplace::EditComponent.menu_item(:tax_rates) do |menu_item| ...`

app/furniture/marketplace/management_component.rb
```diff
@@ -0,0 +1,11 @@
+class Marketplace
+  class ManagementComponent < ApplicationComponent
```
zspencer, line 2: My brain keeps wanting this to be the `Marketplace::EditComponent`, but it's not *really* the `Marketplace#edit` action and is a more general purpose layout... I don't think this comment is helpful but it is what I noticed when I was reading.

zspencer (approved):
Yea! I like it! Do you want to merge or can I?

app/components/button_component.rb
anaulin: Yep, this was intentional. I don't think having these CSS classes separated is necessary, or even makes a great deal of sense (cue debate about putting the text color class under "color" vs "typography").

I think if we want to have different variants of buttons, we should make reusable variants of the button, like I did in this PR with the Secondary button. We don't need accessors for that.

If and gene we have other use cases that require overriding something more granular, we can work on a solution then.

app/furniture/marketplace/management_component.html.erb
anaulin: If and when they are reused, they can become named components, like Gear. I am not sure naming them for a one-time use makes sense.

app/furniture/marketplace/delivery_areas/index.html.erb
anaulin: I'll address in my follow-up.

app/furniture/marketplace/management_component.html.erb
anaulin: I like that idea a lot. I'll come up with something in a follow up.

app/components/buttons/secondary_component.rb
anaulin: I have to think about it.

In order to be used as slots it is more convenient for them to be class names, but they could be used in a slot with a custom factory method as well, called in a block.

🤔

app/components/button_component.rb
zspencer: That makes sense, and we can always use custom little factory methods on the componet if a full component would be "too much"

app/furniture/marketplace/management_component.html.erb
zspencer: I like names as a way of revealing intent, since I can't really visualize what this is supposed to look like. Even a little `<!-- price tag icon -->` or something would be appreciated.

app/furniture/marketplace/delivery_areas/index.html.erb
anaulin: FWIW, this was pre-existing, I just wrapped this whole thing in the new component.

Issue #1326, opened 2023-04-08
🥔✨ `Marketplace`: Buying 🥡 `Products`


zspencer:
-  https://github.com/zinc-collective/convene/issues/831

## Use Cases

- When https://github.com/zinc-collective/convene/issues/1326 a #2043 moves through four stages:
  1. Beginning their #2215 including finding a  #2044 and setting their https://github.com/zinc-collective/convene/issues/1136
  2. Assembling their #2215 by browsing the #2153 and adds  #2169(s) to their https://github.com/zinc-collective/convene/issues/2155. 
  3. Checking Out, which consists of reviewing their #2155, confirming their https://github.com/zinc-collective/convene/issues/1136 and https://github.com/zinc-collective/convene/issues/1763, and making a Payment to finalize their #2215
  4. Picking up or Receiving the https://github.com/zinc-collective/convene/issues/1325 for their   #2215

- [x] 🥔 Preparing to Buy 
  - [ ] #2378
  - [x] #2153
  - [ ] #2155
  - [x] 🥔 Adding `Product`s to `Cart`
  - [x] 🥔 Adjusting `Product` quantities in `Cart`
  - [x] 🥔 Removing `Product`s from `Cart`
  - [x] 🥔 Setting `Delivery` details
- [x] 🥔 `Marketplace`:  🕺`Shopper` Checks Out 🛒`Cart`
  - [x] https://github.com/zinc-collective/convene/issues/1672
- [x] 🥔 Paying for `Order`
  - [x] 🥔 Pay using Credit or Debit Card via Stripe 
  - [x] 🥔 Buyer receives Order Placed email with permalink to `Order`
  - [x] https://github.com/zinc-collective/convene/issues/1238
  - [ ] Pay using Cash.app https://github.com/zinc-collective/convene/issues/832
  - [ ] Pay using Venmo https://github.com/zinc-collective/convene/issues/832
  - [ ] Pay using Zelle https://github.com/zinc-collective/convene/issues/832
  
## Domain Model
  - https://github.com/zinc-collective/convene/issues/2169
  - https://github.com/zinc-collective/convene/issues/2153
  - https://github.com/zinc-collective/convene/issues/2155
  - https://github.com/zinc-collective/convene/issues/2215
  
## Bugs
  - [x] https://github.com/zinc-collective/convene/issues/1657
  - [x] https://github.com/zinc-collective/convene/issues/1983
  - [x] https://github.com/zinc-collective/convene/issues/2002
[...]

Issue #1325, opened 2023-04-08
🥔✨ `Marketplace`: `Delivery`


zspencer:
- https://github.com/zinc-collective/convene/issues/831  offers  https://github.com/zinc-collective/convene/issues/1325 of `Order`s via `Distributor`s

## Use Cases
- [x] #1510
- [ ] #1763
- [x] https://github.com/zinc-collective/convene/issues/2641

## Domain Model
- [x] https://github.com/zinc-collective/convene/issues/1185
- [x] https://github.com/zinc-collective/convene/issues/1136

## Design
- [ ] #1509

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

Issue #1324, opened 2023-04-08
🥔✨ `Marketplace`: Selling 🥡`Products`


zspencer:
- https://github.com/zinc-collective/convene/issues/831
- https://github.com/zinc-collective/convene/issues/1137
- https://github.com/zinc-collective/convene/issues/1331

## Use Cases
- [x] 🥔✨ Selling `Products` for local `Delivery`
- [ ] #1574
- [x] #2168
- [x] 🥔 Updating `Product`s
- [x] #1576
- [x] https://github.com/zinc-collective/convene/issues/1469
- [ ] Pricing Products
  - [x] 🥔 with a Price per Unit
  - [ ] #1575
- [x] #2023
- [ ] https://github.com/zinc-collective/convene/issues/1446
- [x] 🥔 `Order` Placed Notifications
  - [x] 🥔 via email
  - [ ] via SMS
- [x] #1428
- [ ] #1955

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #1332, opened 2023-04-09
🥗🧹🥔✨ `Marketplace`: Show `DeliveryArea` in `Order::EmailReceiptComponent`

zspencer:

I didn't use the `marketplace_order, :full` factory outside of the `Order::PlacedMailerPreview` and `Order::ReceivedMailerPreview`; and apparently had flubbed the implementation; so that was not working...

So I fixed it and used it in the `Order::EmailReceiptComponent` spec to test the `Order#delivery_area`

I also noticed we didn't wire in associations between the `Marketplace::DeliveryArea` into the `Marketplace::Cart` and `Marketplace::Order` so I added some specs for those.


anaulin (approved):
🧾 🚚 🥦

PR #1333, opened 2023-04-09
🌸 `Components`: Reduce `Card` padding on small screens

zspencer:

The `p-6` felt like just a bit too much on mobile screens; so I made it `p-4` and then when we get to tablet size it's `p-6`.

### After 
Screenshot 2023-04-09 at 12 49 53 PM
Screenshot 2023-04-09 at 12 49 46 PM


anaulin (approved):
Nice!

Issue #1331, opened 2023-04-09
🥔✨`Marketplace`: `Order` `Notification`s and `Timeline`


zspencer:
- https://github.com/zinc-collective/convene/issues/831

### Use Cases
- [x] #1661
- [x] `Distributor` and `Vendor` see everyone's `Order`s
- [x] #1654
- [x] `Shopper` receives `Order::PlacedNotification`
- [x] #1511
- [x] ✨ `Order` shows recorded `Events`

### Bugs
- [x] https://github.com/zinc-collective/convene/issues/1660
- [x] https://github.com/zinc-collective/convene/issues/1656

rosschapman:
Looks done enough to me.

PR #1334, opened 2023-04-09
🌸 `Marketplace`: Tidy up `Order#show` and `Order#index`

zspencer:

This will look better when
https://github.com/zinc-collective/convene/pull/1333 gets merged,

I also thought about moving it into an `Order::OrderComponent` so it could be more easily tested...


anaulin (approved):
🌷

PR #1335, opened 2023-04-09
🛠️ `Neighborhood`: Add error field workaround to display association errors correctly on forms

anaulin:
This PR extends our `errors` partial to show association validation errors on the association foreign key field. This removes the need to have a duplicate presence validation on non-optional `belongs_to` relationships.

A more full solution is possible by patching ActionView::Helpers::ActiveModelInstanceTag, but it also would be more complex and doesn't seem worth it at the moment. If/when https://github.com/rails/rails/issues/28772 is resolved, we won't need this workaround anymore.

For an example of the more complete patch, see: https://web.archive.org/web/20190605211237/http://iada.nl/blog/article/rails-tip-display-association-validation-errors-fields

After this change, form errors look like so:
image

### DIFF
--- a/app/views/application/_error.html.erb
+++ b/app/views/application/_error.html.erb
@@ -1,3 +1,12 @@
-<%- if model.errors.messages_for(attribute).present? %>
-  <p class="field_with_errors"><%= model.errors.messages_for(attribute).join(', ') %></p>
+<% error_messages = model.errors.messages_for(attribute) %>
+<% %w(_id _ids).each do |suffix| %>
+  <%# Workaround for the fact that Rails doesn't add errors from the association foreign key to
+      the association itself. Yet. See: https://github.com/rails/rails/issues/28772 %>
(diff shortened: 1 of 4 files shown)


app/views/application/_error.html.erb
```diff
@@ -1,3 +1,12 @@
-<%- if model.errors.messages_for(attribute).present? %>
-  <p class="field_with_errors"><%= model.errors.messages_for(attribute).join(', ') %></p>
-<%- end %>
\ No newline at end of file
+<% error_messages = model.errors.messages_for(attribute) %>
+<% %w(_id _ids).each do |suffix| %>
+  <%# Workaround for the fact that Rails doesn't add errors from the association foreign key to
+      the association itself. Yet. See: https://github.com/rails/rails/issues/28772 %>
```
zspencer, line 4: 🙃✨

zspencer (approved):
Nice! Thank you!

PR #1339, opened 2023-04-10
🛠️ `Neighborhood`: Enforce `rubocop-capybara` and quiet the notice

zspencer:

Surprised there weren't any errors, but I'll take it!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,6 +4,7 @@ require:
   - standard
   - rubocop-rails
   - rubocop-rspec
+  - rubocop-capybara
 
 inherit_gem:
   standard: config/base.yml


.rubocop.yml
```diff
@@ -4,6 +4,7 @@ require:
   - standard
   - rubocop-rails
   - rubocop-rspec
+  - rubocop-capybara
```
KellyAH, line 7: what notice is being quieted w/ this change?

zspencer:
```
The following RuboCop extension libraries are installed but not loaded in config:
  * rubocop-capybara

You can opt out of this message by adding the following to your config (see https://docs.rubocop.org/rubocop/extensions.html#extension-suggestions for more options):
  AllCops:
    SuggestExtensions: false
```

anaulin:
Thanks for this, it was annoying the heck out of me. 😅

PR #1338, opened 2023-04-10
🌸🥗✨ `Marketplace`: Tidy up managing `TaxRate`s

zspencer:
- https://github.com/zinc-collective/convene/issues/1137
- https://github.com/zinc-collective/convene/issues/1187

OK this is a bit bigger than it probably should be but I came in like a wreeeckiin' ball and:

- 🛠️ `Components`: `ApplicationComponent` exposes `policy` and `current_person`
- 🛠️ `Components`: Component Specs have `polymorphic_path`
- 🥗 `Marketplace`: Tested `TaxRate`, `TaxRateComponent` and `TaxRatesController`
- 🌸 `Marketplace`: `TaxRate#edit`, `TaxRate#update`, and `TaxRate#destroy` via TurboStreams
- 🌸 `Marketplace`: `TaxRate#index` and `TaxRate#show` are prettier

### After 
Screenshot 2023-04-09 at 6 15 57 PM
Screenshot 2023-04-09 at 6 15 53 PM

### Before
Screenshot 2023-04-09 at 6 38 32 PM
Screenshot 2023-04-09 at 6 38 28 PM
Screenshot 2023-04-09 at 6 38 23 PM
Screenshot 2023-04-09 at 6 38 17 PM

PR #1340, opened 2023-04-10
🌸`Marketplace`: Bring `Product` and `TaxRate` UI closer

zspencer:
- https://github.com/zinc-collective/convene/pull/1338

I think I like the consistency of using the `icons.edit` / `edit.link_to` and `icons.destroy` / destroy.link_to` in the footer of the card, so there are words with the icons by default.

### After
Screenshot 2023-04-09 at 7 39 05 PM
Screenshot 2023-04-09 at 7 39 00 PM
Screenshot 2023-04-09 at 7 38 52 PM

### Before 
Screenshot 2023-04-09 at 7 48 23 PM
Screenshot 2023-04-09 at 7 48 17 PM
Screenshot 2023-04-09 at 7 48 09 PM

### DIFF
--- a/app/furniture/marketplace/locales/en.yml
+++ b/app/furniture/marketplace/locales/en.yml
@@ -55,9 +55,10 @@ en:
       create:
         success: "Added Product '%{name}'"
       edit:
-        link_to: "Edit Product '%{name}'"
+        link_to: "Change"
+        link_to: "Change Product '%{name}'"
       destroy:
-        link_to: "Remove Product '%{name}"
+        link_to: "Remove"
     tax_rates:
       index:
         link_to: "Tax Rates"
--- a/app/furniture/marketplace/locales/en.yml
+++ b/app/furniture/marketplace/locales/en.yml
@@ -57,7 +57,7 @@ en:
       edit:
         link_to: "Edit Product '%{name}'"
       destroy:
-        link_to: "Remove"
+        link_to: "Remove Product '%{name}'"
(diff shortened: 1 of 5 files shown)


app/furniture/marketplace/locales/en.yml
```diff
@@ -55,9 +55,10 @@ en:
       create:
         success: "Added Product '%{name}'"
       edit:
-        link_to: "Edit Product '%{name}'"
+        link_to: "Change"
+        link_to: "Change Product '%{name}'"
```
KellyAH, line 59: IMHO `Edit` is clearer than `change` especially for modifying the details of a product. `Change` IMHO makes it sound like the product is changing to a different product.

KellyAH: But I'm also okay w/ this terminology change (pun intended 😉 ) too .... so 🤷‍♀️

anaulin: Yeah, I think I like "Edit" a bit better as well. Feels more standard, less surprising/confusing.

anaulin (approved):

app/furniture/marketplace/locales/en.yml
zspencer: I am down for keeping it as `Edit` my main goal here was consolidating it into a top-level `edit.link_to` which always just says "Edit" which we can use when we don't want to define a custom `<resource>.edit.link_to`

PR #1342, opened 2023-04-10
🛠️🥗 `Components`: expose `policy` method

zspencer:
- https://github.com/zinc-collective/convene/pull/1338

### DIFF
--- a/app/components/application_component.rb
+++ b/app/components/application_component.rb
@@ -23,4 +23,13 @@ class ApplicationComponent < ViewComponent::Base
     return if args.blank?
     super
   end
+
+  def policy(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


anaulin (commented):
This is ok, but I am not sure it is necessary. In my pending PR https://github.com/zinc-collective/convene/pull/1329 I have accomplished the same thing by adding
```
delegate :current_person, :policy, to: :helpers
```

to `ApplicationComponent`, and it seems to work fine.

zspencer:
That is a much tidier way to do this, but it means we can't set the `current_user` in specs, because `ViewComponent::Base#helpers` is not accessible until after `render`

anaulin (approved):
Hmm. Ok then. :-)

PR #1352, opened 2023-04-13
🧹`Marketplace`: Remove unused `TaxRate` partial!

zspencer:

This also makes our `have_rendered_turbo_stream` matcher play more nicely with `remove` actions.
Co-authored-by: Kelly Hong <KellyAH@users.noreply.github.com

### DIFF
--- a/spec/support/turbo.rb
+++ b/spec/support/turbo.rb
@@ -25,6 +25,7 @@ module Spec
       def matches?(response)
         @response = response
         assert_turbo_stream(action: action, target: target) do
+          rendering[:allow_inferred_rendering] = false if action == :remove
           assert_dom_equal(turbo_stream.action(action, target, content, **rendering), response.body)
         end
       rescue Minitest::Assertion => e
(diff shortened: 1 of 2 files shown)


spec/support/turbo.rb
```diff
@@ -25,6 +25,7 @@ def initialize(action, target, content = nil, turbo_stream:, **rendering, &callb
       def matches?(response)
         @response = response
         assert_turbo_stream(action: action, target: target) do
+          rendering[:allow_inferred_rendering] = false if action == :remove
```
anaulin, line 28: This change seems unrelated to the description and the rest of this PR. I assume you need this change for something else somewhere else and just decided to pull it into this PR, and this is not a mistake?

anaulin (approved):

spec/support/turbo.rb
zspencer: Yea, the second commit has more details, but basically `assert_turbo_stream` presumes you want to `:allow_inferred_rendering`, so when I removed this partial the `TaxRatesController#destroy` tests started failing because it wanted to render. When you look at the underlying code for the `TurboStreamBuilder`, however, the `remove` action when called directly sets `allow_inferred_rendering` to false.

I considered calling `send(action, ...`) so that we would be using the `TurboStreamBuilder`s defaults, but ran into some weirdness that I didn't want to take time to pin down.

anaulin: My point was not about the implementation, but about the fact that this change does not seem to belong in this PR.

zspencer: Ah, well the reason it is in this PR was because the test was failing when I removed the partial.

Issue #1177, opened 2023-03-03
🛠️ `Neighborhood`: Install `strong_migrations` to catch risky operations


KellyAH:
installing strong_migrations were mentioned in https://github.com/zinc-collective/convene/pull/1169#discussion_r1124954327

[strong_migrations](https://github.com/ankane/strong_migrations) will help us catch risky operations and safely ignore and drop deprecated database columns

rosschapman:
➕ for this because I hit a bad error when setting up my environment. There are two migrations that reference a `UtilityHookup` model that no longer exists. 

`db/migrate/20210710224638_encrypt_utility_hookup_configuration.rb`
`db/migrate/20230119012952_remove_jitsi_plaid_utilities.rb`

Although, after downgrading from Postgres 14 to Postgres 12 I could _somehow_ run `db:migrate` successfully -- don't ask me why (yet! 😅).

It sounds like this gem would prevent this problem. Relatedly, what's the best way to get rid of bad/defunct migrations?

rosschapman:
@zspencer @KellyAH perhaps this would be a good first task for me while I'm getting situated with the system, and could work on during tomorrow's huddle?

KellyAH:
> get rid of bad/defunct migrations

Maybe https://guides.rubyonrails.org/active_record_migrations.html#schema-dumping-and-you can help.

anaulin:
@rosschapman if you feel moved to tackle this, go for it! 💪🏼 Generally speaking, no need to ask for permission, you can just leave a note here saying that you plan to work on it, to avoid effort duplication, and take it on.

zspencer:
Wooo !! THank you @rosschapman !!!

PR #1354, opened 2023-04-13
🛠 `Neighborhood`: Adds strong_migrations gem 

rosschapman:
- https://github.com/zinc-collective/convene/issues/1177

This change is part of an effort to add guard rails for devs writing migrations and catch *unsafe* and *illegal* operations on the database.

NOTE: There are other types of "bad" migrations that will not be caught by this gem. For example, referencing an Active Record model to execute DML. For those cases we should consider other static analysis methods like rubocop.


zspencer (approved):
Thank you!!!

PR #1357, opened 2023-04-13
🌸✨ `Marketplace`: Require confirmation before deleting `Product`s and `TaxRate`s

KellyAH:

# problem
I noticed when deleting a `Marketplace::Product` or `Marketplace::TaxRate`, there was no "are you sure?" confirmation modal. This could lead to customer's accidentally deleting a product or tax rate.

# Changes
- add deletion confirmation modal to `Marketplace::Product`.
- add deletion confirmation modal to `Marketplace::TaxRate`.
- apply new button component to `Marketplace::TaxRate` edit and delete buttons.

**before**
Screenshot 2023-04-12 at 9 57 44 PM

**after**
Screenshot 2023-04-12 at 9 49 52 PM

Screenshot 2023-04-12 at 9 49 40 PM

Screenshot 2023-04-12 at 9 51 11 PM

### DIFF
--- a/app/furniture/marketplace/products/_product.html.erb
+++ b/app/furniture/marketplace/products/_product.html.erb
@@ -19,6 +19,6 @@
 
   <footer class="mt-3 flex justify-between">
     <%= render Buttons::SecondaryComponent.new(label: "#{t('icons.edit')} #{t('edit.link_to')}",  title: t('marketplace.products.edit.link_to'), href: product.location(:edit), method: :get) %>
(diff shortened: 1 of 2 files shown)


zspencer (approved):
Love it! 

I think we probably want to consider erring on the side of not allowing people to delete things that should not be deleted (`Product`s, for example, may want to be flagged as "archived" rather than actually deleted.)

I also like to review the Norman Groups guidelines sometimes since I am always rusty about what kind of places deserve confirmation dialogs: https://www.nngroup.com/articles/confirmation-dialog/

PR #1370, opened 2023-04-16
🧹`Components`: add Svg `icon` variants and Button `scheme`s

anaulin:
Follow-up on https://github.com/zinc-collective/convene/pull/1329, removing some of the unnecessary complexity in components: instead of using inheritance for reuse (ewww), now Svg has an `icon` parameter, and Button has a `scheme` parameter. This also adds a "secondary" button.

The only visible change with this PR is that the "Add Stripe API Key" button now looks like a primary button:
image


spec/components/previews/button_component_preview.rb
```diff
@@ -1,5 +1,5 @@
 class ButtonComponentPreview < ViewComponent::Preview
   def test
-    render(ButtonComponent.new(label: "Awooo!", title: "gaaa!", href: "#", classes: ""))
+    render(ButtonComponent.new(label: "Awooo!", title: "gaaa!", href: "#", scheme: :secondary))
   end
```
zspencer, line 4: Maybe we could have two preview methods: `secondary` and `primary`? Or we could expose `scheme` as an argument in the params so Lookbook lets us override it?

app/components/button_component.rb
```diff
@@ -18,12 +24,17 @@ def initialize(
...
+    @scheme = scheme
 
     super(data: data, **kwargs)
   end
 
   private
 
+  def classes
+    (base_classes + send(SCHEME_MAPPINGS[@scheme])).join(" ")
+  end
```
zspencer, line 36: Does this mean we can't overload `@classes` anymore to fine-tune margin or padding for the particular usage?

zspencer (approved):
I think this is a good step, but I worry about losing the ability to overload the base classes for things like typography or spacing changes. 

I wonder what it would be like if what we currently define as `base_classes` stays as the default value for `classes:`, so that when you pick `primary` or `secondary` it's just `@classes + variant_classes`

app/components/button_component.rb
anaulin: Not if you're using a `scheme`, no. My thinking is that eventually we'll add e.g. `size` variants, which will set different paddings, font size, etc, and be able to combine a scheme with a size.

anaulin: More generally, I'd prefer we start with relatively rigid components with a limited number of variants. It feels like we should not need to be fine-tuning things like a button's padding on a one-off basis.

spec/components/previews/button_component_preview.rb
anaulin: That sounds nice in theory. In practice, our Lookbook seems to not behave the way the documents state, and half of the previews we have don't even load, so I don't have a lot of motivation to fiddle with this.

I did try to add one method per `scheme` option, but Lookbook only shows me the results of the last method that I define. 🤷🏼‍♀️

anaulin: I got it working, somehow. I think Lookbook reload is doing something funky? And also dynamically declaring methods doesn't seem to work. I hate everything.

PR #1371, opened 2023-04-16
🥗🥔✨ `Marketplace`: add `order_by` and `delivery_window` to `DeliveryArea`

zspencer:

In order to support different timing for different `DeliveryArea`s, we've given them `delivery_window` and `order_by` fields.

These fields will be used to communicate on the `Cart#show` and `Order#show` screens what the timing expectations are for particular orders.

Example:

- Catering orders for Lions Dance Cafe can be placed any time and delivered anytime, but
- Dinner orders must be placed the same day by 3pm

### After
Screenshot 2023-04-16 at 10 56 24 AM
Screenshot 2023-04-16 at 10 56 15 AM
Screenshot 2023-04-16 at 10 56 09 AM

### DIFF
--- a/app/furniture/marketplace/delivery/windows/_window.html.erb
+++ b/app/furniture/marketplace/delivery/windows/_window.html.erb
@@ -1,5 +1,8 @@
+delivery
 <%- if window.time_like? %>
-  <%= l(window.value, format: :day_month_date_hour_minute) %>
+  at <%= l(window.value, format: :day_month_date_hour_minute) %>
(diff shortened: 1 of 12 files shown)


anaulin (approved):
Is the plan to remove the `delivery_window` from `Marketplace` and replace it with this?

zspencer:
> Is the plan to remove the delivery_window from Marketplace and replace it with this?

I hope so! I think it has a lot more value as a unit for sure.

Issue #1364, opened 2023-04-15
🥔✨ `Space`:  `Agreement`s


zspencer:
- https://github.com/zinc-collective/convene/issues/1154

`Spaces` may have a Cookies Agreement, Privacy Agreement, Terms of Service, Content Policy, etc. 

We don't currently offer a set of Terms of Service or a broader library, but we can at least let people put theirs in.
 

## Use Cases

- [x] Creating an Agreement
- [x] Finding Agreements
- [x] Reading an Agreement
- [ ] Amending an Agreement
- [ ] Accepting a Required Agreement
  - [ ] Before Viewing the Space
  - [ ] Before Writing To the Space
  - [ ] Before Joining the Space

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #1374, opened 2023-04-16
🥗🥔✨ `Agreements`: can be `created`!

zspencer:
- https://github.com/zinc-collective/convene/issues/1364

This is only for the classic HTML Request <=> Response Cycle. I realized I probably want to do the `#index` Routes before `#create`, or at least get started on the bits to expose adding an `Agreement` in the `Spaces#edit` page.

### DIFF
--- a/app/controllers/space/agreements_controller.rb
+++ b/app/controllers/space/agreements_controller.rb
@@ -3,6 +3,10 @@ class Space
     def show
     end
 
+    def new
+      agreement
+    end
+
     def create
       agreement.save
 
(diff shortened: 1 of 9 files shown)


app/controllers/space/agreements_controller.rb
```diff
@@ -1,11 +1,28 @@
 class Space
   class AgreementsController < ApplicationController
     def show
-      authorize(agreement)
+    end
+
+    def create
+      agreement.save
+
+      if agreement.errors.present?
```
anaulin, line 9: Unusual. Why not `if agreement.save`?

anaulin (approved):
Specs need fixing before merge

app/controllers/space/agreements_controller.rb
zspencer: Ah sorry it's because I've been working in a code base with `decent_exposure` and `responders`. In those, the ergonomics tend to push you towards checking for the presence of errors since the `save` call often happens  automatically.

PR #1378, opened 2023-04-17
🥗 `Marketplace`: Woops, both `Delivery` patches worked independently...

zspencer:

I suppose this is why the "Prevent Merging unless Branch is Up To Date With Upstream" button is a thing...

This should fix the main branch, but I'm curious about what people think about adding constraints to not allow merging branches that are behind main?


zspencer:
So, we *had* branch protection set up to make sure stuff is up-to-date, but if there aren't any status checks required it doeesn't work apparently?!

anaulin (approved):
I am pro-constraint "can't merge unless up-to-date with `main`", because of stuff like this.

I've learned the hard way that if you rename a "required" check in the YAML file (e.g. instead of "Run Tests" you rename it to "Run tests"), it simply becomes "not required", you can merge without it passing, and GitHub doesn't even warn you. 🤷🏼‍♀️

zspencer:
Love to hear it 😭

PR #1396, opened 2023-04-20
✨ `Marketplace`: ask for missing delivery details to proceed to checkout

anaulin:

This can use a lot more UI / visual love, but at least it is a sketch of a slightly improved experience: when we are missing delivery details, instead of just hiding the "Checkout" button, show a primary button asking for the delivery details.

### With this change
https://user-images.githubusercontent.com/6729309/233237901-3c4f0bf9-1c56-417b-9be0-8c53686ac1a3.mp4

### DIFF
--- a/app/furniture/marketplace/cart.rb
+++ b/app/furniture/marketplace/cart.rb
@@ -43,12 +43,8 @@ class Marketplace
       product_total + delivery_fee + tax_total
     end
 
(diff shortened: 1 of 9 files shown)


zspencer (approved):
I like that this brings the buttons that are dependent on each other closer together!

PR #1408, opened 2023-04-26
`Components`: Isolate why Lookbook is sad sometimes

zspencer:

OK, so from what I can tell, it is related to nested constants. LookBook (or maybe ViewComponent?!) does not appear to follow the Zeitwerk-style constant auto-loading rules; and rather depends on it's own internal system for detecting changes to files and reloading the Ruby AST.

When you change a ComponentPreview that is not namespaced, it happily reloads and everything is great. When you change a ComponentPreview that *is* namespaced it throws a giant fit.

I have not yet created an isolated reproduction (Fresh Rails 7, Ruby 3.1, ViewComponent, etc) but I figured I would push up at least the changes I fiddled with to figure that out last Wednesday.

### DIFF
--- a/spec/components/previews/button_component_preview.rb
+++ b/spec/components/previews/button_component_preview.rb
@@ -1,10 +1,11 @@
 class ButtonComponentPreview < ViewComponent::Preview
-  def default
+  # @param classes
+  def default(classes: "font-bold bg-rose-500 text-white hover:text-red-50 p-5 no-underline")
     render(ButtonComponent.new(
       label: "I'm a basic button with custom classes",
       title: "gaaa!",
       href: "#",
-      classes: "font-bold bg-rose-500 text-white hover:text-red-50 p-5 no-underline"
+      classes: classes
     ))
   end
 
--- a/spec/components/previews/marketplace/product_component_preview.rb
+++ b/spec/components/previews/marketplace/product_component_preview.rb
@@ -2,8 +2,8 @@ class Marketplace
   class ProductComponentPreview < ViewComponent::Preview
     # @param name
     # @param price
-    def card(name: "Fancy Pants", price: 8.5)
-      render(ProductComponent.new(product: Product.new(name: name, price: price)))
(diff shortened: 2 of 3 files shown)


anaulin (approved):

zspencer:
Yesterday I dug in further, and it does appear to be a `Lookbook` specific issue; since the `Previews` are still available at `rails/view_components`.

I spelunked through as much of the stack as I could to figure it out; and it appears that *sometimes* there is a `ThreadError` exception raised deep in the bowels of the framework trace. I'll pull it out and capture it next time I get it.

I think next steps are to either:

1. Make an isolated Rails / ViewComponent / Lookbook repo and create an issue in `ViewComponent/Lookbook` asking for help identifying why the ViewComponents get discarded
2. Wait for `Lookbook` v2 to be released fully?

Thoughts @anaulin ?

anaulin:
@zspencer my take is that this is not _that_ much of an annoyance for now, so I think it is up to you how much more work do you want to put into this vs punting for some future time (e.g. when v2 is released). You already did great work here diving deep and isolating the issue to at least Lookbook!

Probably unrelated, but your comment about the occasional `ThreadException` made me think of how on one of the projects I work on (Rails 6.1), since the latest Puma update (~2 weeks ago), there is an intermittent issue where if you get an exception in development, it seems like code in that worker stops being auto-reloaded; we _think_ it has to do with the Puma update, but haven't really isolated it yet. Now I'm wondering if there is something deep in the Puma bowels that interacts poorly with Lookbook. Or possibly this is entirely unrelated. 🤷🏼‍♀️

anaulin:
I guess we're on Lookbook v2 now: https://github.com/zinc-collective/convene/pull/1415#pullrequestreview-1405070008

zspencer:
That's exciting! I think I'm going to move on with my life for now; and hope that as lookbook and puma get updated it fixes itself. I know I should report it to the lookbook folks, but I don't have the bandwidth to put together a good enough isolated example...

PR #1419, opened 2023-04-30
🛠️ `Neighborhood`: Use own devcontainer and persist user data

daltonrpruitt:
- Supercedes #1256. 

View it for earlier details.


zspencer (approved):
OMG YESSSSS IT WORKED! My codespace booted in < 3m and started back up in like, 10s!!! AND THE DATA WAS STILL THERE!

🎉🎉🎉

PR #1420, opened 2023-04-30
🥗🧹 `Marketplace`: `Delivery` constraints live on `DeliveryArea`

zspencer:

Allowing a `Marketplace` to specify it's delivery fees, constraints, etc. and a `DeliveryArea` to override them was putting us in a bit of a mess from "how does data even?!"

This consolidates everything onto `DeliveryArea`, yay!


anaulin (approved):

app/furniture/marketplace/cart/delivery_expectations_component.rb
```diff
@@ -3,14 +3,18 @@ class Cart
     class DeliveryExpectationsComponent < ApplicationComponent
       attr_accessor :cart, :delivery_window, :order_by
 
-      def initialize(cart:, order_by: cart.marketplace.order_by,
-        delivery_window: cart.marketplace.delivery_window, **kwargs)
+      def initialize(cart:, order_by: cart.delivery_area&.order_by,
+        delivery_window: cart.delivery_area&.delivery_window, **kwargs)
```
anaulin, line 7: Is this ever used with a different `order_by` and `delivery_window` passed in? It feels like this could be tightened to just accept `cart`.

Not part of your changes, just thinking out loud.

zspencer: That would be nice! I don't remember exactly why I had done it this way the first time; but I think it was because I was using component previews or something to test.

PR #1422, opened 2023-04-30
🥗🥔✨🧹 `Marketplace`: `DeliveryConstraint`s only set in `DeliveryArea`

zspencer:
- Prerequisite for https://github.com/zinc-collective/convene/pull/1420

This does not get rid of the reading of the fields; but it does make it so they are no longer writable; which will make the "stop reading from them!" PR a bit tidier.


app/furniture/marketplace/cart.rb
```diff
@@ -12,6 +12,8 @@ class Cart < Record
 
     belongs_to :shopper, inverse_of: :carts
 
+    belongs_to :delivery_area, inverse_of: :carts, optional: true
+
```
anaulin, line 16: It feels like a cart should _always_ have a Delivery Area (and the Marketplace should always have at least one Delivery Area). That way there is always a place for delivery fee, etc.

zspencer: I agree, but that felt like a bigger step. Right now, every `Marketplace` has multiple-delivery areas, and there is an affordance for setting the DeliveryArea. If we were to make it required prior-to-cart-creation, we would have to re-work the UI to support picking a Delivery Area before we create the Cart.

KellyAH (approved):

PR #1423, opened 2023-04-30
🧹 `Marketplace`: return only `Marketplace`s!

zspencer:

There's something appealing about switching to Rails' built in `Single Table Inheritance` for `Furniture`, rather than our flyweight registry; but I also know that STI can be a bit of a smell...

Anyway, now when you call `Marketplace.all` and get *only* `Marketplace`s. Maybe someday we'll pull this up in to the `Furniture` classes responsibilities so we don't have to do it on every single piece of Furniture...


KellyAH (approved):
💎

PR #1432, opened 2023-05-04
🧹🥔✨ `Marketplace`: Move `time_like?` logic for delivery window into presentation layer, remove Window class and type

anaulin:
Another step towards rationalizing and tightening the "delivery window" story: for now, these are simply strings (which might be formatted as a time for presentation).

Also tweaks one word in the "delivery expectations" component.

* https://github.com/zinc-collective/convene/issues/1185

### Before
image

### After
image


zspencer (approved):
I will one day find a use-case for `ActiveModel::Type` that works really really well, but I agree it's not the right call here.

PR #1440, opened 2023-05-05
✨ `Marketplace`: Display not-ready-for-shopping Marketplace to Space members, with a notice

anaulin:
* https://github.com/zinc-collective/convene/issues/831

This PR brings back displaying a not-ready-yet Marketplace to the space members, together with a notice about missing configuration.

Also adds a new `AlertComponent` for displaying notices.

### Before
image

### After: for space member
image

### After: for logged-out user
image

### Alert Component previews
image

image


app/components/alert_component.rb
```diff
@@ -0,0 +1,70 @@
...
+    @scheme = scheme
+    @title = title
+    @icon = icon
+
+    super(data: data, **kwargs)
+  end
+
+  private
+
+  def bg_class
```
anaulin, line 19: Maybe these would be better as hashes, instead of methods? 🤷🏼‍♀️

app/furniture/marketplace/locales/en.yml
```diff
@@ -37,6 +37,10 @@ en:
       manage_header: "Marketplace"
       configuration_menu: "Configuration"
       edit: "Manage Marketplace"
+      config_missing_title: "Marketplace configuration missing"
+      config_missing_explainer: |
+        This marketplace is not ready for shopping, and is not visible to the public.
+        Add missing configuration by going to 'Manage Marketplace'.
```
zspencer, line 43: Any chance we could make `Manage Marketplace` a link? I don't know how i18n handles html tho; but I imagine something like`t("..", edit_link: link_to(t('marketplace.marketplaces.edit.link_to'), marketplace.location(:edit))` may work?

app/components/alert_component.rb
zspencer: No idea, let's see how it goes for a bit.

app/components/alert_component.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+class AlertComponent < ApplicationComponent
```
zspencer, line 3: Nice! This will be useful in lots of places!

zspencer (approved):
Ship it!

app/furniture/marketplace/locales/en.yml
anaulin: Done.

Issue #1430, opened 2023-05-04
🌱🥔✨ `Utility`: `AWS::S3`


zspencer:
- https://github.com/zinc-collective/convene/issues/252

## Used By

- https://github.com/zinc-collective/convene/issues/1428
- https://github.com/zinc-collective/convene-journal/issues/1
- https://github.com/zinc-collective/convene/issues/201

## Use Cases
- [x] `Contributor` may use standard `ActiveStorage` syntax to enable `Uploads` on `Furniture`

## Rough Plan

We probably want each `Space` to have their own S3 bucket.  However, we don't want each client to *need* an AWS account, so we are hoping to have a Neighborhood level `AWS` (or `Vultr` or *insert provider here* configuration) which we pass through the costs to the individual Space owners.

There's a doc on how to set up ActiveStorage so we can upload files to a Space-by-Space S3 Bucket: https://blog.geogo.in/activestorage-dont-limit-yourself-in-a-single-bucket-4af165adbf84

anaulin:
I don't see the need to have AWS S3 as a configurable Utility for now, or maybe ever. I don't think anyone running a Space needs to be bothered to have their own AWS account, and a hypothetical Neighborhood operator already can configure AWS for their Neighborhood via env variables.

Having a per-space bucket, while nice, doesn't seems necessary for the use cases that we have for now (Spotlight, and soon Marketplace product photos and Journal attachments), because the amount of storage use that these features generate is unlikely to be so big that we'll need to figure out how to pass through the costs to the Space owners. At least not in the short or mid term.

We can organize objects in S3 into sets that can be easily mapped to a space by using folders and/or a common naming schema (e.g. by having a folder name being the space id, or by always having the space id in the object name). Although with our current level of non-usage, I'd argue that this is also not really a requirement or a blocker for adding product photos.

zspencer:
To be clear, a `Utility` in the `Convene` domain is how I think of all interfaces between a Space or Neighborhood and an outside service. Whether the `AWS::S3` `Utility` is attachable to a `Space` or `Neighborhood` or configurable during runtime, build-time, or deploy-time is less important to me than:

- Communicating to potential `Operators` and `Clients` that Convene can store files on AWS S3
- Consolidating the code for dealing with AWS in one place (once we hit rule-of-three)

I think on the 10-month time-scale, one-bucket-per-neighborhood is fine so long as we have a way to safely navigate:

[...]

Issue #1428, opened 2023-05-03
🥡 `Marketplace`: `Product::Photo`s


zspencer:
- https://github.com/zinc-collective/convene/issues/1324
- https://github.com/zinc-collective/convene/issues/1326

### Prerequisites
- [x] https://github.com/zinc-collective/convene/issues/1430

### Use Cases

- [x] Seller uploads `Product::Photo`
  - [ ] Seller is informed of the aspect ratio and resolution of a good `Product::Photo`.
- [x] Shopper sees `Product::Photo`

### Refactor
- [ ] #2138
### Bugs

- [x] 🐞 `Product::Photo` moves content on the page as the image loads

zspencer:
One thing I noticed in testing is that when the photo loads in it does not have a set size; so the page will shuffle around a little bit.

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #1441, opened 2023-05-05
🌸✨ `Marketplace`: `Seller` may add `Photo` to `Product`

anaulin:
- https://github.com/zinc-collective/convene/issues/1430

This adds very basic support for adding photos to a `Marketplace::Product`, including:
* ability to upload a photo to the product
* the Active Storage blob filename is automatically prepended with the `space.id`
  * I tried making it into what S3 will consider a "folder", turning the filename into `space.id/<filename>`, but the default Active Storage behavior is to strip all directories, so that didn't work
* adds `vips` image library installation as part of setup
  * supposedly the default Heroku Ruby buildpack now includes vips, so this should work on Heroku as well (see: https://github.com/heroku/heroku-buildpack-ruby/issues/1200)
  
### WIP screenshots
image
  
image

  
Still todo:
* [ ] display uploaded image in the cart / user-facing marketplace
* [ ] correctly display the existing image in the product form when editing an existing product
* [ ] decide if we want to automatically resize photos to a certain maximum size
* [ ] add tests


app/furniture/marketplace/product.rb
```diff
@@ -25,5 +27,15 @@ class Product < Record
...
 
     monetize :price_cents
+
+    before_commit :standardize_attachment_name
+
+    def standardize_attachment_name
+      return unless photo.attached? && photo.blob.persisted?
+      return if photo.blob.filename.to_s.start_with?(space.id.to_s)
+
+      new_name = "#{space.id}-#{photo.blob.filename}"
```
anaulin, line 37: It turns out that my naive approach to make this `space.id/...` didn't work, because Active Storage strips out the directories.

@zspencer is pre-pending the filenames with the space id (but having all files be together in the bucket) enough for our needs right now?

zspencer: I think so, I still think bucket-per-space is a better strategy for segmenting data but I can live with prefixing with space-id for now.

anaulin: PS. If we end up going with this approach, and liking it, I imagine that this callback method could be extracted into a module to be shared by any model that has attachments.

zspencer (commented):
V exciting! Thank you!

anaulin:
I probably won't be able to move this forward until next Wednesday's ensemble. If anyone else feels like picking up this PR and moving it forward before then, please feel free!

zspencer:
For now, I'm thinking we take the photos as the maximum size they are; and we add variants on a per-location basis. I'm not sure if this is a long-term good idea; but in the near term it should give us flexibility for figuring out what the sizing should be.

zspencer:
I have a set of PRs that are ready to merge into this that fulfill the remaining checkboxes. @anaulin - Do you want to review / merge them before we merge this or?

anaulin:
@zspencer brilliant! feel free to merge ALL the things!

zspencer (approved):
I approooooove

PR #1447, opened 2023-05-07
🛠️ `Neighborhood`: Add some `lint` scripts

zspencer:

I normally just write these every time, but eh maybe someone else will benefit from me putting them in here.

I've also seen a really lovely `make`-a-like written in Rust that the former CTO of NPM seems to like; but I can't remember it's name 🙃.


anaulin (approved):
Thank you! (I always rely on my shell history, which is really not the best.)

PR #1448, opened 2023-05-07
🥗🧹 `Marketplace`: Extract and test `MarketplaceComponent`

zspencer:

I'm not particularly happy with how I'm doing the render test... But calling `output.blank?` didn't work; because apparently a fragment counts as not-blank, even if it's empty...


app/furniture/marketplace/marketplace_component.rb
```diff
@@ -0,0 +1,27 @@
+class Marketplace
+  class MarketplaceComponent < ApplicationComponent
+    attr_accessor :marketplace
+    delegate :location, :carts, :ready_for_shopping?, to: :marketplace
+
+    def initialize(marketplace:, data: {}, classes: "", **kwargs)
+      super(data: data, classes: classes, **kwargs)
```
anaulin, line 7: I think you could just leave the `data` and `classes` parameters to be part of `**kwargs` and not have to deal with them explicitly.

Relatedly, recently I came across `dry-initializer`, which the Evil Martians folks seem to like to use for their ViewComponents, and I am wondering if that would be a good way for us to tighten up our component constructors as well:
https://dry-rb.org/gems/dry-initializer/3.0/inheritance/
https://github.com/palkan/view_component-contrib#hanging-initialize-out-to-dry

app/furniture/marketplace/marketplace_component.html.erb
```diff
@@ -0,0 +1,18 @@
+<% if !ready_for_shopping? && policy(marketplace).edit? %>
+  <%= render AlertComponent.new(scheme: :warning, icon: :exclamation_triangle,
+                                title: t('marketplace.marketplace.config_missing_title')) do %>
+    <%= t('marketplace.marketplace.config_missing_explainer_html',
+          edit_link: link_to(t('marketplace.marketplace.edit'), location(:edit)) ) %>
+  <% end %>
+<% end %>
+
+<%- if ready_for_shopping? || policy(marketplace).edit? %>
```
anaulin, line 9: This is now the same as `render?`, so we could remove the `if` altogether.

anaulin (approved):
Beautiful, thanks!

app/furniture/marketplace/marketplace_component.rb
zspencer: Ooo, I do like a lot of the the Evil Martians gems! I'd be down to give it a shot!

PR #1460, opened 2023-05-11
`Marketplace`: Further limit display size of `Product` photos

anaulin:
* https://github.com/zinc-collective/convene/issues/1428

I feel like 500x500 can easily overwhelm the page, so I'm limiting the display size further. I picked this size based on some comparisons for other similar apps:

Yelp:
image

Instacart:
image

image

#### Before

image

image

image

#### After
image

image

image

image

### DIFF
--- a/app/furniture/marketplace/cart_products/_cart_product.html.erb
+++ b/app/furniture/marketplace/cart_products/_cart_product.html.erb
@@ -4,7 +4,7 @@
   <td class="w-full max-w-0 py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:w-auto sm:max-w-none sm:pl-6">
     <%- if product.photo.present? %>
       <figure>
-        <%= image_tag product.photo.variant(resize_to_limit: [550, 550]).processed.url, class: "rounded-lg"
+        <%= image_tag product.photo.variant(resize_to_limit: [150, 150]).processed.url, class: "rounded-lg"
  %>
         <figcaption class="text-center py-2">
           <%=product.name%>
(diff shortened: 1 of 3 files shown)


app/furniture/marketplace/cart_products/_cart_product.html.erb
```diff
@@ -4,7 +4,7 @@
   <td class="w-full max-w-0 py-4 pl-4 pr-3 text-sm font-medium text-gray-900 sm:w-auto sm:max-w-none sm:pl-6">
     <%- if product.photo.present? %>
       <figure>
-        <%= image_tag product.photo.variant(resize_to_limit: [550, 550]).processed.url, class: "rounded-lg"
+        <%= image_tag product.photo.variant(resize_to_limit: [150, 150]).processed.url, class: "rounded-lg"
```
zspencer, line 7: We could add named variants if we'd like: https://www.bigbinary.com/blog/rails-7-adds-ability-to-use-predefined-variants#rails-7-onwards

But I also think that keeping the sizes in the UI layer (`View` or `ViewComponent` are both good)

anaulin: We decided to leave it in the views and components for now, keep it in the views instead of the model.

app/furniture/marketplace/products/_form.html.erb
```diff
@@ -6,7 +6,7 @@
     <%= render "collection_check_boxes", { attribute: :tax_rate_ids, collection: marketplace.tax_rates, value_method: :id, text_method: :label, form: f} %>
     <%- if product.photo.present? %>
       <div>
-        <%= image_tag product.photo, class: "rounded-lg" %>
+        <%= image_tag product.photo.variant(resize_to_limit: [150, 150]), class: "rounded-lg" %>
```
zspencer, line 9: Rule of 3 makes me want a `Product::PhotoComponent` 🙃 but also that may be too much!

zspencer (approved):

app/furniture/marketplace/products/_form.html.erb
anaulin: Maybe not too much. 🤔

PR #1461, opened 2023-05-11
🌸 `Spaces`: Prevent Add `Agreement` button from overlapping

KellyAH:
- https://github.com/zinc-collective/convene/issues/1364

Fix the "Add Agreement +" button overlapping agreement text on the `Spaces` edit page.

space.agreements.help_text on the `Spaces` `edit` page.

⚠️ **Warning**: This change could the layout of other elements on other pages that CSS &:last-of-type also applies too. But it's such a small change, it might not have a detrimental affect.

**Before**
before

**After**
after

### DIFF
--- a/app/assets/stylesheets/base/typography.scss
+++ b/app/assets/stylesheets/base/typography.scss
@@ -33,7 +33,7 @@
     @apply mb-2;
 
     &:last-of-type {
-      @apply mb-0;
+      @apply mb-2;
     }
   }
 


app/assets/stylesheets/base/typography.scss
```diff
@@ -33,7 +33,7 @@
     @apply mb-2;
 
     &:last-of-type {
-      @apply mb-0;
+      @apply mb-2;
```
KellyAH, line 36: This worked but there may be a better fix.

I tried adding a tailwind margin setting to the `base_classes` of app/components/button_component.rb and to the agreement text
```
    <p class="text-sm italic"><%= t('space.agreements.help_text') %></p>
```
in app/views/spaces/edit.html.erb but it did not fix the issue.

zspencer, line 36: We may be able to get rid of the `&:last-of-type { ` targeting completely; since the `@apply mb-2` is included in the higher selector

zspencer (approved):
Ship it!

PR #1468, opened 2023-05-14
🧹 `Marketplace`: Add Foreign key for `TaxRate#bazaar`

zspencer:

See: https://github.com/ankane/strong_migrations#adding-a-foreign-key

Apply after https://github.com/zinc-collective/convene/pull/1466/


anaulin (approved):
🗝️

PR #1477, opened 2023-05-18
🛠️ `Neighborhood` : Deploy to render

daltonrpruitt:
This is starting as just an attempt at deploying to [Render.com](https://render.com/), following various [tutorials](https://render.com/docs/deploy-rails)/guides online. And I wanted to track the work somewhere in convene.

In a fork for now so I can deploy to Render without needing admin privileges in root convene.

If this works in some capacity, I will add documentation, but aside from configuring the link between Render and GitHub, I'm hoping most of this work will be file configuration setup that users wouldn't have to do anyways.


daltonrpruitt:
If we add a zinc account and add me as a member (and set limits to prevent any mess up on my end from costing excessive fees...), I can get rid of this entire fork and just work in convene directly.

zspencer:
@anaulin - are you available to set up a Convene Rendr account? Or should we keep things constrained to Heroku for now despite their ossification?

anaulin:
I can look into setting up a Zinc Render account, hopefully between now and next Wednesday.

anaulin:
I created a Render account, owned by operations@zinc.coop (password is in shared Operations vault). To have a "team" with the ability to add other people, we'll be paying $19/mo/user + compute costs, which I think is fine while we experiment.

@daltonrpruitt please let me know which e-mail address should I invite to the Render account, so that you can have access.

zspencer:
Thanks @anaulin ! We may wind up shutting this trajectory down since the $20/mo/user cost is... pretty absurd even compared to Heroku's pricing.

Since we're doing this mostly to figure out what options to present to other Neighborhood Operators; I'd have a hard time recommending something at that price tier that I haven't used in production.

I guess we stick to Heroku + Self-Deployment on Ubuntu 22.04?

config/environments/production.rb
```diff
@@ -22,7 +22,7 @@
 
   # Disable serving static files from the `/public` folder by default since
   # Apache or NGINX already handles this.
-  config.public_file_server.enabled = ENV["RAILS_SERVE_STATIC_FILES"].present?
+  config.public_file_server.enabled = ENV["RAILS_SERVE_STATIC_FILES"].present? || ENV['RENDER'].present?
```
zspencer, line 25: I would rather we instructed folks to set the `RAILS_SERVE_STATIC_FILES` variable instead of adding a `RENDER` check

config/database.yml
```diff
@@ -80,6 +80,7 @@ test:
 #
 production:
   <<: *default
-  database: convene_web_production
-  username: convene_web
-  password: <%= ENV['CONVENE_WEB_DATABASE_PASSWORD'] %>
+  url: <%= ENV['DATABASE_URL'] %>
```
zspencer, line 83: Oh this is good; we should just do this everywhere.

config/puma.rb
```diff
@@ -25,14 +25,14 @@
...
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
+workers ENV.fetch("WEB_CONCURRENCY") { 2 }
 
 # Use the `preload_app!` method when specifying a `workers` number.
 # This directive tells Puma to first boot the application and load code
 # before forking the application. This takes advantage of Copy On Write
 # process behavior so workers use less memory.
 #
-# preload_app!
+preload_app!
```
zspencer, line 35: Thank you for noticing this! I think Heroku also recommends it?

render.yaml
```diff
@@ -0,0 +1,23 @@
...
+    buildCommand: "./bin/render-build.sh"
+    startCommand: "bundle exec rails s"
+    envVars:
+      - key: DATABASE_URL
+        fromDatabase:
+          name: postgres
+          property: connectionString
+      # - key: RAILS_MASTER_KEY
+      #   sync: false
+  - type: redis
```
zspencer, line 19: It's interesting that redis is a "service" while Postgres is a "database"

.gitignore
```diff
@@ -30,3 +30,5 @@ app/assets/builds/
 .yardoc
 
 .devcontainer/output
+
+/config/master.key
```
zspencer, line 34: Smert

render.yaml
```diff
@@ -0,0 +1,23 @@
...
+    name: rails
+    env: ruby
+    buildCommand: "./bin/render-build.sh"
+    startCommand: "bundle exec rails s"
+    envVars:
+      - key: DATABASE_URL
+        fromDatabase:
+          name: postgres
+          property: connectionString
+      # - key: RAILS_MASTER_KEY
```
zspencer, line 17: OK! I think I remember now what the master key stuff is about. We don't use `config/credentials.yml.enc` so we should not need this.

zspencer: That said, every neighborhood needs to generate a random `LOCKBOX_MASTER_KEY`

zspencer (approved):
I think this is a great start; is as a decent introduction to deploying Convene on a Platform As a Service such as Render. 

We'll have to think about how much we want to invest in this particular direction; but getting through to a full working deploy still feels like a worthy endeavor.

daltonrpruitt:
Yeah.... sticking with Heroku is probably wise if it doesn't have that team cost. I think avoiding the team thing is why I was trying to do all of this in a fork and with my own account, but like you said Zee, it isn't scalable for us or others long-term. My bad for not realizing that sooner. 

Heroku's teams functionality is included for teams of 5 or less, and the Econ plan is $5/month for the project (I think?), so much more doable. 

I will start looking more at documenting the Heroku process.

.gitignore
zspencer: it looks like it can be set via an environment variable as well - https://edgeguides.rubyonrails.org/security.html#custom-credentials

anaulin:
Fair enough. I've downgraded the Zinc Render account to "Individual".

zspencer:
>  My bad for not realizing that sooner.

We're just stepping forward down a path of exploration; learning as we go.

But yea, bummer that the pricing model for Render isn't great; but I'm glad we explored it rather than disqualifying it out right because Heroku is what we know!

config/puma.rb
zspencer: It does! https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#preload-app

daltonrpruitt: I only did this because the render docs showed it in their example. I have not actually dug in to see what this does in practice (and I especially didn't dig into Puma yet). But glad it may help sometime! May want to move some of this into a different PR, or repurpose this one as a general spruce-up of deployment stuff.

render.yaml
daltonrpruitt: Based on my understanding, redis is firstly an in-memory storage used for caching purposes, and is not **technically** intended to be for used where persistence is critical (in practice I believe this isn't absolutely true due to its default 1-second flush to file settings, but I can't say I've tested anything in this area yet). Without looking more into it, this may be the reason why.

.devcontainer/devcontainer.json
```diff
@@ -2,7 +2,7 @@
 // README at: https://github.com/devcontainers/templates/tree/main/src/ruby-rails-postgres
 {
   "name": "convene devcontainer",
-  "image": "ghcr.io/zinc-collective/convene-devcontainer:latest",
+  "image": "ghcr.io/daltonrpruitt/convene-devcontainer:latest",
```
daltonrpruitt, line 5: Marking for change later (lazy now)

bin/render-build.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# exit on error
+set -o errexit
+
+bundle install
+bundle exec rake assets:precompile
+bundle exec rake assets:clean
+bundle exec rake db:migrate
```
daltonrpruitt, line 8: File not necessary if not at least testing render deployment.

render.yaml
```diff
@@ -0,0 +1,23 @@
+databases:
```
daltonrpruitt, line 1: File not needed if not doing render deployment.

config/puma.rb
zspencer: Yea, I think an independent PR to enabling preloading app + workers is a good idea; especially so we can evaluate how it impacts using `binding.pry`/`byebug` in development and test mode.

render.yaml
zspencer: Ahhh that makes sense. Definitely seems like a render-specific thing.

.gitignore
daltonrpruitt: Also, this was added automatically by whatever command was used to generate the key in the first place. Good for protecting me from myself....

anaulin:
Sounds good, no objections to closing this!

zspencer:
No objection on my end.

PR #1478, opened 2023-05-18
`Marketplace`: Simplify Management component and make it work better on small screens

anaulin:
* https://github.com/zinc-collective/convene/issues/1324

This improves the behavior of this component on small screens, and also moves some of the logic into re-usable parts:
* adds an optional `footer` section to Card component
* adds an optional `icon` to Button component
* moves all SVG icons into the Svg component

#### Before
image

image

#### After
image

image


zspencer (approved):
Yeah!

PR #1480, opened 2023-05-18
`Space`: break up `edit` page into partials, and start unifying it to use Card component for each area

anaulin:
* https://github.com/zinc-collective/convene/issues/1154

Baby steps towards a better Space management page.

### Before
image

### After
image


zspencer (approved):
Yesssssss

Issue #1481, opened 2023-05-18
🌸✨ `Members`: `Invitation` email should convey which `Space` and `Person` it came from


anaulin:
- https://github.com/zinc-collective/convene/issues/1514

The current Space invitation email doesn't have much info about who / what is actually inviting you and to what, unless you already know what Convene is:

![image](https://github.com/zinc-collective/convene/assets/6729309/6483d058-7abd-471b-8a55-58cfc0af90a8)
![image](https://github.com/zinc-collective/convene/assets/6729309/ecafe8f6-fcf5-4c70-84e2-211956149a06)

It would be good to:
- [x] add the Space name to the subject line (e.g. "You are invited to SPACE NAME", instead of "Invitation email") https://github.com/zinc-collective/convene/pull/1483
- [x] add the name of the invitor to either the subject line (e.g. "PERSON NAME invites you  SPACE NAME") or the "from" text (e.g. "Convene Support on behalf of PERSON NAME")

anaulin:
@KellyAH assigning this to you in case you are looking for something to do, but don't feel obliged to do it!

KellyAH:
Ooo exciting! I've never worked with the Rails Mailer templates before. Thanks so much! I'll give it a shot today and this weekend!

anaulin:
@KellyAH thank you! Don't hesitate to send questions if you get stuck.

And in case you aren't aware of this feature: Rails mailer previews can be very helpful, see https://guides.rubyonrails.org/action_mailer_basics.html#previewing-emails

zspencer:
Great catch! I've linked the card to the "nearest-upstream-leaf" in the product garden wiki thing. We probably should split `Invitations` into it's own high level issue since `Members` is getting pretty overloaded; but I will do that "someday when I think about it."

zspencer:
> Sentry issue: [CONVENE-1H](https://zinc-collective.sentry.io/issues/4210828106/?referrer=github_integration)

Whoops! Apparently sending an email with a `from` field missing an actual *email address* does not uhhh.... work...

We'll probably want to:

- [ ] Fix the bug
- [ ] Investigate adjusting `maildev` to angrily reject malformed emails
- [ ] Included the email address set in the `EMAIL_DEFAULT_FROM` variable.

PR #1483, opened 2023-05-20
🌸✨ `Members`: `Invitation` email conveys which `Space` and `Person` it came from

KellyAH:
- for https://github.com/zinc-collective/convene/issues/1481
---
# Changes
- modify invitation email `From`, `Subject`, `Body` text to more clearly communicate the invitor's name and the `Space` the invitee is being asked to join.

before

**After**
Screenshot 2023-05-21 at 8 20 59 PM


KellyAH:
IMHO adding the invitor name to the Subject could be too wordy. What do ya'll think? @anaulin @zspencer 

too wordy maybe

anaulin:
What about putting the invite name in the "from" field I stead?

It says "Neighborhood support", but maybe it could say something like "INVITOR NAME via NEIGHBORHOOD". Or some other wording that you prefer. The main point is to convey both the invite name somewhere in the "from" or the "subject".

For inspiration, here are some screencaps of invites from other systems that I found in my inbox:

zspencer (approved):
Doh, I think I should have left most of my comments from the other PR on this one! Apologies!

I would prefer if we:
- Did not add an additional parameter to the `SpaceInvitationMailer`
- Kept the messaging focused on our client's Space, rather than on Convene

But I'm available for you to merge this as is if you want to get it nailed down.

KellyAH:
Let me rework this on Sunday from @anaulin 's and @zspencer 's suggestions, feedback and requirements. I think I can meet everyone's needs.

KellyAH:
My bad 🤦‍♀️ . I confused the `From` field for the email ending salutation. Thanks for the examples @anaulin  they helped a lot.

Ok I fixed it up. Hope it's better now!

KellyAH:
🤔 Darn browser tests in `responding_to_invitations.feature` is failing in parts of code I don't think I touched.

`AssertionError [ERR_ASSERTION]: By(css selector, *[aria-label="Configure Space"])`

`Couldn't find email {"to":"katara-9d14ec59-5507-4416-8e8d-74ce32c3c734@example.com"}`

Can't see the browser when tests run in code spaces. 😞  I'll need to spin up local again.....

zspencer:
Sometimes the browser specs flake, If you re-run them they often pass. I've kicked off a re-run to see if that's the case here.

app/views/space_invitation_mailer/space_invitation_email.text.erb
```diff
@@ -1,9 +1,9 @@
 Hi <%= @invitation.name %>,
 
-You've been invited to <%= @invitation.space.name %> by <%= @invitation.invitor_display_name %>.
+<%= @invitation.invitor_display_name %> has invited you to join the <%= @invitation.space.name %> Space.
```
zspencer, line 3: Let's ditch the word "Space" from this; since the space name should stand alone.

zspencer (approved):
Thank you this is a great improvement!

It may be cool (but not particularly urgent) to set up something where we allow the Space to customize the template email copy using something like [Liquid](https://shopify.github.io/liquid/)

zspencer:
OK, with the re-run I'm guessing it's because of how we're searching the sent mail in the test that's failing. I'll take a stab at getting it fixed and push a PR this-a-way.

app/views/space_invitation_mailer/space_invitation_email.text.erb
KellyAH: fixed in a3ecf1e1bc40279f3745269e09862ab73f1ccad7

KellyAH: And removed "the" so it sounds better.
Screenshot 2023-05-22 at 9 03 05 PM

zspencer:
I think this is good to go! We could probably do a follow up PR to try and make the output better when the system tests' email query files (like, why does it just output the "to" when there is also a "text" matcher?! And how do we get a better stack trace?!) but I didn't want to get nerd-sniped by that until I wrap up the tests for fixing the quantitys.

PR #1479, opened 2023-05-18
✨ 🥔 🥗   `Spaces` add Enforce_SSL flag

KellyAH:
- https://github.com/zinc-collective/convene/issues/1473

Add Enforce_SSL to `Spaces`

Issue #1473, opened 2023-05-18
✨🛠️🔐 `Spaces`: Enforce `ssl` connection


zspencer:
- https://github.com/zinc-collective/convene/issues/1154

We're in a bit of a pickle here because we're using Cloudflare's free tier for DNS and WAF for `convene.zinc.coop`; which requires us to disable `force_ssl` for that Space to prevent a redirect loop.

However, for spaces that are *not* using a Web Application Firewall or other intermediary; we *do* want to force ssl.

## Use Cases
- [x] `Operator` may force every space to use SSL by setting the `FORCE_SSL` environment variable to `true` 
- [x] Visitors are redirected to `https` if they visit a space that Enforces SSL using HTTP
- [x] `Member` may force all requests to a `Space` to use `ssl`

zspencer:
@KellyAH - This may be a bit less of a stretch than you wanted; but it will give you experience digging into the Rails `Routing` and `Controller` layers. If you want to tackle it, just leave your face on and roll and tag me on PRs. If you don't just tell me "eh, no" and I'll pick it up :soon:

zspencer:
@KellyAH - When I tested this via the UI in production, the checkbox did not persist in between page-loads. I was able to confirm that when I updated the `enforce_ssl` flag on the `Piikup` space it does do the redirection tho!

KellyAH:
> @KellyAH - When I tested this via the UI in production, the checkbox did not persist in between page-loads. I was able to confirm that when I updated the `enforce_ssl` flag on the `Piikup` space it does do the redirection tho!

So is the Enforce SSL checkbox broken on prod right now?

zspencer:
I didn't have time to do much data collection; but I'm going to set up a production space and get a repro for ya.

zspencer:
OK never mind; clearly I must have not been paying enough attention because it was during the onboarding for Cafe Gabriela.

I *have* created a new space (https://sandbox.zinc.coop/spaces/sandbox/) that we can use for testing stuff in production with lower stakes tho!

KellyAH:
crud! I just spun up local off latest master and it's getting SSL errors! It looks like /spaces pages are auto redirecting to https. Is your local throwing SSL errors too when loading any web pages.

Did the merge of https://github.com/zinc-collective/convene/pull/1504 break our local environments and prod?

[...]

PR #1484, opened 2023-05-20
✨ `Space` add "Website Settings" section and enforce ssl checkbox

KellyAH:
- for https://github.com/zinc-collective/convene/issues/1473
---

# Changes

- Add "Website Settings" section
- Add enforce ssl checkbox to `Spaces` edit page
- only space members can see the new "Website Settings" section.
- [x] Update space policy with `:enforce_ssl` permitted param
- [x] add logic update the `spaces` table `enforce_ssl` with the  checkbox value boolean (dependent on to https://github.com/zinc-collective/convene/pull/1479)
- [x] add unit tests

Screenshot 2023-05-24 at 5 51 16 PM


KellyAH:
# Test Cases to test on Local

- If a user **checks** the Enforce SSL checkbox on `/spaces/system-test/edit`, and then clicks the `Save changes` button, the checkbox stays **checked** and the DB `space` table `enforce_ssl` value is `true`
- If a user **unchecks** the Enforce SSL checkbox on `/spaces/system-test/edit`, and then clicks the `Save changes` button, the checkbox stays **unchecked** and the DB `space` table `enforce_ssl` value is `false`
- Refresh/force refresh `/spaces/system-test/edit` preserves the Enforce SSL checkbox state.

KellyAH:
@zspencer I forgot to ask you where on the `/spaces/<space-name>/edit` page do you want the `Enforce SSL` checkbox to be added?  Utilities Section a different section, or a new `Security` section?

anaulin:
It feels like we're starting to need a new "general space settings" section. (With maybe a better / more concise title, probably.)

zspencer:
For now an "Operator Only Settings" section feels like maybe the right place; that way we could also expose the `branded_domain` field there

app/controllers/invitations_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
 
   def create
     if invitation.save
-      SpaceInvitationMailer.space_invitation_email(invitation).deliver_later
+      SpaceInvitationMailer.space_invitation_email(invitation, space).deliver_later
```
zspencer, line 9: The `Invitation` class exposes a `space` instance method; so we should not need to pass it in as a second parameter.

app/views/space_invitation_mailer/space_invitation_email.text.erb
```diff
@@ -6,4 +6,4 @@ You may view, accept or ignore this invitation at <%= polymorphic_url(@invitatio
 
 We hope to see you there! 🥳🎈🎉
 
-- The <%= @invitation.space.name %> Crew
+- Convene Support on behalf of <%= @invitation.invitor_display_name %>
```
zspencer, line 9: I would prefer if we kept it as `The {space_name} Crew on behalf of {invitor_display_name}` since it's really about inviting them to a specific space.

app/views/spaces/_utilities.html.erb
```diff
@@ -32,4 +32,7 @@
   </ul>
 
   <p><%= link_to t('utilities.new.link_to'), space.location(:new, child: :utility) %></p>
+
+  <%= check_box_tag(:enforce_ssl) %>
+  <%= label_tag(:enforce_ssl, "Enforce SSL") %>
```
zspencer, line 37: This probably wants to be in a [`form_with(model: space.location)`](https://guides.rubyonrails.org/form_helpers.html#binding-a-form-to-an-object) block.

zspencer (commented):
This is a great direction! Hopefully the comments I left were useful!

KellyAH:
Will continue on sunday 5/21

app/views/space_invitation_mailer/space_invitation_email.text.erb
KellyAH: fixed in 7eea88dc691af50a7ec30540f1fb8cee0066103b

app/controllers/invitations_controller.rb
KellyAH: fixed in 483e144e6167707e34e169c5abf9d2f902391108

app/views/spaces/_utilities.html.erb
KellyAH: fixed in 14fd46eacada467268b924c3461c557ec1d140e7 but I can't get the checkbox to be right next to the label text 🤔

KellyAH:
@zspencer should I add browser tests or any unit tests for the new "Operator Only Settings" section? Or is it too early

zspencer:
It would be good to have a `spec/requests/spaces_controller_request_spec#update` test  to make sure that the `enable_ssl` flag gets set to true when the form comes through; but I think a full system test would be too much since we're probably going to move this around soon anyway.

app/views/spaces/_utilities.html.erb
zspencer: Hmmm, you'll probably need to use the `checkbox` or `label` methods  block syntax to put the checkbox within the label; rather than treat them as siblings.

Something like this maybe?

```erb
<%= form.label :enforce_ssl do
  <%= form.checkbox :enforce_ssl %> Enforce SSL
<%- end %>

```

zspencer (approved):
So long as you have confirmed that the data flips when the form submits I am comfortable with you merging this.

zspencer:
Maybe this is actually "Website Settings" and not "Operator Only Settings" 🙃

KellyAH:
> Maybe this is actually "Website Settings" and not "Operator Only Settings" 🙃

Renamed to "Website Settings" in the code and UI in 5910f5c4041f214cda4f1f36fc3995a15e8136f3

app/views/spaces/_utilities.html.erb
KellyAH: fixed in https://github.com/zinc-collective/convene/pull/1484/commits/0343d527e40d6defe99afe9b0294c4e040cb5384 and added a Save button

app/controllers/spaces_controller.rb
```diff
@@ -60,6 +60,10 @@ def space_params
       policy_scope(Space).friendly.find(params[:id])
     elsif params[:space]
       policy_scope(Space).new(space_params)
+    elsif params[:enforce_ssl] == 1
+      policy_scope(Space).friendly.find(params[:id]).update(enforce_ssl: true)
+    elsif params[:enforce_ssl] == 0
+      policy_scope(Space).friendly.find(params[:id]).update(enforce_ssl: false)
```
KellyAH, line 66: Not sure if this is the correct way to update `Space` record based on the checkbox value.

spec/requests/spaces_controller_request_spec.rb
```diff
@@ -124,6 +124,25 @@
...
+        put polymorphic_path(space), params: {space: {entrance_id: new_entrance.id, enforce_ssl: 1}}
+
+        expect(space.enforce_ssl).to be true
+      end
+
+      it "disables Enforce SSL on a Space" do
+        new_entrance = space.rooms.sample
+        put polymorphic_path(space), params: {space: {entrance_id: new_entrance.id, enforce_ssl: 0}}
+
+        expect(space.enforce_ssl).to be false
```
KellyAH, line 143: Tests still fail. IDK ahy.

I see the `Space` record updating `enforce_ssl` to the correct value  in Puma logs when submitting the form on local.

app/controllers/spaces_controller.rb
zspencer: Hmm, this does seem a little off-the-rails to me. Normally, if `space_params` includes `enforce_ssl`  it will update just fine from the browser without needing to do a comparison check; so you shouldn't have to change the controller since you're permitting the new param in the `SpacePolicy`.

app/views/spaces/_website_settings.html.erb
```diff
@@ -0,0 +1,14 @@
...
+    <h3><%= t('website_settings.index.link_to') %></h3>
+    <p class="text-sm italic"><%= t('website_settings.help_text') %></p>
+  </header>
+
+  <%= form_with model: space.location do | web_settings_form | %>
+    <%= web_settings_form.label :enforce_ssl do %>
+      <%= web_settings_form.check_box :enforce_ssl %>
+        Enforce SSL - Automatically redirects all Visitor traffic to this Space to https.
+      <% end %>
+    <%= web_settings_form.submit %>
```
zspencer, line 12: Nice

spec/requests/spaces_controller_request_spec.rb
zspencer: Both the `Controller` `space`  instance and the `spec` `space` instance have the same `id` (since they're referencing the same piece of data where it's stored in the database) but will have a different location in memory; so changes to the `space` instance in the `Controller` will only be reflected in `space` instance in the `spec` once the `space` instance in the `spec` re-queries the database to get the updated value.

TL/DR: You may want to call `space.reload` after executing the `put` because the computer caches the data in the `Space` record independently between the `Controller` and `spec` until it is re-queried..

spec/requests/spaces_controller_request_spec.rb
```diff
@@ -124,6 +124,25 @@
...
       end
     end
+
+    context "when a Operator" do
+      let(:space) { create(:space, :with_members, :with_rooms) }
+      let(:operator) { create(:person, operator: true) }
+
+      it "enables Enforce SSL on a Space" do
+        new_entrance = space.rooms.sample
+        put polymorphic_path(space), params: {space: {entrance_id: new_entrance.id, enforce_ssl: 1}}
```
zspencer, line 134: You can use `false` and `true` in your params (and it is generally preferred)

zspencer (approved):
I think this is good progress; and would be happy to merge it once the tests pass. That said; there are a couple ways to bring this closer to "Rails-normal":

1. Revert the changes in `SpacesController`
2. Use `true` and `false` in your `spaces_controller_request_spec`

PR #1485, opened 2023-05-22
🧹 `Neighborhood`: Remove `SystemTest` from First-Run

zspencer:
- https://github.com/zinc-collective/convene/issues/890

We want people to go through the "Set up your Operator and Space" process every time a new neighborhood is provisioned; the `SystemTestSpace` scaffolding was useful at the beginning because it let us skip building out a better first run experience; but now it's time to start pulling it out.


anaulin (approved):
Thank you!

PR #1486, opened 2023-05-22
`Invitation`: Update the spec to reflect the email subject change

zspencer:
We currently look for the RSVP Link in our system tests by matching against the subject line of the email. Which works! However, we don't do much with our system tests so it's easy to forget.

### DIFF
--- a/features/lib/Invitation.js
+++ b/features/lib/Invitation.js
@@ -68,7 +68,7 @@ class Invitation {
   emails() {
     return this.emailServer().emailsWhere({
       to: this.emailAddress,
-      text: (text) => /You've been invited/.test(text),
+      text: (text) => /invited you to/.test(text),
     });
   }
   /**


KellyAH (approved):

features/lib/Invitation.js
```diff
@@ -68,7 +68,7 @@ class Invitation {
   emails() {
     return this.emailServer().emailsWhere({
       to: this.emailAddress,
-      text: (text) => /You've been invited/.test(text),
+      text: (text) => /invited you to/.test(text),
```
KellyAH, line 71: How would anyone be able to easily  tell  this is the subject line of the email? Does it need a comment or a better key name?

zspencer: That said, I'm not sure this code will exist in 3~6 months if we go further with implementing system specs in Ruby rather than Cucumber/JavaScript

KellyAH: I look forward to implementing system specs in Ruby. Either more unit tests and/or browser tests. I need the practice.

zspencer: @KellyAH - Re: Writing more specs - Maybe it would be worth doing an inventory of the different controllers in the `app/furniture/marketplace` directory and their actions; and then figuring out which controllers / actions are missing corresponding request specs.

From there we can fill them in with per-action tests to cover permutations around:

- Who is signed in?
- What parameters may (or may not) be set in the `#update` and `#create` actions?
- What change(s) if any should they make to the database?
- What do they tell the browser to do (render some html with particular context or views/components? redirect?)

Goal would be to do an initial "what's low-hanging fruit?" pass; with some amount of pushing the "what does the person actually see?" down into the component specs; and any interaction that would require more than 1 HTTP request into system specs (not that we have any system specs yet...)

PR #1488, opened 2023-05-22
`Marketplace`: Don't need `TaxRate` onto Bazaar every deploy

zspencer:

Since they're all attached and will be into the future we don't need to migrate
them on every deploy anymore.


anaulin (approved):
🧹

PR #1487, opened 2023-05-22
🥗🧹 `Marketplace`: Extract and test `CartProductComponent`

zspencer:
- https://github.com/zinc-collective/convene/issues/1326
- https://github.com/zinc-collective/convene/issues/1187

Initially I tried getting the ComponentSpec to kind of act like a system
spec; where I clicked the buttons and then asserted that the page
re-rendered correctly.

Unfortunately, I couldn't get it to work well enough to commit; so I
think I will stab at writing our first system specs in a separate PR.

PR #1492, opened 2023-05-23
🥗 `Marketplace`: Quick spec for the `Marketplaces#show` endpoint

zspencer:
- Pulled out of https://github.com/zinc-collective/convene/pull/1487

Missed a bit in my rename; so I added some tests to the endpoint to catch it
in the future.


anaulin (approved):
🧹

PR #1493, opened 2023-05-23
🥗🐞 `Marketplace`: `Shopper` may change `CartProduct` quantity

zspencer:
- https://github.com/zinc-collective/convene/pull/1487/

This pulls out just the request/response cycle fixes from the ballooning PR. I'll pull the extraction of the `CartProductComponent` into it's own PR that's based off of this one.

### DIFF
--- a/spec/furniture/marketplace/cart_products_controller_request_spec.rb
+++ b/spec/furniture/marketplace/cart_products_controller_request_spec.rb
@@ -5,36 +5,117 @@ RSpec.describe Marketplace::CartProductsController, type: :request do
   let(:space) { marketplace.space }
   let(:room) { marketplace.room }
   let(:member) { create(:membership, space: space).member }
+  let(:document_root_element) { Nokogiri::HTML::Document.parse(response.body) }
+  let(:request_as) { :html }
 
   describe "#create" do
-    it "Add a Product to the Cart" do
-      product = create(:marketplace_product, marketplace: marketplace)
-      cart = create(:marketplace_cart, marketplace: marketplace)
+    subject(:perform_request) do
+      post path, as: request_as, params: {cart_product: {product_id: product.id, quantity: 1}}
+      response
+    end
 
-      post polymorphic_path(cart.location(child: :cart_products), params: {cart_product: {product_id: product.id}})
+    let(:product) { create(:marketplace_product, marketplace: marketplace) }
+    let(:cart) { create(:marketplace_cart, marketplace: marketplace) }
+    let(:path) { polymorphic_path(cart.location(child: :cart_products)) }
+
+    it "Add a Product to the Cart" do
+      perform_request
+
       expect(response).to redirect_to([space, room])
+    end
+
+    context "when a turbo stream" do
+      let(:request_as) { :turbo_stream }
+
+      it "Replaces the cart product, cart footer and cart total" do
+        perform_request && cart.reload
+
+        assert_select("turbo-stream[action='replace'][target='cart-product-#{product.id}']") do
+          assert_select("*[data-cart-product-quantity]", text: "1")
+        end
+
+        assert_select("turbo-stream[action='replace'][target='cart-footer-#{cart.id}']")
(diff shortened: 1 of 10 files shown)


zspencer:
@anaulin - Do you have thoughts on how to tidy up the duplication across the three specs? I was tempted to use the `have_rendered_turbo_stream` matcher, but:

1. I figured this would be more explicit about what is important about what the server sends back to the client
2. The output is an absolute mess

In the [next PR](https://github.com/zinc-collective/convene/pull/1487), I componentize the `_cart_product.html.erb` partial and test the details of the ui; but I'm still not sure how to quickly determine between what is or is not useful to test at the `request` layer outside of database changes / response codes  & redirects / very light checking against the response body.

spec/furniture/marketplace/cart_products_controller_request_spec.rb
```diff
@@ -5,36 +5,117 @@
   let(:space) { marketplace.space }
   let(:room) { marketplace.room }
   let(:member) { create(:membership, space: space).member }
+  let(:document_root_element) { Nokogiri::HTML::Document.parse(response.body) }
+  let(:request_as) { :html }
 
   describe "#create" do
+    subject(:perform_request) do
+      post path, as: request_as, params: {cart_product: {product_id: product.id, quantity: 1}}
+      response
```
anaulin, line 14: I am not sure you need to be returning the `response` here. I think it might be returned by the `post` itself and, if it isn't, you can just access `response` directly in the tests anyway (and that's what you do in the test case for `#update`, lines 49 and 63).

spec/furniture/marketplace/cart_products_controller_request_spec.rb
```diff
@@ -5,36 +5,117 @@
...
+      let(:request_as) { :turbo_stream }
+
+      it "Replaces the cart product, cart footer and cart total" do
+        perform_request && cart.reload
+
+        assert_select("turbo-stream[action='replace'][target='cart-product-#{product.id}']") do
+          assert_select("*[data-cart-product-quantity]", text: "1")
+        end
+
+        assert_select("turbo-stream[action='replace'][target='cart-footer-#{cart.id}']")
```
anaulin, line 37: I think using the `have_rendered_turbo_stream` matcher here would be fine. One idea to make it more expressive may be to use kwargs for the matcher, so it is clear that you are specifying a `target` and an `action` and whatnot to match against.

spec/furniture/marketplace/cart_products_controller_request_spec.rb
```diff
@@ -5,36 +5,117 @@
...
+        assert_select("turbo-stream[action='replace'][target='cart-total-#{cart.id}'] *[data-cart-total]", text: "Total: #{controller.helpers.humanized_money_with_symbol(cart.price_total)}")
```
anaulin, line 38: Maybe the `have_rendered_turbo_stream` matcher could take a block, and then you could pass in some code to match the content?

anaulin (approved):
Thank you for the tests and fix!

Re: testing and duplication, I think the testing you are doing here is what I would do as well: check that the response is successful and renders roughly the right thing, then also check a couple of key pieces in the content (e.g. the total amount).

The tests feel messy because the response is messy, which is just Turbo Streams. The best idea I have to clean that up is to use the helper you wrote, and maybe even expand it (or add other ones for specific cases).

spec/furniture/marketplace/cart_products_controller_request_spec.rb
zspencer: Yea,  I've been looking at the shoulda-matchers api as a potential inspiration for a Turbo/ViewComponent/Rails-Dom-Testing oriented matcher set; but I hadn't really fiddled with it enough to have a concrete proposal. Maybe I'll putz about at it later now that we're getting a wider set of request and component specs.

spec/furniture/marketplace/cart_products_controller_request_spec.rb
zspencer: Hmm, I've always just kind of done this as habit so I can do `it { is_expected.to redirect... }` but if `post` returns the `response` then I guess I don't need to.

PR #1494, opened 2023-05-23
`Marketplace`: Assert the `CartProduct` shows the `Product::Photo`

zspencer:
- https://github.com/zinc-collective/convene/issues/1430

- Pulled from https://github.com/libvips/ruby-vips/blob/master/.github/workflows/test.yml

I'm not particularly pleased with this, but I didn't find a github action; and I'm not entirely sure why the test passes locally but not on CI.


anaulin (approved):
👍🏼

PR #1506, opened 2023-05-25
Improve "Gizmos" section of the Section (aka `room`) configuration view

anaulin:
* https://github.com/zinc-collective/convene/issues/1155

This PR:
* unifies the Gizmos section into one card
* changes the dropdown to require people to select a furniture kind, so that they don't add things by accident
* enforces that a `furniture_kind` value is set, and shows the error in the UI if there is one

It would be nice to improve this further so that it is not possible to submit the "add new" form without a valid type selected.

### Before
image

### After 
image

#### Error message on failed "add gizmo" action
image

### DIFF
--- a/app/controllers/furnitures_controller.rb
+++ b/app/controllers/furnitures_controller.rb
@@ -7,8 +7,8 @@ class FurnituresController < ApplicationController
   end
 
   def create
-    respond_to do |format|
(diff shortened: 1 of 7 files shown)


zspencer (approved):
Thank you!!!

Issue #1515, opened 2023-05-27
🐞 Invitations are not being Delivered


zspencer:
- https://github.com/zinc-collective/convene/issues/1514

As of noon, May 23rd 2023 invitation emails have been down. 

We noticed the problem on May 26th around 7:00pm.

The bug was introduced here: https://github.com/zinc-collective/convene/pull/1483; as prompted by https://github.com/zinc-collective/convene/issues/1481. I've [linked the Sentry issue](https://github.com/zinc-collective/convene/issues/1481#issuecomment-1565158546) and am beginning remediation.

- [x] Fix the bug
- [x] Deploy the Fix
- [x] Confirm the Fix
- [ ] Try and figure out how to throw an error  if the emails we try and send are not well-formed

zspencer:
As of noon, May 23rd 2023 invitation emails have been down. 

We noticed the problem on May 26th.

The bug was introduced here: https://github.com/zinc-collective/convene/pull/1483; we've identified which change request prompted the bug; and will remediate.

zspencer:
PR is in, I'm going to let the deploy go out and then send myself a new invitation to confirm it's fixed.

zspencer:
Deploy is out, beginning testing.

zspencer:
<img width="1394" alt="Screenshot 2023-05-26 at 7 58 22 PM" src="https://github.com/zinc-collective/convene/assets/50284/76227e40-a348-45d0-97b9-f77a9a6ec232">

And it lives!!!

PR #1516, opened 2023-05-27
🔨🐞 `Invitations`: are deliverable again since they come from a valid Email Address

zspencer:
- https://github.com/zinc-collective/convene/issues/1515

Turns out, when I had suggested we drop the Convene bits from the email I did not think about what that would mean from the email address perspective.

While I don't have a Github issue yet for Space's being able to send email from their domain; I think that would be a good thing for us to support at some point so our clients branding stays front and center and we stay in the background.

In the meantime, since it's coming from a convene domain, we should probably include the convene support name as well.

## After
Screenshot 2023-05-26 at 7 42 32 PM

## Before
Screenshot 2023-05-26 at 7 43 03 PM

### DIFF
--- a/app/mailers/space_invitation_mailer.rb
+++ b/app/mailers/space_invitation_mailer.rb
@@ -4,7 +4,7 @@ class SpaceInvitationMailer < ApplicationMailer
     @space = invitation.space
     mail(
       to: @invitation.email,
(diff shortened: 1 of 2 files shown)


zspencer:
I am using my Godlike powers to merge this to fix the bug since it is blocking me from getting my invitation to the new sandbox space.

PR #1519, opened 2023-05-27
🥗🔨🐞 `Neighborhood`: Footer links to the Neighborhood URL

zspencer:

I noticed that the `Convene` link in the footer was always to the current `root_url`, which will be the `Space` when visiting the `Space` via a `Domain`.

This shifts it to use the `APP_ROOT_URL`, and allows `Operator`s to configure the Neighborhood name and Tagline.

### DIFF
--- a/app/components/neighborhood/link_component.rb
+++ b/app/components/neighborhood/link_component.rb
@@ -1,0 +1,13 @@
+class Neighborhood::LinkComponent < ApplicationComponent
+  include ViewComponent::InlineTemplate
+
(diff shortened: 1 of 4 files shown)


zspencer:
If you, like me, would prefer to not forget to run the linter before committing; you can write a git pre-commit hook like so:

`echo "bin/lint-changed" > .git/hooks/pre-commit && chmod +x .git/hooks/pre-commit`

daltonrpruitt (approved):

Issue #1520, opened 2023-05-28
👩‍🎤 `Contributors` and `Maintainers`


zspencer:
- https://github.com/zinc-collective/convene/issues/1520 design and build https://github.com/zinc-collective/convene/issues/709 for https://github.com/zinc-collective/convene/issues/1475

Anyone may be a `Contributor`, so long as they have submitted at least 1 patch to Convene. 

`Maintainers` are `Contributors` who:

1. Have made substantial contributions to `Convene`, _and_
2. Participate in providing guidance and feedback to `Convene` `Contributors`, _and_
3. Are nominated for `Maintainer` status by themselves or another `Contributor` or  `Maintainer`, _and_
4. Are ratified by current `Maintainers`

## Submitting and Applying Patches
- `Contributors` are encouraged to submit Patches to Convene using Github's Pull Request Feature
- `Maintainers` *should* review `Contributors` Pull Requests within 1~2 weeks; depending on availability.
- `Contributors` *may* approve `Maintainers` Pull Requests; but not the pull requests of other `Contributors`.

### Product
  - Research
  - Design
  - [Writing](https://www.interaction-design.org/literature/article/how-to-get-started-with-ux-writing)
  - Documentation
  - Engineering
    - Human/Computer Interface Engineering
  - Localization
  - Testing
    - System Test Automation
    - Accessibility Testing
  - Marketing
    - Educational Marketing
### Infrastructure 
  - Engineer
### Security
  - Tester
  - Engineer

## Supported Tooling
- [x] #1737
- [x] #1764

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #1521, opened 2023-05-28
📝 `Contributors`: Revise `CONTRIBUTING.md`

zspencer:
- https://github.com/zinc-collective/convene/issues/1520

I'm trying to streamline things back down now that we have a better idea of how things are going.

Still not entirely sure how much should live in-repo vs in-issue but I figure we'll get there as we go.

Looking for feedback re:

- Is there anything that feels *important* to keep that was deleted?
- Is there anything that has been added that feels extraneous?


CONTRIBUTING.md
```diff
@@ -1,55 +1,48 @@
...
+### 1.2 Users and Stakeholders.
+Stakeholders for `Convene` fall into one or more of the following categories:
 
-The root directory provides the human and programmatic interface for Convene
-as well as data-persistence; with two additional submodules.
+- [`Clients`], who are the legal owners of the data of their [`Spaces`] in a [`Neighborhood`].
+- [`Members`], have broad authority over [`Spaces`] and their [`Furniture`]. They are also responsible for [`Visitor`: Support], [`Visitor`: Trust & Safety], [`Member`: Support], and [`Member`: Trust & Safety].
+- [`Visitors`], who access [`Spaces`] and their [`Furniture`] according to [`Member`] specification.
+- [`Operators`], who steward the [`Neighborhood`: Infrastructure], [`Neighborhood`: Trust & Safety], and [`Client`: Support] responsibilities
+- [`Contributors`] and [`Maintainers`], who design and build [`Furniture`].
```
daltonrpruitt, line 30: So, are these that don't link to anything yet meant to be linked later?

zspencer: I am also happy to kick this back to drafts and ensure all `[` have links before merging.

CONTRIBUTING.md
```diff
@@ -1,55 +1,48 @@
...
-Each issue labeled `epic` is pretty big. We tend to use checkboxes to decompose
-big issues into smaller pieces of work for people to bite off. We don't expect
-contributors to complete all the work before submitting their first patch.
-Instead, we prefer Contributors to take little bites and submit small patches so
-that we can develop tight, asynchronous feedback loops.
-
-We use labels like `design`, `documentation`, `test`, `code`, `infrastructure`,
-and `security` to indicate what kind of work is remaining on the issue.
+New contributors may want to review
+[the Zinc contribution guide](https://www.zinc.coop/contributing/) to ground themselves in the norms we try to follow across projects.
```
daltonrpruitt, line 14: I would argue the information in the contribution guide is not as specific as that which is being removed here. \
That said, both are meant to be somewhat general, and the reality is probably a bit more fluid than this doc could provide?

CONTRIBUTING.md
daltonrpruitt: I am of the mindset that as long as that change is understood to exist, then this should be good to go in its present state.

CONTRIBUTING.md
zspencer: I think there's probably some stuff worth keeping that I got rid of (for example the "We don't expect contributors to complete everything for every patch; instead we encourage people to submit pull requests early and ask for guidance and help when they need it."

daltonrpruitt (approved):

PR #1545, opened 2023-06-08
🛠️ `Neighborhood`: Enable bash autocompletion for git in devcontainer

daltonrpruitt:
Based on https://stackoverflow.com/questions/12399002/how-to-configure-git-bash-command-line-completion


zspencer (approved):
omg this rawks

Issue #1449, opened 2023-05-07
🌫️ `Neighborhood`: Provisioning a `Neighborhood`


zspencer:
- https://github.com/zinc-collective/convene/issues/892

Convene Neighborhoods can run on low-cost PaaS, like [Heroku](https://www.heroku.com/home) or [Render](https://render.com/); on Virtual Machines provided by [AWS](https://aws.amazon.com/), [DigitalOcean](https://www.digitalocean.com/), [Linode](https://www.linode.com/), or [Vultr](https://www.vultr.com/).

Ideally, a Convene Neighborhood with one or two low-traffic spaces would cost ~$5/mo. Higher-compute/bandwidth [Neighborhood Hosted Utilities](https://github.com/zinc-collective/convene/issues/891) would drive costs up.

## Use Cases

- [ ] On Heroku
  - [ ] Using Git
  - [x] [Using GitHub](https://github.com/zinc-collective/convene/blob/main/docs/deployment.md#GitHub)
  - [ ] Using Images
- [ ] On Ubuntu 22.04
  - [ ] Using Abra
  - [ ] Manually?
- [ ] On Render
- [ ] On Kubernetes
- [ ] On Docker Swarm?
- [x] https://github.com/zinc-collective/convene/issues/1089

zspencer:
@daltonrpruitt - For Heroku, we should be able to start out encourage people to fork and then follow  [the guide to deploying a Rails app to heroku](https://devcenter.heroku.com/articles/getting-started-with-rails7#deploy-the-application-to-heroku). We may also want to read/encourage operators to read the [guide to using GIT](https://devcenter.heroku.com/categories/deploying-with-git)

We can work towards a Docker Image for Heroku after that; especially since that will help us get standard Kubernetes/Docker Swarm

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1546, opened 2023-06-08
📝  `Operators` : Add steps/discussion for GitHub->Heroku deployment

daltonrpruitt:
First steps to having actual deployment documentation. 
- #1449

I was not able to go through this process exactly due to not wanting to set up my personal credit card yet (since I can't link to an actual fork of convene without having my personal account fully set up), but I did include a link to the contact-us page of zinc for the case that people run into issues.

Feedback welcome, of course!

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -1,13 +1,40 @@
-## Deploying Convene
+# Deploying Convene
 
-A convene Neighborhood will eventually be deployable in several ways.
+A `convene` Neighborhood will eventually be deployable in several ways.
 Currently, we support deploying to:
 - Heroku via:
-  - Git
+  - GitHub
 
 Each of these methods is discussed in detail below:
 
-### Deploying to Heroku
+## Deploying to Heroku
 
-#### Using Git
-TODO: add details...
+### GitHub
+Deploying from GitHub is perhaps arguably the easiest way to deploy to Heroku,
+as it can be done entirely using their GUIs through Heroku-GitHub integration. \
+An overview is discussed below,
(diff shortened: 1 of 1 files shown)


zspencer (approved):
I love it! Thank you so much!

I'm sure we'll learn more as we go; were you able to run through it on the Zinc account? Or did I screw that up when I rebooted the pipeline?

daltonrpruitt:
> I'm sure we'll learn more as we go; were you able to run through it on the Zinc account? Or did I screw that up when I rebooted the pipeline?

I was able to replicate this process mostly in the zinc account, but if I remember correctly, we did run into an issue somewhere when the app was being actually deployed and didn't have time to solve it at that time (so we killed the new pipeline/app to avoid any accidental charges). 
So, this definitely needs to be updated based on whatever is found from a continuing of that, as it could be relevant for all or most potential deployments. 

My reason for making the last statement in that section is that Operators will be deploying from forks in their own accounts, so I would really need to start a new pipeline/app from a fork, but I also haven't done that due to my whole inability/resistance to attach payment information to things... I did try to use a prepaid VISA gift card I had, but it turns out those don't work well with online payments... 🤷 
Anyways, that statement was a catch-all for "no idea what you may see here, but we can help you get through it" (and then document it for posterity).

Ultimately, the document is the Platonic ideal of the deployment process, but I have not verified it fully.

(Man, this comment, which basically was a bunch of excuses, is almost as long as the changes to the actual file. Just thought that was funny 😅 )

zspencer:
> Ultimately, the document is the Platonic ideal of the deployment process, but I have not verified it fully.

I think it's a great start; and we can definitely support folks in chat, the mailing list, or Github Discussions!

Issue #1518, opened 2023-05-27
🛠️ `Neighborhood`: Investigate why emails without an email in the from field didn't throw exceptions


zspencer:
- While investigating https://github.com/zinc-collective/convene/issues/1515, it uncovered a gap in our testing infrastructure. 

Normally, I would expect that emails that can't be delivered would cause something in the test suite to throw a fit. However, it appears `maildev` will happily process emails without any email address in the from field!

It would be lovely if someone could dig into that.

anaulin:
For future reference, here is a code snippet for the Rails console to reproduce sending a malformed email and seeing what error is raised:
```ruby
irb(main):023:0> ActionMailer::Base.mail(to: "ana@zinc.coop", from: "bogus", subject: "test", content_type: "text/html", body: "test").deli
ver_now
D, [2023-06-09T21:56:33.772409 #2] DEBUG -- : ActionMailer::Base#mail: processed outbound mail in 0.7ms
I, [2023-06-09T21:56:34.151286 #2]  INFO -- : Delivered mail 6483a011bcad5_2de832773@9498b1ef-8f87-40b8-865a-13d56d02da86.mail (378.7ms)
D, [2023-06-09T21:56:34.151349 #2] DEBUG -- : Date: Fri, 09 Jun 2023 21:56:33 +0000                                                  
From: bogus                                                                                                                          
To: ana@zinc.coop                                                                                                                    
Message-ID: <6483a011bcad5_2de832773@9498b1ef-8f87-40b8-865a-13d56d02da86.mail>                                                      
Subject: test                                                                                                                        
Mime-Version: 1.0                                                                                                 
Content-Type: text/html;                                                                                          
 charset=UTF-8                                                                                                    
Content-Transfer-Encoding: 7bit                                                                                   
                                                                                                                  
test                                                                                                              
/app/vendor/bundle/ruby/3.2.0/gems/net-smtp-0.3.3/lib/net/smtp.rb:1076:in `check_response': 501 Invalid command or cannot parse from address (Net::SMTPSyntaxError)

```

anaulin:
I think this is resolved for now, and Rails should raise these exceptions in the future.

PR #1549, opened 2023-06-09
🛠️ `Neighborhood`: Raise exceptions on email delivery errors

anaulin:
* https://github.com/zinc-collective/convene/issues/1518

Also tightened up an email spec and unified reading the `EMAIL_DEFAULT_FROM` env varialbe to one location.

I am not 100% certain of the behavior of `raise_delivery_errors`, but if it gets noisy (e.g. if it ends up throwing exceptions for all undeliverable addresses), we can refine this by adding a `rescue_from` to `ApplicationMailer` to report to Sentry only the errors that we can do something about.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,5 +1,7 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: ENV.fetch("EMAIL_DEFAULT_FROM", "from@example.com")
+  DEFAULT_FROM = ENV.fetch("EMAIL_DEFAULT_FROM", "convene-support@example.com")
+
+  default from: DEFAULT_FROM
   prepend_view_path "app/furniture"
 
   layout "mailer"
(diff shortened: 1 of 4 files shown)


app/mailers/application_mailer.rb
```diff
@@ -1,5 +1,7 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: ENV.fetch("EMAIL_DEFAULT_FROM", "from@example.com")
+  DEFAULT_FROM = ENV.fetch("EMAIL_DEFAULT_FROM", "convene-support@example.com")
```
zspencer, line 2: Nice thank you!

config/environments/production.rb
```diff
@@ -66,7 +66,7 @@
 
   # Ignore bad email addresses and do not raise email delivery errors.
   # Set this to true and configure the email server for immediate delivery to raise delivery errors.
-  # config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.raise_delivery_errors = true
```
zspencer, line 69: That explains it!

zspencer (approved):
Thank you this is great!

Issue #1511, opened 2023-05-26
🥔✨ `Marketplace`: Managing `Order::Notification` Email Addresses


zspencer:
- https://github.com/zinc-collective/convene/issues/1331
- https://github.com/zinc-collective/convene/issues/1187

Right now, we require people to put in a comma separated list of email addresses that will receive a notification. This was a reasonable quick hack but we probably want something a little bit nicer.

PR #1561, opened 2023-06-15
🥔✨ `Marketplace`: Sprout model for `OrderNotificationMethod`

zspencer:
- https://github.com/zinc-collective/convene/issues/1511

OK this is awkwardly named, but it is the first tiny step towards allowing us to manage multiple order notification emails without the silly comma separated value.


db/migrate/20230615002110_add_marketplace_order_notification_methods.rb
```diff
@@ -0,0 +1,10 @@
+class AddMarketplaceOrderNotificationMethods < ActiveRecord::Migration[7.0]
+  def change
+    create_table :marketplace_order_notification_methods, id: :uuid do |t|
+      t.references :marketplace, type: :uuid, foreign_key: {to_table: :furnitures}
+      t.string :contact_method, default: :email, null: false
+      t.string :contact_location_ciphertext, null: false
```
KellyAH, line 6: why `contact_location` and not `contact_email`? Does the app expect to have different contact types?

anaulin: I'm guessing Zee is thinking maybe one day this could be e.g. a phone number.

anaulin (approved):
This is fine, although I bet for now we could also get away with not having a new model for this, and using an Array column or similar. (Which we can migrate to a model if/when we expand from emails to phone numbers).

zspencer:
> This is fine, although I bet for now we could also get away with not having a new model for this, and using an Array column or similar. (Which we can migrate to a model if/when we expand from emails to phone numbers).

I considered pulling in `store_model` since you're right; it doesn't need to be it's own gosh-darn table, but I was a coward and wanted to follow the "no-thinking way to add a thing" path to see how it flows.

db/migrate/20230615002110_add_marketplace_order_notification_methods.rb
zspencer: At some point, I would love for us to have a `ContactMethod` within the higher-level Convene Operating System to support this kind of thing; so that each `Gizmo` or other use-case doesn't have to re-invent it every time.

However, I haven't taken the time to actually think more about it or implement it... But basically I imagine there being several fields:

- `#via` (Is this XMPP? Email? SMS? Signal? Voice Call? Physical Mail? WhatsApp? Twitter? ActivityPub?)
- `#location` (the actual address/phone number/etc.)
- `#validated_by` (which `Person` validated the `ContactMethod`)
- `#validated_on` (when the `Person` validated the `ContactMethod`)

Theoretically, this would get squeezed out of the `AuthenticationMethod` class; so that a `Person` can have many `ContactMethod`s with *some* of those `ContactMethod`s being appropriate for authenticating that the person is who they say they are.

anaulin:
Yeah, I think it's NBD either way.

PR #1562, opened 2023-06-15
🌸✨ `Marketplace`: UI to manage `Order::NotificationMethod`

zspencer:
- https://github.com/zinc-collective/convene/issues/1511

This is basically a copy-paste from `Marketplace::DeliveryArea`, so it is *ripe* for testing via the `delivery_area_request_spec.rb`

### TODO:
- [x] Find a better SVG for the `:letter` icon
- [x] Write request specs for
  - [x] #new
  - [x] #create
  - [x] #edit
  - [x] #update
  - [x] #destroy
- [ ] ~Write a Component spec~
- [ ] ~Write a system spec to add, update, and then destroy an `OrderNotificationMethod`~

## After
Screenshot 2023-06-28 at 4 46 36 PM
Screenshot 2023-06-28 at 4 46 32 PM
Screenshot 2023-06-28 at 4 46 27 PM
Screenshot 2023-06-28 at 4 46 23 PM


app/components/svg_component.rb
```diff
@@ -6,6 +6,7 @@ class SvgComponent < ApplicationComponent
     gear: :gear,
     map: :map,
     receipt_percent: :receipt_percent,
+    letter: :receipt_percent,
```
zspencer, line 9: This should probably be a better SVG than the receipt one...

app/furniture/marketplace/breadcrumbs.rb
```diff
@@ -27,6 +27,23 @@
...
+  parent :marketplace_order_notification_methods, order_notification_method.marketplace
+  link t("marketplace.order_notification_methods.index.link_to"),
+    order_notification_method.marketplace.location(child: :order_notification_methods)
+end
+
+crumb :edit_marketplace_order_notification_method do |order_notification_method|
+  parent :marketplace_order_notification_methods, order_notification_method.marketplace
+  link t("marketplace.order_notification_methods.edit.link_to", contact_location: order_notification_method.contact_location)
+  order_notification_method.marketplace.location(child: :order_notification_methods)
+end
```
zspencer, line 45: One day we will not have to define crumbs because we have made them so consistent that we can just infer wht the crumb should be...

ONE DAY....

app/furniture/marketplace/marketplaces/edit.html.erb
```diff
@@ -1,7 +1,6 @@
 <%- breadcrumb :edit_marketplace, marketplace %>
 
 <%= render Marketplace::ManagementComponent.new(marketplace: marketplace) do %>
-  <%= render "form", marketplace: marketplace %>
```
zspencer, line 4: I think it's safe to just delete this form now, since the Marketplace has no attributes on it that can be set by the user any more.

app/furniture/marketplace/order_notification_method_component.rb
```diff
@@ -0,0 +1,36 @@
...
+    def initialize(order_notification_method:, **kwargs)
+      super(**kwargs)
+
+      self.order_notification_method = order_notification_method
+    end
+
+    def edit_button
+      super(title: t("marketplace.order_notification_methods.edit.link_to", contact_location: contact_location),
+            href: order_notification_method.location(:edit))
+    end
```
zspencer, line 15: It would be nice if `#edit_button`, `#edit_button?`, `#destroy_button`, and `#destroy_button?` could be part of a `RecordComponent` or something so we don't have to define them every time.

app/furniture/marketplace/order_notification_methods_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+          if order_notification_method.destroyed?
+            render turbo_stream: turbo_stream.remove(order_notification_method)
+          else
+            render turbo_stream: turbo_stream.replace(order_notification_method)
+          end
+        end
+      end
+    end
+
+    helper_method delegate :order_notification_methods, to: :marketplace
```
zspencer, line 41: This should be `policy_scope`d but I forgot tooooooo

app/components/svg_component.rb
KellyAH: the `ICON_MAPPINGS` was named `letter` so I replaced it with letter icon in 85514fdc1b22a7d4e3713f052521fa2be6e7d88d

letter icon

But, since the icon is applied to a notification menu, I felt a bell icon worked better, thus I added it in a3e90d9337cb107af31120c16c4064fc633452e9. 
bell icon


Please choose which on you want or suggest other icon suggestions.

KellyAH: Oh I see now, it's notification via email. Hm maybe the open_letter icon works better than.

Maybe we should also change the `Order Notifications` text to a clearer `Order Email Notifications` 🤔

zspencer: Well, it's notifications via email... _for now_; but if you check the [`Order` `Notifications` and History](https://github.com/zinc-collective/convene/issues/1331) we do have plans to include at the very least Square Point of Sale; and maybe even SMS!

zspencer: Also, I love the bell!

zspencer:
I'm going to split out the addition of the `OrderNotification` page from the creation of the OrderNotification model; since that is a small enough concrete step.

app/furniture/marketplace/order/notification_methods_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+        @order_notification_method ||= if params[:id]
+          order_notification_methods.find(params[:id])
+        elsif params[:order_notification_method]
+          order_notification_methods.new(order_notification_method_params)
+        else
+          order_notification_methods.new
+        end.tap do |order_notification_method|
+          authorize(order_notification_method)
+        end
+      end
```
zspencer, line 60: I feel like this could be replaced across almost all our controllers with [`decent_exposure`](https://github.com/hashrocket/decent_exposure)

zspencer:
In an ideal world; we would not need nearly this much ceremony to sprout the CRUD UI for a new model. We would need to define the `Model`, `ModelController`, `ModelPolicy`, `ModelPolicy::Scope`, `ModelComponent` and `Model::FormComponent` and the appropriate `locales` and `routes`; but `breadcrumbs` and most, if not all, of the `views` should be pretty standardized.

I'm not sure if it's prudent for me to do that now-right-now; but I am very tempted to take a refactor pass because basic CRUD should not require 300 lines of code :-p.

anaulin (approved):
Thank you!

app/furniture/marketplace/breadcrumbs.rb
```diff
@@ -27,6 +27,23 @@
...
 end
 
+crumb :marketplace_order_notification_methods do |marketplace|
+  parent :edit_marketplace, marketplace
+  link t("marketplace.order.notification_methods.index.link_to"), marketplace.location(child: :order_notification_methods)
+end
+
+crumb :new_marketplace_order_notification_method do |order_notification_method|
+  parent :marketplace_order_notification_methods, order_notification_method.marketplace
+  link t("marketplace.order.notification_methods.index.link_to"),
```
KellyAH, line 37: Isn't this supposed to be?
```suggestion
  link t("marketplace.order.notification_methods.new.link_to"),
```

app/furniture/marketplace/order/notification_methods_controller.rb
rosschapman: 🆒 TIL

db/migrate/20230615002110_add_marketplace_notification_methods.rb
```diff
@@ -1,6 +1,6 @@
-class AddMarketplaceOrderNotificationMethods < ActiveRecord::Migration[7.0]
+class AddMarketplaceNotificationMethods < ActiveRecord::Migration[7.0]
   def change
-    create_table :marketplace_order_notification_methods, id: :uuid do |t|
+    create_table :marketplace_notification_methods, id: :uuid do |t|
```
zspencer, line 3: 🤦 I did a silly

Issue #1567, opened 2023-06-15
🌸✨ `Journal`: The "Add Journal Entry" link takes a long time to scroll to


zspencer:
- https://github.com/zinc-collective/convene/issues/1566

When signed in on https://beta.zeespencer.com, I noticed that the "Add Journal Entry" link takes a while to scroll to. Maybe we add one to the top as well? Or do something else to make it  a bit easier to access?

zspencer:
@KellyAH or @daltonrpruitt - Is this something either of you want to tackle? If so, just assign yourself to it! Otherwise I will maybe pick it up when I want to build something that is "Not Actually Important, But Still Kinda Meaningful (At Least To Me)"

KellyAH:
Adding a "Add Journal Entry" button to the top of the page sounds good. I'll give it a try.

KellyAH:
I will start this tomorrow 6/20. I need to update my resume w/ rails work.

zspencer:
Thank you! This is much nicer!

PR #1583, opened 2023-06-21
`Journal`: Add `New Entry` button to the top of the `Journal` view

KellyAH:
- For https://github.com/zinc-collective/convene/issues/1567

- add new journal entry button to the top of the page
change

### DIFF
--- a/app/furniture/journal/journals/_journal.html.erb
+++ b/app/furniture/journal/journals/_journal.html.erb
@@ -1,4 +1,5 @@
 <section class="max-w-3xl mx-auto">
+  <%- new_entry = journal.entries.new %>
   <%- if policy(new_entry).new? %>
     <div class="text-center p-2">
       <%= link_to "New Journal Entry", journal.location(:new, child: :entry) %>
@@ -13,7 +14,6 @@
 
   <%== pagy_nav(@pagy, nav_extra: 'flex justify-between') %>
 
-  <%- new_entry = journal.entries.new %>
   <%- if policy(new_entry).new? %>
     <footer class="text-center p-2">
       <%= link_to "New Journal Entry", journal.location(:new, child: :entry) %>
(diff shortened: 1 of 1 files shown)


app/furniture/journal/journals/_journal.html.erb
```diff
@@ -1,4 +1,10 @@
 <section class="max-w-3xl mx-auto">
+  <%- if policy(new_entry).new? %>
+    <div class="text-center p-2">
+      <%= link_to "New Journal Entry", journal.location(:new, child: :entry) %>
+    </div>
+  <%- end %>
+
```
KellyAH, line 8: I had to code this blind. 😓 

I can't get to the page for creating new journal entries on my local. I added a Journal Gizmo to a room but can't figure out where to go to add an entry or look at journal entry lists etc. 

Is our UI that bad or did I forget to set up something?
Screenshot 2023-06-20 at 7 45 40 PM

I keep clicking the room name of the room I added the Journal gizmo to, but nothing happens. 
Screenshot 2023-06-20 at 7 46 13 PM

And I tried manually crafting the correct URL path based off the route syntax. No dice 😞 

`/spaces/:space_id/rooms/:room_id/journals/:journal_id/entries/new(.:format)`


`http://localhost:3000/spaces/test/rooms/food-room/journals/1/entries/new`

zspencer: I wonder what would happen if you put a bit of content inside of the `if` and outside the `if` in order to identify whether the `if policy(entry).new?` was why it was being hidden or if it was a different issue.

My guess is that the `Journal::EntryPolicy` expects the `current_person`  to be a `member?`  of the `current_space`; and it doesn't care that you are an `operator?`.

If that does seem like the case, LMK and I'll create an enhancement issue (or you can create one?) so we don't lose track!

zspencer (approved):
This looks great!

Issue #1565, opened 2023-06-15
🐞 Clicking  "Remove Gizmo 🗑️" button from a `Gizmo`'s edit page does not actually remove the `Gizmo` from the `Section`


zspencer:

zspencer:
@KellyAH  or @daltonrpruitt  - Are either of you willing to write up some repro steps for this (or a test that isolates it?)

KellyAH:
I have 5 assigned issues right now. I don't have anymore bandwidth for this one. Can you take it @daltonrpruitt ? Thanks. 🙏 🙇‍♀️

daltonrpruitt:
Sure thing. Will try do that in a couple hours ish.

daltonrpruitt:
Steps to reproduce (can be put above if needed):
1. In a running instance of `convene` (the base dev instance for now), sign in as "space-member" via regular steps
2. Go into "System Test"
3. Click on either edit/settings/configure icon ⚙️ 
4. Under "Sections", click on "Listed Room 1" 
5. Under "Gizmos", click the edit icon ⚙️ next to "Markdown Text Block"
6. Click "Remove Gizmo 🗑️"

Expected Behavior: Return to Edit screen for "Listed Room 1" with "Markdown Text Block" gone (so no Gizmos in this case)
Actual Behavior: Returned to Edit screen for "Listed Room 1" with "Markdown Text Block" still present

Steps 5-6 can be repeated any number of times for a Gizmo with the same results. 
This also applies to the other Gizmos that have "Remove Gizmo 🗑️" as an option in their configure page: "Livestream" and "Embedded Form"

zspencer:
These were great! I added some typographical flourishes and put them in the main issue body so it's easier to find without scrolling.

zspencer:
I'm also going to assign this to myself so I know what to pick up next! If I get to it before next wednesday, it means Convene will have a system test that @KellyAH can crib off of for https://github.com/zinc-collective/convene/pull/1562!

zspencer:
@daltonrpruitt - Can you confirm that this is fixed in production when you get a chance and then close this out?

daltonrpruitt:
I have tried to go through this a few times today, but I keep running into different issues when I try to get to the configure/edit page of my space.
The latest one occurs in a fresh instance when I try to go to the configuration/edit page (clicking the gear) for a room I add. The error is shown below. I think it has something to do with some of the changes we were looking at today? I can try to look at it more later. 
[...]

Issue #1472, opened 2023-05-18
🧹 Rename from `Room` to `Section` and  `Furniture` to `Gizmo`


zspencer:
- https://github.com/zinc-collective/convene/issues/1155
- https://github.com/zinc-collective/convene/issues/709

During one of our demos with Piikup, we noticed there was a degree of struggle with the "Room" and "Furniture" nomenclature. @anaulin brought that to the ensembles, and we talked a bit about shifting away from "Room" (which is a holdover from when we were focusing on the `Videobridge` Furniture.

- [x] Replace "Room" and "Rooms" in the `views` and `locales` files with "Section" and "Sections" https://github.com/zinc-collective/convene/pull/1474
- [x] Replace "Furniture" in the `views` and `locales` files with "Gizmo" and "Gizmos" https://github.com/zinc-collective/convene/pull/1476
- [x] Sleep on it! (Or Demo it?)
- [x] Decide if we want to keep the change (We do, it's "Better than Furniture", per @KellyAH)
- [x] Update the [`Furniture` Github Issue Label](https://github.com/zinc-collective/convene/labels/🗜%EF%B8%8F%20Gizmos) to `Gizmo` 
- [ ] Rename the open `Furniture` Issues
- [ ] Rename the open `Room` issues
- [x] Update any of our Docs https://github.com/zinc-collective/convene/pull/1571
- [ ] Rename the Models and Tables

zspencer:
I've gone ahead and updated the issue title for all the `Furniture` and renamed the `Furniture` label to `Gizmo`

KellyAH:
@zspencer @anaulin Are we happy with the renames of  "Section" "Gizmo"/"Gizmos"? 

Should I proceed with the remaining refactor tasks to  Update Docs and Rename the Models and Tables to match the rename? Or wait longer till we get more feedback/info?

anaulin:
I'd wait a little longer on the models and tables (like, a few months at least?). It would be nice to drop a note in the docs about how we are calling them differently in the UI.

KellyAH:
> It would be nice to drop a note in the docs about how we are calling them differently in the UI.

These docs still mentions `Room/Rooms`:
* https://github.com/zinc-collective/convene/blob/main/features/support/parameter-types/README.md links to https://github.com/zinc-collective/convene/blob/main/features/support/parameter-types/rooms.js which doesn't exist anymore.
[...]

PR #1588, opened 2023-06-22
🥗 `Gizmo`: System Spec to Add and Remove Gizmo

zspencer:
- https://github.com/zinc-collective/convene/issues/1565
- https://github.com/zinc-collective/convene/issues/1472

For some reason the automated check doesn't *FAIL* despite it failing when we manually execute it!

The "cause" is that the gizmo form has two `_method` fields, one is the hidden `_method` field that is automatically inserted by `form_with` when the `model` is `persisted?`; the other is added by a `button` that sets the `_method` to delete so that it can be removed.

In the system spec; the driven `firefox` takes the `_method` set by the clicked `button`. In the user-land; it sends *both* methods; and Rails picks the first one.

The "solution" will be to ditch the second `_method` (possibly even removing the ability to Remove a Gizmo from the `update` form.)

### DIFF
--- a/app/assets/stylesheets/components/button.scss
+++ b/app/assets/stylesheets/components/button.scss
@@ -1,10 +1,11 @@
+/* @todo DELETE ME */
 @layer components {
   [type="reset"],
   [type="button"],
   [type="submit"],
   button,
   .button {
-    @apply font-bold py-2 px-4 my-1 rounded bg-primary-500 text-white text-center transition ease-in-out duration-150;
+    @apply font-bold py-2 px-4 rounded bg-primary-500 text-white text-center transition ease-in-out duration-150;
 
     &:hover {
       @apply bg-primary-700;
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -36,6 +36,7 @@ class Room < ApplicationRecord
 
   has_many :furnitures, dependent: :destroy, inverse_of: :room
   accepts_nested_attributes_for :furnitures
+  alias_method :gizmos, :furnitures
 
   def full_slug
     "#{space.slug}--#{slug}"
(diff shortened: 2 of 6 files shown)


bin/setup
```diff
@@ -48,6 +48,13 @@ else
   cp .env.example .env
 fi
 
+if test -f ".env.development"; then
+  echo "Found .env.development, leaving it in place!"
+else
+  echo "Copying .env.development.example to .env.development"
+  cp .env.development.example .env
+fi
+
```
daltonrpruitt, line 57: I have a feeling the `.devcontainer/startup.bash` script may need to be updated with this new `.env` changes?
Specifically, the block starting at [line 92](https://github.com/zinc-collective/convene/blob/51ff6f302ac669370e4636e8f3a6baa22a9c13fe/.devcontainer/startup.bash#L92C3-L92) may need to be adjusted.
This assumption may be wrong, but I wanted to bring it up sooner rather than later.

zspencer: It would be nice if the three places we use (bin/setup, bin/setup-rails, and .devcontainer/startup.bash) all used the same shell script to create these default ffiles; perhaps `bin/setup-env`? Are you available to do that refactor?

.env.example
```diff
@@ -10,11 +10,11 @@
...
-SMTP_PORT=1025
-SMTP_DOMAIN=localhost
-SMTP_AUTHENTICATION=plain
-SMTP_ADDRESS=localhost
-SMTP_ENABLE_TLS=false
+# SMTP_PORT=1025
+# SMTP_DOMAIN=localhost
+# SMTP_AUTHENTICATION=plain
+# SMTP_ADDRESS=localhost
+# SMTP_ENABLE_TLS=false
```
KellyAH, line 17: Please break out system test step to it's own PR.

bin/setup
daltonrpruitt: Sure, I can do that.

One script to do a conditionally copy of both `.env.example` and `.env.development.example`?
And then in the `.devcontainer/startup.bash` script, do the `sed` stuff after running this script?
Mostly just probing to make sure I'm understanding what needs to be here.

zspencer: Yep! I think the sed stuff should stay in the startup.bash

.env.example
zspencer: Done! Plz review!

KellyAH (approved):
LGTM

PR #1589, opened 2023-06-22
`Contributor`: override .env.development

zspencer:
I realized that `.env` is used in every environment (`production`, `test` and `development`); which means that the ActionMailer test deliveries ... actually go all the way through to maildev!

Which was not ideal.


anaulin (approved):
Oops. Good catch!

PR #1592, opened 2023-06-25
🐞🔨 `Gizmo`: Fix bug where removing a Gizmo fails

zspencer:
- https://github.com/zinc-collective/convene/issues/1565

OK this is also a bit much; but it does fix the bug. 

Basically, two `_method`s make things explode.


KellyAH (approved):
LGTM

PR #1594, opened 2023-06-25
🧹 `Neighborhood`: Get rid of unused `Feature` model

zspencer:

We used this early on to provide a degree of feature flagging; but we haven't used it lately and we will probably want to rely on `flippa` or something more fully featured and better maintained as we mature.


daltonrpruitt (approved):
Seems pretty straightforward.

Issue #1513, opened 2023-05-27
🌱✨ Setting the `Space` Domain


zspencer:
- https://github.com/zinc-collective/convene/issues/74

In [Space] -> [Edit] -> [Website Settings], `Members` may specify the `Domain` name their `Space` is reachable at on the public web.

## Use Cases
- [x] Show the `Space`'s `Domain` https://github.com/zinc-collective/convene/pull/1600
- [x] Setting the `Space` Domain Name https://github.com/zinc-collective/convene/pull/1595
- [ ] Enforcing Access through the `Domain`

zspencer:
@KellyAH - This may be a good one to tackle soon; it's already a bit work-in-progress; in that Operators can set `Space#branded_domain` via the rails console; however it would be better if:

- Any `Space` `Member` could set the domain via the UI 
- We could toggle whether accessing the `Space` must be through the `Domain` or not.

When you're ready, I can give you a few places to look and a better breakdown of the technical next steps; or (as always) feel free to explore on your own if that suits your training style better.

KellyAH:
I will start working on this tomorrow during our wed session. 6/21

anaulin:
It would be nice to start by at least _showing_ the branded domain for the Space in the configuration screen (if there is a branded domain).

rosschapman:
🧹  Relegated last task to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1595, opened 2023-06-25
🥔✨ `Space`: `Member` may set `Domain`

KellyAH:
- For https://github.com/zinc-collective/convene/issues/1513

After setting a domain on Space configure page, the Space#update will redirect to that new domain and encounter the issue below. Thus users cannot load the Space configure page to see the domain setting on their space.

e.g. in below example, space.branded_domain = "games.local"

Screenshot 2023-06-25 at 11 45 48 AM

https://github.com/zinc-collective/convene/assets/50284/9945c53e-24cb-4331-a8a2-45ec13569584

### DIFF
--- a/app/controllers/spaces_controller.rb
+++ b/app/controllers/spaces_controller.rb
@@ -42,7 +42,7 @@ class SpacesController < ApplicationController
   def update
     if space.update(space_params)
       flash[:notice] = t(".success")
-      redirect_to space
+      redirect_to space, allow_other_host: true
     else
       flash.now[:alert] = t(".error")
       render :edit
--- a/app/controllers/spaces_controller.rb
+++ b/app/controllers/spaces_controller.rb
@@ -42,7 +42,7 @@ class SpacesController < ApplicationController
   def update
     if space.update(space_params)
       flash[:notice] = t(".success")
-      redirect_to space, allow_other_host: true
+      redirect_to space.location(:edit), allow_other_host: true
     else
       flash.now[:alert] = t(".error")
       render :edit
--- a/spec/requests/spaces_controller_request_spec.rb
+++ b/spec/requests/spaces_controller_request_spec.rb
@@ -126,10 +126,14 @@ RSpec.describe SpacesController do
 
       it "updates the Space" do
         new_entrance = space.rooms.sample
-        put polymorphic_path(space), params: {space: {entrance_id: new_entrance.id, enforce_ssl: 1}}
+        put polymorphic_path(space), params: {space: {entrance_id: new_entrance.id, enforce_ssl: true, branded_domain: "zeespencer.com"}}
 
-        expect(space.reload.entrance).to eql(new_entrance)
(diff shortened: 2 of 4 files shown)


zspencer:
TBH, the cheap-and-easy thing would be to add the `allow_other_hosts: true` on the redirect in the `SpacesController#update` action. It's a little weird, that's for sure; but I think it's probably fine.

app/controllers/spaces_controller.rb
```diff
@@ -42,7 +42,7 @@ def create
   def update
     if space.update(space_params)
```
KellyAH, line 43: `space.update` is failing and doing a roll back. But I do not see any error indicating why. 

Screenshot 2023-06-28 at 8 27 52 PM

anaulin: You wouldn't see the errors in the server log this way. What you want to do is inspect `space.errors` after the failed update in some way. The typical Rails way is that you have an `else` branch that you use to display the errors when the update is not successful.

zspencer: Ideally, the form would also show any validation errors underneath of the field (you can see how to do this by looking at  [`app/views/application/_text_field.html.erb`](https://github.com/zinc-collective/convene/blob/main/app/views/application/_text_field.html.erb), which shows how to use the `_error` partial for this.

zspencer:
@KellyAH - I pushed this over the finish line per your request in Discord. I'll let you decide if you would rather someone other than me from the @zinc-collective/convene-contributors or @zinc-collective/convene-maintainers should do the review!

app/controllers/spaces_controller.rb
```diff
@@ -42,7 +42,7 @@ def create
...
       flash[:notice] = t(".success")
-      redirect_to space
+      redirect_to space.location(:edit), allow_other_host: true
```
zspencer, line 45: Since the `Space#edit` page is so big; it felt weird to be sending them back to the `Space#show` page on change. Also we didn't have any tests about where it was supposed to go which implied to me that it was left unspecified.

spec/requests/spaces_controller_request_spec.rb
```diff
@@ -126,10 +126,14 @@
 
       it "updates the Space" do
         new_entrance = space.rooms.sample
-        put polymorphic_path(space), params: {space: {entrance_id: new_entrance.id, enforce_ssl: 1}}
+        put polymorphic_path(space), params: {space: {entrance_id: new_entrance.id, enforce_ssl: true, branded_domain: "zeespencer.com"}}
```
zspencer, line 129: `true` felt more natural here than `1`

zspencer (approved):

KellyAH:
Zee's changes LGTM! @anaulin do you think this is okay for me to merge?

anaulin:
@KellyAH yes, go ahead!

In general, if someone already approved it, as far as I'm concerned there is no need to double-check. 🚢

PR #1597, opened 2023-06-25
🥗 `Neighborhood`: We can write system specs in Ruby!

zspencer:

Extracted from https://github.com/zinc-collective/convene/pull/1588


KellyAH (approved):
LGTM

PR #1598, opened 2023-06-25
🧹 `Gizmo`: Access a `Room`'s Gizmos via `.gizmos`

zspencer:
- https://github.com/zinc-collective/convene/issues/1472

Theoretically, this will get us a bit closer to Gizmofication...

I'm available to do a big-bang rename; but I figured starting with replacing where we call `.furniture` and `.furnitures` with `.gizmo` and `.gizmos` would be a smaller step; but it also is a bit scatter-shot.

I *think* it should be possible to rename the `Furniture` class (and it's related `Controller`, `Policy`, and `Views`) but I got skeered.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -139,7 +139,7 @@ class ApplicationController < ActionController::Base
   end
 
   def space_repository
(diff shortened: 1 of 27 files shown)


KellyAH (approved):
LGTM. thanks for making this a smaller refactor. it is easier for me to review.

PR #1600, opened 2023-06-25
✨ `Space` config displays branded domain

KellyAH:
- display the branded domain for a Space on the configuration screen (if there is a branded domain).
- [ ] ~~unit test when space.branded_domain=nil (redirect issue prevents us from creating unit test for when domain is set https://github.com/zinc-collective/convene/pull/1595)~~

**Domain exists**
domain exists

**Domain is nil**
nil domain


app/views/spaces/_website_settings.html.erb
```diff
@@ -4,6 +4,13 @@
     <p class="text-sm italic"><%= t('website_settings.help_text') %></p>
   </header>
 
+  <h4 class="text-base"> Domain </h4>
+  <% if space.branded_domain.present? %>
+    <p class="text-sm"><%= space.branded_domain %></p>
```
zspencer, line 9: You could put the `if` inside the `p` to reduce divergence between the branches. Also, `space.branded_domain.presence || "No Domain"` will evaluate to "No Domain" when `space.branded_domain` is `nil` or `#empty?` and the value held in `space.branded_domain` otherwise.

Which means you could do this!

```erb
<p class="text-sm"><%= space.branded_domain.presence || "No Domain"%></p>

```

zspencer (approved):
Looks good! to test this you will probably want to add an expectation to the `SpacesController#show` tests in `spec/requests/spaces_controller_request_spec`

*or*

You could make a `Space::WebsiteSettingsComponent` which would be testable outside the request/response cycle (and therefore _slightly_ easier)

KellyAH:
gah I still need to add 1 test. I'll do it tomorrow. it's getting late and https://github.com/zinc-collective/convene/pull/1602 ate the rest of my 🧠 ⚡ .

app/views/spaces/_website_settings.html.erb
KellyAH: fixed https://github.com/zinc-collective/convene/pull/1600/commits/c5b6ce041bd90d69444aa073276de51e14ab7b44

app/views/spaces/_website_settings.html.erb
```diff
@@ -4,6 +4,13 @@
     <p class="text-sm italic"><%= t('website_settings.help_text') %></p>
   </header>
 
+  <h4 class="text-base"> Domain </h4>
+  <% if space.branded_domain.present? %>
+    <p class="text-sm"><%=  space.branded_domain.presence || "No Domain" %></p>
+  <% else %>
+    <p class="text-sm"> No domain </p>
```
zspencer, line 11: Do we need the `if`/`else` anymore?

zspencer (approved):

PR #1601, opened 2023-06-25
🧹 `Neighborhood`: update CI pipeline names

KellyAH:
Update test pipeline names to add clarity.

- change `Run tests` ➡️  `Run Rspec tests` because the pipe runs Rspec Unit tests and now, the new System test `spec/system/furniture_spec.rb` (added in https://github.com/zinc-collective/convene/pull/1588) which also includes some browser testing.

- change `Run browser tests`  ➡️ `Run Cucumber tests` because the pipe not no longer the only pipe that runs browser tests.

chagnes 1
chagnes 2

### DIFF
--- a/.github/workflows/test-convene-web.yml
+++ b/.github/workflows/test-convene-web.yml
@@ -53,7 +53,7 @@ jobs:
         run: yarn install
 
   test-rspec:
-    name: Run tests
+    name: Run Rspec tests
     runs-on: ubuntu-latest
     needs: [setup]
 
@@ -119,7 +119,7 @@ jobs:
       - name: Setup rails
         run: bin/setup-rails && bin/rails assets:precompile
 
-      - name: Run Tests
+      - name: Run tests
         env:
           HEADLESS: true
         run: bundle exec rspec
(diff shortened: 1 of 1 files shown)


.github/workflows/test-convene-web.yml
```diff
@@ -131,7 +131,7 @@ jobs:
           path: tmp/capybara/*.png
 
   test-features:
-    name: Run browser tests
+    name: Run Cucumber tests
```
KellyAH, line 134: `Run Cucumber feature tests` seemed like a really long name thus I shortened it. But I can add it back if people prefer it.

.github/workflows/test-convene-web.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         run: yarn install
 
   test-rspec:
-    name: Run tests
+    name: Run Rspec tests
```
KellyAH, line 56: Does this pipeline rename need to be added in another place?

oops

zspencer:
I believe it's because Github Actions requires us to update the `required` set with the new names. If you mark this as ready for review I will update the settings and merge

zspencer (approved):
Ayyyy! I updated the thing and it says it's good to go; so cnsider this ur :seal: of :approval:

PR #1593, opened 2023-06-25
✨`Space`: Prompt to add `Gizmo` to  `Entrance` 

KellyAH:
- for https://github.com/zinc-collective/convene/issues/1585
in ref for this [comment](https://github.com/zinc-collective/convene/issues/1585#issuecomment-1605286548) 

Screenshot 2023-06-25 at 11 05 43 AM

Issue #1585, opened 2023-06-22
🥔✨ `Spaces`: New Space Onboarding!


KellyAH:
- Once https://github.com/zinc-collective/convene/issues/1154 are registered, it would be nice if the https://github.com/zinc-collective/convene/issues/117 were guided through the process of tailoring it to their needs.

## Use Cases
- [x] 🥔✨ Prompt to add `Member`s 
- [x] 🥔✨ Prompt to add `Section`
- [x] 🥔✨ Prompt to set `Entrance`
- [x] 🥔✨ Prompt to set `Gizmo` for entrance missing  a `Gizmo` https://github.com/zinc-collective/convene/pull/1593
- [ ] 🥔✨ Better handle when a `Space` does not have an `Entrance` (or make it impossible to get into a state where the `Space` does not have an `Entrance` if it has rooms)

zspencer:
@KellyAH what do you think about adding a prompt when the current room doesn’t have any gizmos?

KellyAH:
> @KellyAH what do you think about adding a prompt when the current room doesn’t have any gizmos?

I'll add that in a new PR

KellyAH:
> Automatically presume the only Section is the Entrance
What does this mean exactly? There can be multiple sections. What am I supposed to build?

zspencer:
> > Automatically presume the only Section is the Entrance
> 
> What does this mean exactly? There can be multiple sections. What am I supposed to build?


The problem I am trying to solve is there is a blank screen when a `Space` has a `Section` but `Space#entrance` is not set. I think the way to resolve that could go down several pathways: 

- We could adjust what happens when a `Section` is created in a `Space` with no `Section`s yet to also set the new `Section` as the `Space#entrance`
- We could override `Space#entrance` (perhaps implementing it as `super.presence || sections.first` or something); so that if there are any `Section`s in the `Space` it at least has a quasi-entrance.
- ???

You will likely need to explore both pathways a little bit (or come up with your own 3rd pathway) before making the call about which is easiest for you to make happen. My instincts are overloading the `Space#entrance` method is easiest.

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #1602, opened 2023-06-26
🐞🔨 `Space`: When the `entrance` is missing, it explodes

KellyAH:
- https://github.com/zinc-collective/convene/issues/1585

Trying to fix an issue @daltonrpruitt  found in https://github.com/zinc-collective/convene/issues/1565#issuecomment-1606496774 and add a missing condition to an if statement that got forgotten when https://github.com/zinc-collective/convene/pull/1593 was built

**prompts that display, if no section, no entrance and no entrance gizmos exist**
prompts that display when no section nor entrance nor gizmos

**prompts that display, if 1 section exists, but no entrance is set**
when 1 section:room exists

**prompts that display, if 1 section exists, it is set as entrance, but entrance has no gizmo**
when 1 section exist and is set as entrance but no gizmos

**prompts that display, if section entrance has gizmo**
when section entrance has gizmo

### DIFF
--- a/app/models/space.rb
+++ b/app/models/space.rb
@@ -49,5 +49,9 @@ class Space < ApplicationRecord
     []
   end
 
+  def configured?(current_space)
+    current_space.rooms.blank? || current_space.entrance_id.blank? || current_space.members.blank? || current_space.entrance&.furnitures.blank?
+  end
+
   attr_accessor :blueprint
 end
--- a/app/views/spaces/edit.html.erb
+++ b/app/views/spaces/edit.html.erb
@@ -1,6 +1,6 @@
 <% breadcrumb :edit_space, space %>
 
-<% if space.entrance_id.blank? || space.rooms.blank? || space.members.blank? %>
+<% if space.rooms.blank? || space.entrance_id.blank? || space.members.blank? || space.entrance.furnitures.blank? %>
   <%= render AlertComponent.new(scheme: :warning, icon: :exclamation_triangle,
     title: t('spaces.edit.config_missing_title')) do %>
     <%= t('spaces.edit.config_missing_explainer_html' ) %>
--- a/app/views/spaces/edit.html.erb
+++ b/app/views/spaces/edit.html.erb
@@ -24,7 +24,7 @@
         </li>
       <% end %>
 
-      <% if space.entrance&.furnitures.blank? %>
+      <% if space.entrance_id.present? && space.entrance&.furnitures.blank? %>
         <li>
           <%= t('spaces.edit.config_missing_entrance_furniture_text_html', room_name: link_to(space.entrance.name, space.entrance.location(:edit) )) %>
         </li>
--- a/app/views/spaces/edit.html.erb
+++ b/app/views/spaces/edit.html.erb
@@ -1,6 +1,6 @@
 <% breadcrumb :edit_space, space %>
 
-<% if space.rooms.blank? || space.entrance_id.blank? || space.members.blank? || space.entrance&.furnitures.blank? %>
+<% if space.configured?(space) %>
   <%= render AlertComponent.new(scheme: :warning, icon: :exclamation_triangle,
     title: t('spaces.edit.config_missing_title')) do %>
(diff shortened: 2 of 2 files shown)


KellyAH:
crud my local got messed up somehow. `bin/setup` fails to connect with postgres yet it's up and served on the correct port. 🙀 I tried restarting docker, my computer, etc. I think a config on my local is wrong of or a pid file maybe is failed to be deleted 🤷‍♀️ 

Screenshot 2023-06-25 at 7 52 31 PM

Coded this blind and using code spaces to test it 🤞 🙏

app/views/spaces/edit.html.erb
```diff
@@ -1,6 +1,6 @@
 <% breadcrumb :edit_space, space %>
 
-<% if space.entrance_id.blank? || space.rooms.blank? || space.members.blank? %>
+<% if space.rooms.blank? || space.entrance_id.blank? || space.members.blank? || space.entrance&.furnitures.blank? %>
```
zspencer, line 3: Could we pull this up to `Space#configured?` so that the view only has one method to call?

app/views/spaces/edit.html.erb
```diff
@@ -24,7 +24,7 @@
         </li>
       <% end %>
 
-      <% if space.entrance.furnitures.blank? %>
+      <% if space.entrance&.furnitures.blank? %>
```
zspencer, line 27: I *think* nil respods to `blank?` so it should be fine; but in general once you safe-navigate you'll want to safe-navigate the rest: `space.entrance&.furnitures&.blank?`

zspencer (approved):
Looks good!

app/views/spaces/edit.html.erb
```diff
@@ -24,7 +24,7 @@
         </li>
       <% end %>
 
-      <% if space.entrance.furnitures.blank? %>
+      <% if space.entrance_id.present? && space.entrance&.furnitures.blank? %>
```
KellyAH, line 27: I think I see what you mean Zee when you mentioned https://github.com/zinc-collective/convene/issues/1565#issuecomment-1606518350

app/views/spaces/edit.html.erb
KellyAH: added in 7f31d6aa0264553d84944a07f2f9f7f4beddf8f8 but super janky and at least the page loads 😅 

Tho I added it to Space model. I can't recall if views invoking model methods was a correct rails or not. 🤦 

I tried adding it to the Space controller as a helper method or regular method.... but it broke the page. I'll keep trying. 

I might be overthinking this.... 😅

app/views/spaces/edit.html.erb
KellyAH: ah i forgot. thanks! added c61aa9db7578961cf8c734bd697be56c9f34b4da

daltonrpruitt (approved):
I like this method of organizing the errors (can't set up entrance gizmos if you don't have an entrance!).

app/models/space.rb
```diff
@@ -49,5 +49,9 @@ def parent_location
     []
   end
 
+  def configured?(current_space)
+    current_space.rooms.blank? || current_space.entrance_id.blank? || current_space.members.blank? || current_space.entrance&.furnitures.blank?
+  end
```
KellyAH, line 54: This might need refactoring later. I tried moving `configured?` to the Spaces controller but couldn't get it working. It's too late at night to keep messing with it. 

Maybe we can chat about it wednesday and fix whatever mental model got me stuck when I was working on this.

KellyAH:
merged! I 🙏 it doesn't break space config page.

app/models/space.rb
zspencer: Hmm, I wonder if what we want is a `Space::OnboardingComponent` that can have these methods? Then we could more easily test the whole kit and kaboodle, and we wouldn't be putting responsibility on the `Space` itself.

zspencer (commented):
Thank you!!!!

PR #458, opened 2021-12-13
Remove Webpacker

user512:
This PR intended to remove `webpacker` while keeping webpack for now, with the flexibility to swap to other build tool when we wanted to.

Tasks to complete:

- [x] All tests on CI should pass
- [x] Have someone other than me to smoke test locally/ on review app
- [x] Figure out production build (Review app)
- [x] Have another developer to see their development experience on this new build process
- [x] Ensure node version is above 12.20.0

Below are the nondeterministic tests, sometimes my test all passes locally, sometimes not, the first one seems to always fail on the first run.
```
1) Scenario: Identity Verification Via Emailed Link # features/identification.feature:45
   ✔ Given an unauthenticated Space Member has requested to be identified via Email # features/steps/identification_steps.js:8
   ✖ When the unauthenticated Space Member opens the Identification Verification Link emailed to them # features/steps/identification_steps.js:26
       TypeError: Cannot read properties of undefined (reading 'text')
           at Actor.authenticationUrl (/home/runner/work/convene/convene/features/lib/Actor.js:40:26)
           at processTicksAndRejections (node:internal/process/task_queues:96:5)
   - Then the Space Member is Verified as the Owner of that Email Address # features/steps/identification_steps.js:44
   - And the Space Member has become Authenticated # features/steps/identification_steps.js:54
   ✔ After # features/support/env.js:20
```
```
  @built
  Scenario: Identity Verification via Emailed Code # features/identification.feature:83
    Given an unauthenticated Space Member has requested to be identified via Email
    When the unauthenticated Space Member provides the Identification Code emailed to them
    Then the Space Member is Verified as the Owner of that Email Address
    ✖ failed
      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
      + actual - expected

      + 'guest@example.com'
      - 'space-member@example.com'
          + expected - actual

          -guest@example.com
          +space-member@example.com

          at CustomWorld.<anonymous> (/Users/tomlee/dev_work/zinc/convene/features/steps/identification_steps.js:50:12)
          at processTicksAndRejections (node:internal/process/task_queues:96:5)
    And the Space Member has become Authenticated
    - skipped
```
```
  @built
  Scenario: Locking an Unlocked Room # features/rooms/locking-and-unlocking-rooms.feature:33
    Given a Space with an Unlocked Room
Screenshot created:  features/test_reports/Entering_a_Locked_Room.png
    When a Space Member locks the Unlocked Room with a valid Access Code
this.email space-member@example.com
[...]

PR #1617, opened 2023-06-29
🧹🏡 Remove `cssbundling` and `jsbundling` gems, which we aren't using

anaulin:
Simplifying our Gemfile a bit.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,8 +20,6 @@ gem "rails", "~> 7.0"
 gem "puma", "~> 6.3"
 
 # Browser Layer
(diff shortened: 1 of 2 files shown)


zspencer (approved):
Wooo! I don't remember why we added them but I'm glad we don't seem to need them.

anaulin:
I guess we are actually using these gems, oh well. Doing some spelunking, for future reference, they were added by Tom in this PR where he removed webpacker: https://github.com/zinc-collective/convene/pull/458

PR #1618, opened 2023-06-29
🏠 `Neighborhood`: Consolidate default setup of `.env*` files

daltonrpruitt:
We were performing the same actions in three different places. 
This consolidates those into one single script.
This also fixes a bug in copying the `.env.development.example` file over the `.env` file, which was only being done in the `.devcontainer/startup.bash` script.

Co-authored-by: Ana <anaulin@users.noreply.github.com>


zspencer (approved):
Thank you!!!

PR #1620, opened 2023-06-29
🧹 Eliminate duplication in `Marketplace::ManagementComponent`

zspencer:
-  https://github.com/zinc-collective/convene/issues/831

Now we have a small set of buttons where the vew code only has the *differences*! yay!

### DIFF
--- a/app/furniture/marketplace/management_component.rb
+++ b/app/furniture/marketplace/management_component.rb
@@ -7,5 +7,17 @@ class Marketplace
 
       self.marketplace = marketplace
     end
+
+    def button_to(location, icon:)
+      label, href = if location.is_a?(Symbol)
+        [t("marketplace.marketplace.#{location}.link_to"),
+          marketplace.location(location)]
+      else
+        [t("marketplace.#{location[:child]}.index.link_to"),
+          marketplace.location(**location)]
+      end
+
+      ButtonComponent.new(label: label, icon: icon, href: href, turbo_stream: false, method: :get, scheme: :secondary)
+    end
   end
 end
--- a/app/furniture/marketplace/management_component.rb
+++ b/app/furniture/marketplace/management_component.rb
@@ -8,7 +8,7 @@ class Marketplace
       self.marketplace = marketplace
     end
 
-    def button_to(location, icon:)
+    def button(location, icon:)
       label, href = if location.is_a?(Symbol)
         [t("marketplace.marketplace.#{location}.link_to"),
           marketplace.location(location)]
--- a/app/furniture/marketplace/marketplace_component.html.erb
+++ b/app/furniture/marketplace/marketplace_component.html.erb
@@ -2,7 +2,7 @@
   <%= render AlertComponent.new(scheme: :warning, icon: :exclamation_triangle,
                                 title: t('marketplace.marketplace.config_missing_title')) do %>
     <%= t('marketplace.marketplace.config_missing_explainer_html',
-          edit_link: link_to(t('marketplace.marketplace.edit'), location(:edit)) ) %>
+          edit_link: link_to(t('marketplace.marketplace.edit.link_to'), location(:edit)) ) %>
   <% end %>
 <% end %>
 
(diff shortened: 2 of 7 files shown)


app/furniture/marketplace/marketplace_component.html.erb
```diff
@@ -2,15 +2,15 @@
...
+          edit_link: link_to(t('marketplace.marketplace.edit.link_to'), location(:edit)) ) %>
   <% end %>
 <% end %>
 
 <%= render cart %>
 
 <%- if policy(marketplace).edit? %>
   <div class="my-2 text-right">
-    <%= render ButtonComponent.new(scheme: :secondary, label: t('marketplace.marketplace.edit'),
+    <%= render ButtonComponent.new(scheme: :secondary, label: t('marketplace.marketplace.edit.link_to'),
```
daltonrpruitt, line 13: Genuine curiosity (since I'm still learning):
What is the reason the `button_to` can't be used here?

zspencer: the `MarketplaceComponent` is distinct from the `Marketplace::ManagementComponent`; and `button_to`  is only defined on the `Marketplace::ManagementComponent` at this time.

I could see a world where we do have the `button_to` available on `Marketplace::Component`, which would allow us to use it everywhere!

KellyAH (approved):

app/furniture/marketplace/marketplace_component.html.erb
daltonrpruitt: Coolio! I figured it had something to do with the hierarchy, but I wasn't seeing it. Thank you for the clarification!

app/furniture/marketplace/management_component.rb
```diff
@@ -7,5 +7,17 @@ def initialize(marketplace:, **kwargs)
 
       self.marketplace = marketplace
     end
+
+    def button_to(location, icon:)
```
anaulin, line 11: This is probably ok, but it could lead to some confusion, since it is shadowing the built-in Rails `button_to` helper.

Makes me wonder if what we want is to extend the `button_to` helper to take some arguments to use our `ButtonComponent` instead or in addition to this change.

zspencer: Yea, I think there's definitely a risk there; especially since in this case we need to also call `render`; where is in classic rails `button_to` does not require a render. 

My instincts are that it would be nice to start replace the built-in rails helpers (including the form-builder) so that folks who are coming straight from rails don't have to deal with the bunny-hop...

zspencer: I renamed to `button` for now; since `button_to` also presumes to make a form; and I didn't want to deal with that.

app/furniture/marketplace/management_component.rb
```diff
@@ -7,5 +7,17 @@ def initialize(marketplace:, **kwargs)
...
+      label, href = if location.is_a?(Symbol)
+        [t("marketplace.marketplace.#{location}.link_to"),
+          marketplace.location(location)]
+      else
+        [t("marketplace.#{location[:child]}.index.link_to"),
+          marketplace.location(**location)]
+      end
+
+      ButtonComponent.new(label: label, icon: icon, href: href, turbo_stream: false, method: :get, scheme: :secondary)
+    end
```
rosschapman, line 21: ❓ Could this be a separate component called `MarketplaceButton`?

I'm just becoming familiar with the ViewComponent framework and patterns, so you might have to school me here. But I don't understand the advantage of using this "container" class. Could you achieve the same thing with getting the marketplace instance from the controller and passing it to a `MarketplaceButton` like so? 

```erb
<%= render MarketPlaceButton(marketplace: marketplace, :edit, icon: :gear) if policy(marketplace).edit? %>
```

In JavaScript frontends/frameworks I've worked in we commonly build up domain-specific components from base components this way.

zspencer: I could see a very tiny `Marketplace::ManagementComponent::ButtonComponent` being useful for encapsulating this, but I don't see how that is substantively better than the helper method at this time.

Basically in my encapsulation-journey I tend to reach for things in the following order:

- a `variable` or `loop`
-  a `method` 
- a `Class`

At this point, the complexity doesn't feel significant enough to justify a class;; especially since the whole ruby-side of the component is < 100 lines at this point.

PR #1630, opened 2023-07-04
🐞🔨 `Journal`: Show error messages when `Entry` cannot be saved

zspencer:

Womp womp; apparently I had forgotten to do the `unprocessable_entity`, which meant that `turbo` was failing miserably instead of replacing the form when creating or updating a journal entry.

Well; now it's fixed!

### DIFF
--- a/app/furniture/journal/entry.rb
+++ b/app/furniture/journal/entry.rb
@@ -9,6 +9,7 @@ class Journal
     scope :recent, -> { order("published_at DESC NULLS FIRST") }
 
     attribute :headline, :string
+    validates :headline, presence: true
     attribute :body, :string
     validates :body, presence: true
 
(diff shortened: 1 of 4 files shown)


app/furniture/journal/entry.rb
```diff
@@ -9,6 +9,7 @@ class Entry < ApplicationRecord
     scope :recent, -> { order("published_at DESC NULLS FIRST") }
 
     attribute :headline, :string
+    validates :headline, presence: true
```
anaulin, line 12: Not really part of your PR, but now that we have the `strip_attributes` gem, it might be worth throwing that on for `headline` and `body`, just so that we don't store trailing/preceding whitespace.

anaulin (approved):
👍🏼

app/furniture/journal/entry.rb
zspencer: Yah! I was thinking about that then I forgot 🤦

PR #1632, opened 2023-07-04
🌸 `Journal`: The `Entry#body` form field fills the screen

zspencer:

I was doing some writing and noticed that I tend to drag the text area to make it bigger.

Now I won't have to do that!

## After
Screenshot 2023-07-03 at 9 30 26 PM

## Before
Screenshot 2023-07-03 at 9 30 41 PM


anaulin (approved):
Nice!

PR #1633, opened 2023-07-04
🥔✨ `Journal`: extract `Term`s from `Entry`s

zspencer:

Not yet linking to them; but that should be a decent next step.


app/furniture/journal/entry.rb
```diff
@@ -24,8 +24,11 @@ class Entry < ApplicationRecord
 
     # @!attribute journal
     #   @return [Journal::Journal]
-    belongs_to :journal, class_name: "Journal::Journal", inverse_of: :entries
-    delegate :room, :space, to: :journal
+    belongs_to :journal, inverse_of: :entries
+    has_one :room, through: :journal
+    has_one :space, through: :journal
```
zspencer, line 29: This was some minor cleanup from some poorly-connected linkages in the object graph. Happy to split this out.

app/furniture/journal/term.rb
```diff
@@ -0,0 +1,9 @@
+class Journal
+  class Term < ApplicationRecord
+    location(parent: :journal)
+
+    self.table_name = "journal_terms"
+    validates :canonical_term, presence: true
```
zspencer, line 6: Probably should also validate uniqueness scoped to the journal.

app/furniture/journal/entry.rb
```diff
@@ -46,6 +49,15 @@ def self.renderer
       )
     end
 
+    def extract_terms
+      body.scan(/#(\w+)/)&.flatten&.each do |term|
+        next if journal.terms.where("aliases ILIKE ?", term)
+          .or(journal.terms.where(canonical_term: term)).exists?
+
+        journal.terms.find_or_create_by!(canonical_term: term)
+      end
```
zspencer, line 58: TBH this method feels like it wants to be a factory method on `Term`, something like `Term.extract(body, journal: journal)`  (unless we can get `journal.terms.extract(body)` to work; which would be even better)

app/furniture/journal/renderer.rb
```diff
@@ -6,7 +6,9 @@ def autolink(link, link_type)
     end
 
     def postprocess(doc)
-      doc.gsub(/@([a-zA-Z\d]*)@(.*\.[a-zA-Z]*)/, '<a href="https://\2/@\1">@\1@\2</a>')
+      doc
+        .gsub(/@([a-zA-Z\d]*)@(.*\.[a-zA-Z]*)/, '<a href="https://\2/@\1">@\1@\2</a>')
+      # .gsub(/#(\w+)/, '<a href="../terms/\1">#\1</a>')
```
zspencer, line 11: Leftovers; just wanted to commit it before I fell asleep.

anaulin:
Are these meant to be... tags? keywords? something else? I feel like using more "standard" words from the blogging world might make it easier for both coders and people using the app.

app/furniture/journal/renderer.rb
anaulin: Should this commented-out code be removed, or are you keeping it for the future?

db/schema.rb
```diff
@@ -117,6 +117,15 @@
     t.index ["journal_id"], name: "index_journal_entries_on_journal_id"
   end
 
+  create_table "journal_terms", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "journal_id"
+    t.string "canonical_term"
+    t.text "aliases"
```
anaulin, line 123: Would be nice to make this an array, or something more structured than just text (which I assume right now it means to be comma-separated?).

anaulin:
Oh, I see now that you are extracting hashtags from the body. So maybe "tags" would be a better name?

anaulin (approved):
Nothing blocking, but I would love for this new thing to be named in a more "standard" way, e.g. "tags" or "keyword".

db/schema.rb
zspencer: I'll give it a shot! I have not used array types at the database level before but it shouldn't be too hard.

zspencer:
I like `Keyword`! I just didn't want to use `Tag` or `Hashtag`

app/furniture/journal/renderer.rb
zspencer: It should not be in this branch; but the follow on one...

PR #1637, opened 2023-07-06
`Journal`: Strip whitespace from `slug`, `headline`, and `name`

zspencer:
- https://github.com/zinc-collective/convene/pull/1630#discussion_r1253762788

> Not really part of your PR, but now that we have the
> `strip_attributes` gem, it might be worth throwing that on for
> `headline` and `body`, just so that we don't store trailing/preceding
> whitespace.


anaulin (approved):
Strip those attributes 💪🏼

PR #1638, opened 2023-07-06
🐞🔨🧹 `Marketplace`: Fix fumbled rename

zspencer:
- https://github.com/zinc-collective/convene/pull/1562#discussion_r1253794323

I did a rename operation and:
  - Renamed a table in the migration, _and_
  - Failed to confirm the migration was in only my branch.

That was v bright of me; and now it is fixed!

This also fixes a couple weird things that happen in the `db/schema.rb` file because there's a fight between what rubocop wants and what `rails db:schema:dump` auto-generates.


anaulin (approved):
🚢

Issue #1641, opened 2023-07-06
🥔✨ `Marketplace`: Onboarding `Vendors`!


zspencer:
- https://github.com/zinc-collective/convene/issues/831

## Use Cases
- [x] 🥔✨ `Marketplace`: Onboarding for `Products`
- [x] 🥔✨ `Marketplace`: Onboarding for `Delivery Areas`
- [x] 🥔✨ `Marketplace`: Onboarding for `Taxes`
- [x] 🥔✨ `Marketplace`: Onboarding for `Notifications`
- [ ] 🥔✨ `Marketplace`: Onboarding for `Payments`

zspencer:
I think this is good enough to be Potato tier; but am available for feedback if anyone is interested!

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

Issue #1576, opened 2023-06-19
🥔✨ `Marketplace`: Collecting Payments


zspencer:
- https://github.com/zinc-collective/convene/issues/831
- This describes the `Seller` experience. For the `Buyer` experience, see https://github.com/zinc-collective/convene/issues/1326

## Use Cases

- [x] https://github.com/zinc-collective/convene/issues/1622
- [ ] `Marketplace`: Accepting Payments via MercadoPago
- [ ] `Marketplace`: Accepting Payments via Paypal
- [ ] `Marketplace`: Accepting Payments via an Internal Currency

## Bugs

- [x] https://github.com/zinc-collective/convene/issues/1810

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #1643, opened 2023-07-06
🥔✨ `Marketplace`: Check `StripeAccount` when `Onboarding`

zspencer:
- https://github.com/zinc-collective/convene/issues/1641
- https://github.com/zinc-collective/convene/issues/1576

Now we encourage people to set up their `StripeAccount`!

We still need cleaner UI for that, but @rosschapman is working on that part.

### DIFF
--- a/app/furniture/marketplace/locales/en.yml
+++ b/app/furniture/marketplace/locales/en.yml
@@ -81,6 +81,8 @@ en:
       destroy:
         link_to: "Remove Product '%{name}'"
     stripe_accounts:
+      new:
+        link_to: "Add Stripe Account"
       show:
         link_to: "Stripe Account"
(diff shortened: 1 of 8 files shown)


app/furniture/marketplace/locales/en.yml
```diff
@@ -69,6 +69,10 @@ en:
         link_to: "Order History"
       show:
         link_to: "Order %{order_id}"
+    stripe_accounts:
+      new:
+        link_to:
+          "Manage Stripe"
```
daltonrpruitt, line 75: This seems to be a second `stripe_accounts` section at the same level as the one below?

zspencer: Woops! Good catch!

daltonrpruitt (approved):

PR #1647, opened 2023-07-09
🧹 `Marketplace`: Pull `StripeAccount` to its own page

zspencer:
- https://github.com/zinc-collective/convene/issues/1576
- https://github.com/zinc-collective/convene/issues/1641

This makes it possible to link directly to the `StripeAccount`. Maybe it should be `PaymentMethods` or something once we start adding in other payment sources; but this should do it for now; and make it a bit easier to start improving the Marketplace Onboarding flow.

### DIFF
--- a/app/components/svg_component.rb
+++ b/app/components/svg_component.rb
@@ -2,6 +2,10 @@ class SvgComponent < ApplicationComponent
   # { symbol: method returning path for symbol }
   ICON_MAPPINGS = {
     cart: :cart,
+    # @todo I don't know where we have been getting the SVG icons;
+    #       and I am hesitant to add one unless it's coming from
+    #       a library we are already using... - ZS
+    money: :cart,
     exclamation_triangle: :exclamation_triangle,
     gear: :gear,
     map: :map,
--- a/app/furniture/marketplace/management_component.rb
+++ b/app/furniture/marketplace/management_component.rb
@@ -16,8 +16,11 @@ class Marketplace
       label, href = if location.is_a?(Symbol)
         [t("marketplace.marketplace.#{location}.link_to"),
           marketplace.location(location)]
-      else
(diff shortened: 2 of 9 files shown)


app/components/svg_component.rb
```diff
@@ -2,6 +2,10 @@ class SvgComponent < ApplicationComponent
   # { symbol: method returning path for symbol }
   ICON_MAPPINGS = {
     cart: :cart,
+    # @todo I don't know where we have been getting the SVG icons;
+    #       and I am hesitant to add one unless it's coming from
+    #       a library we are already using... - ZS
+    money: :cart,
```
zspencer, line 8: @KellyAH or @anaulin - you both have found SVG icons; do you have a site you recommend with generous licensing?

app/furniture/marketplace/management_component.rb
```diff
@@ -16,9 +16,12 @@ def button(location, icon:)
       label, href = if location.is_a?(Symbol)
         [t("marketplace.marketplace.#{location}.link_to"),
           marketplace.location(location)]
-      else
+      elsif location[:child].to_s.pluralize == location[:child].to_s
```
daltonrpruitt, line 19: Non-blocking:
I feel that this check could be later extracted to some sort of `is_plural?` helper function, but I don't know if that would somewhat defeat the purpose behind the plural/singular design "constraints".

zspencer: yea, I think this is quite clugey; and a [`Refinement`](https://ruby-doc.org/core-3.1.0/Refinement.html) that allows strings and symbols to tell us whether they are singular or plural would be much preferred; especially if we see this kind of check emerging as a pattern.

daltonrpruitt (approved):

PR #1648, opened 2023-07-10
🥗🧹`Journal`: Make relationships of `Entry` Rails-able

zspencer:
- https://github.com/zinc-collective/convene/pull/1633/

I noticed I'd left the active-record relationship cables on the `Journal::Entry` loose. Getting rid of `delegate`s in favor of the `has_one :x, through: :y`


anaulin (approved):
🚢

PR #1649, opened 2023-07-10
🥔✨ `Journal`: Sprout `Keywords#show`

zspencer:

## UI

...Truly, a thing of beauty...
Screenshot 2023-07-20 at 10 39 14 AM

### DIFF
--- a/app/furniture/journal/keywords/show.html.erb
+++ b/app/furniture/journal/keywords/show.html.erb
@@ -1,0 +1,2 @@
+<%= keyword.canonical_keyword %>
+<%= keyword.aliases %>
--- a/app/furniture/journal/keywords/show.html.erb
+++ b/app/furniture/journal/keywords/show.html.erb
@@ -1,10 +1,9 @@
 <h1><%= keyword.canonical_keyword %></h1>
-<ul>
-<%= keyword&.aliases&.each do |aliass| %>
-  <li><%= aliass %></li>
-<% end %>
-</ul>
+<%- if keyword.aliases.present? %>
+  <p class="italic">Also known as <%= to_sentence(keyword.aliases) %></p>
+<%- end %>
 
+<h2>Entries about <span class="italic"><%= keyword.canonical_keyword %></span></h2>
 <ul>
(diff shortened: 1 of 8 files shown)


zspencer:
Actually, I think this needs moar love, and probably wants to come after #1663 so the keyword show page can show the `Entry` anyway!

app/furniture/journal/keywords/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1><%= keyword.canonical_keyword %></h1>
+<ul>
+<%= keyword&.aliases&.each do |aliass| %>
+  <li><%= aliass %></li>
```
anaulin, line 4: sssssssssss (thanks, Ruby)

anaulin (approved):
Codewise LGTM, but I feel like the UI needs a bit more context. Just looking at the screenshot I had no idea what I was looking at, or what the UI was intended for.

Is it meant to be something like:
"""
### Other keywords related to 'tested'
* tests
* testing
"""

Or something else?

I'm not clear on what is the use case here.

Is this for a reader of a Journal to find keywords aliases? What for? If we wanted to show a reader all posts under a topic we could just show them a list of all entries with a keyword and any of its aliases.

Or is this UI for a journal owner to manage aliases? Or something else?

Regardless, nothing blocking for merge.

app/furniture/journal/keywords/show.html.erb
zspencer: I was a/nnoyed/mused at the consequence of choosing `aliases` as a collection; and may wind up finding a different name to prevent the silliness.

zspencer:
Right now I am thinking it's just going to enumerate the entries that reference the keyword; but there are some other places I want to go:

- Showing a Definition for a Keywords  (or the `Entry` which serves as the canonical definition for the `Keyword`?)
- Managing Aliases and Merging Keywords
- Subscribe to `Entries` with the `Keyword`

I suppose I could start with `{journal_path}/entries?keyword[]=test` rather than `/keywords/{keyword}` for the Filtering and Subscribing to`Entries` by `Keyword`...

PR #1650, opened 2023-07-10
🧹 `Journal`: Move `Entry` rendering to `EntryComponent`

zspencer:

This refactor allows us to do post-processing of the markdown at the Component level, which has access to the rails URL builders; which should allow us to link to the `Keyword` within the `Entry#body`

### DIFF
--- a/app/furniture/journal/entry_component.rb
+++ b/app/furniture/journal/entry_component.rb
@@ -1,0 +1,30 @@
+class Journal
+  class EntryComponent < ApplicationComponent
+    include RendersMarkdown
+    attr_accessor :entry
+
+    def initialize(*args, entry:, **kwargs)
+      self.entry = entry
+      super(*args, **kwargs)
+    end
+
+    def body_html
+      render_markdown(entry.body)
+    end
+
+    def published_at
+      entry.published_at.to_fs(:long_ordinal)
+    end
+
+    def self.renderer
+      @_renderer ||= Redcarpet::Markdown.new(
+        Renderer.new(filter_html: true, with_toc_data: true),
+        autolink: true, strikethrough: true,
+        no_intra_emphasis: true,
+        lax_spacing: true,
+        fenced_code_blocks: true, disable_indented_code_blocks: true,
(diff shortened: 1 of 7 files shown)


zspencer:
Here's where it links to the `Keyword`: https://github.com/zinc-collective/convene/pull/1684

app/furniture/journal/entry_component.rb
```diff
@@ -0,0 +1,30 @@
...
+  class EntryComponent < ApplicationComponent
+    include RendersMarkdown
+    attr_accessor :entry
+
+    def initialize(*args, entry:, **kwargs)
+      self.entry = entry
+      super(*args, **kwargs)
+    end
+
+    def body_html
```
rosschapman, line 11: What about `content_html` not to confuse with the`<body>` tag. This might be more consistent with other components that define header/content/footer slots?

spec/furniture/journal/entry_component_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
+
+RSpec.describe Journal::EntryComponent, type: :component do
+  subject(:output) { render_inline(component) }
+
+  let(:component) { described_class.new(entry: entry) }
+  let(:entry) { create(:journal_entry, body: "https://www.google.com @zee@weirder.earth #GoodTimes") }
+
+  it { is_expected.to have_link("https://www.google.com", href: "https://www.google.com") }
```
rosschapman, line 9: Any reason you're just testing the link existence and not other strings?

rosschapman (approved):

app/furniture/journal/entry_component.rb
zspencer: Ah, I like that! I will do a follow up PR to rename `Entry#body` to `Entry#content`

spec/furniture/journal/entry_component_spec.rb
zspencer: Well, I could have done some tests to ensure headings have the appropriate ids and such; but this was the thrust of the work I was doing so it's what I focused on.

PR #1663, opened 2023-07-13
🧹 `Journal`: `Entries` store their `Keywords`

zspencer:

This gets us a little bit closer to being able to browse `Entry` by `Keyword`s, since each `Entry` will know it's keywords

### DIFF
--- a/app/furniture/journal/keyword.rb
+++ b/app/furniture/journal/keyword.rb
@@ -7,9 +7,11 @@ class Journal
     belongs_to :journal, inverse_of: :keywords
 
     def self.extract_and_create_from!(body)
-      body.scan(/#(\w+)/)&.flatten&.each do |keyword|
-        next if where(":aliases = ANY (aliases)", aliases: keyword)
-          .or(where(canonical_keyword: keyword)).exists?
+      body.scan(/#(\w+)/)&.flatten&.map do |keyword|
+        existing_keyword = where(":aliases = ANY (aliases)", aliases: keyword)
+          .or(where(canonical_keyword: keyword)).first
+
+        next existing_keyword if existing_keyword.present?
 
         find_or_create_by!(canonical_keyword: keyword)
       end
--- a/app/furniture/journal/keyword.rb
+++ b/app/furniture/journal/keyword.rb
@@ -6,10 +6,14 @@ class Journal
     validates :canonical_keyword, presence: true, uniqueness: {case_sensitive: false, scope: :journal_id}
     belongs_to :journal, inverse_of: :keywords
 
+    scope(:search, lambda do |*keywords|
+      where("lower(aliases::text)::text[] && ARRAY[?]::text[]", keywords.map(&:downcase))
+          .or(where("lower(canonical_keyword) IN (?)", keywords.map(&:downcase)))
+    end)
+
     def self.extract_and_create_from!(body)
       body.scan(/#(\w+)/)&.flatten&.map do |keyword|
-        existing_keyword = where(":aliases = ANY (aliases)", aliases: keyword)
-          .or(where(canonical_keyword: keyword)).first
+        existing_keyword = search(keyword).first
 
         next existing_keyword if existing_keyword.present?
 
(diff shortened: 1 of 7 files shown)


app/furniture/journal/keyword.rb
```diff
@@ -7,9 +7,11 @@ class Keyword < ApplicationRecord
     belongs_to :journal, inverse_of: :keywords
 
     def self.extract_and_create_from!(body)
-      body.scan(/#(\w+)/)&.flatten&.each do |keyword|
-        next if where(":aliases = ANY (aliases)", aliases: keyword)
-          .or(where(canonical_keyword: keyword)).exists?
+      body.scan(/#(\w+)/)&.flatten&.map do |keyword|
+        existing_keyword = where(":aliases = ANY (aliases)", aliases: keyword)
+          .or(where(canonical_keyword: keyword)).first
```
anaulin, line 12: I realize this query is from before your PR, but reading it over right now it struck me that it feels like it should be a nice little scope, which would help tighten up this entire loop a bit.

db/migrate/20230713230334_journal_add_keywords_to_entries.rb
```diff
@@ -0,0 +1,5 @@
+class JournalAddKeywordsToEntries < ActiveRecord::Migration[7.0]
+  def change
+    add_column :journal_entries, :keywords, :string, array: true
```
anaulin, line 3: I'm surprised these are not tied in any way to the ids of the Journal keywords -- I would have expected you'd want that to show "all entries with keyword X in Journal".

anaulin (approved):
Having keywords at both the Journal and the Entry level feels a little bit overkill to me, but I'm sure you have some larger plan for what that's necessary.

daltonrpruitt (approved):
I spent too long trying to figure out where the keywords were shown in the UI....
I should read more deeply. 😬

app/furniture/journal/keyword.rb
zspencer: Agree; I would love for this to be a scope!

db/migrate/20230713230334_journal_add_keywords_to_entries.rb
zspencer: You know, I was thinking a join table would be "too much" but I suppose it's not?

The way I was planning to find the `Entries` was indexing this keywords field and using the scope you suggested above; but I do worry that that would make the relationy bits a bit less railsy

zspencer:
> Having keywords at both the Journal and the Entry level feels a little bit overkill to me, but I'm sure you have some larger plan for what that's necessary.

I was thinking that an `Entry` >=< `Keyword` relationship is a many-to-many, and the `Journal` =< `Keyword` is one-to-many. 

> I spent too long trying to figure out where the keywords were shown in the UI.... 
> I should read more deeply. 😬

I am trying to take very very tiny steps; instead of just YOLO-Shipping the whole thing lol. It does make it a bit harder to figure out the trajectory tho.

app/furniture/journal/keyword.rb
```diff
@@ -6,10 +6,16 @@ class Keyword < ApplicationRecord
     validates :canonical_keyword, presence: true, uniqueness: {case_sensitive: false, scope: :journal_id}
     belongs_to :journal, inverse_of: :keywords
 
+    scope(:search, lambda do |*keywords|
+      where("lower(aliases::text)::text[] && ARRAY[?]::text[]", keywords.map(&:downcase))
+          .or(where("lower(canonical_keyword) IN (?)", keywords.map(&:downcase)))
```
zspencer, line 11: There is probably a way less awkward way of doing this; but this is what I was able to come up with 🙃

zspencer (commented):
OK, I've extracted the scope; and I am wanting to see how far I can get before sprouting the join table


The reason I'm avoiding the join table is because I work in another code base with hundreds of thousands of "taggings" in a join table; and the performance characteristics are pretty bad; even with fully indexing the taggings table across the tag_id and entry_id in both directions!

My gut says that storing the tags an Entry uses and indexing *that* will be far more performant because using the canonical keyword as a kind of "natural join" when traversing from the keyword to the entries; and from the entries to the keywords.

But I'm willing to drop that concern.

PR #1675, opened 2023-07-15
🥗🧹 `Journal`: `Entry` stores each `Keyword` it uses only once.

zspencer:

Turns out we were storing the keyword every time it was used; which... isn't bad but isn't great either!


anaulin (approved):
Good catch!

PR #1680, opened 2023-07-19
🥗 `Marketplace`: Buying `Products` as a `Guest` System Test

zspencer:

The bug is fixed in https://github.com/zinc-collective/convene/pull/1685

For grokability reasons, I am considering splitting this into a few different PRs:

- A introduces the new ENV variable
- B extracts the shared system spec helpers
- C implements the stripe cli forwarding
- D implements the system test

### DIFF
--- a/spec/furniture/marketplace/buying_products_system_spec.rb
+++ b/spec/furniture/marketplace/buying_products_system_spec.rb
@@ -61,8 +61,10 @@ describe "Marketplace: Buying Products", type: :system do
 
   def with_stripe(listen_url:, &block)
     signing_secret = nil
+    waiting = 0
+    command = "stripe listen --api-key #{ENV.fetch("DEFAULT_STRIPE_API_KEY")} --forward-to #{listen_url}"
     thread = Thread.new do
-      Open3.popen3("stripe listen --forward-to #{listen_url}") do |stdin, stdout, stderr, thread|
+      Open3.popen3(command) do |stdin, stdout, stderr, thread|
         # read each stream from a new thread
         {out: stdout, err: stderr}.each do |key, stream|
           Thread.new do
@@ -74,8 +76,9 @@ describe "Marketplace: Buying Products", type: :system do
         thread.join
       end
     end
-    sleep(1) until signing_secret
+    raise "Can't connect to Stripe using command `#{command}`" if waiting > 5
 
+    sleep(1) until signing_secret
     yield(signing_secret)
     thread.kill
   end
--- a/spec/furniture/marketplace/buying_products_system_spec.rb
+++ b/spec/furniture/marketplace/buying_products_system_spec.rb
@@ -76,9 +76,13 @@ describe "Marketplace: Buying Products", type: :system do
         thread.join
       end
     end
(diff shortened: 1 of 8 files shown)


zspencer:
Oh just glorious... It *does* work on CI! And it works locally if I run `bin/rails assets:precompile`

So maybe there's something to clean up there; too.

zspencer:
Soooo, this totally goes through the full checkout flow, and exposes the bug that was reported; which makes me *very happy*

The issue is likely in 1 of 3 places:

- The `Marketplace::OrderPolicy::Scope`, which may not be correctly handling `Guest` `Shopper`s
- Somewhere in the `Marketplace::Controller`, where I anticipate I am doing some... arcana... to persist data in the database in relation to a cookie... Which I probably shouldn't be maybe doing?
- The `OrderPolicy#show?` method;  which perhaps could allow anyone to see any order within some "reasonable" duration? (ugh I do not like this)

zspencer:
Woooo! Stripe gets hit in CI; the tests go green; and people should be getting traceable errors when things are not quite right.

I expect there are many places to get lost when trying to debug things going sideways in this horrifying multi-threaded hydra...

spec/furniture/marketplace/buying_products_system_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    fill_in("cardCvc", with: card_cvc)
+    fill_in("billingName", with: billing_name)
+    fill_in("email", with: email)
+    fill_in("billingPostalCode", with: billing_postal_code)
+
+    find("label[for='enableStripePass']").click
+    find("*[data-testid='hosted-payment-submit-button']").click
+  end
+
+  def with_stripe(listen_url:, &block)
```
anaulin, line 62: This method is complex enough that I feel it warrants a two-sentence comment explaining that it spawns the Stripe CLI and uses threads to listen to its stdout/stderr to get the output, and times out after 5 secs of polling and sleeping waiting for the secret.

spec/furniture/marketplace/buying_products_system_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    signing_secret = nil
+    waiting = 0
+    command = "stripe listen --api-key #{ENV.fetch("DEFAULT_STRIPE_API_KEY")} --forward-to #{listen_url}"
+    thread = Thread.new do
+      Open3.popen3(command) do |stdin, stdout, stderr, thread|
```
anaulin, line 67:
```suggestion
      Open3.popen3(command) do |_, stdout, stderr, thread|
```

anaulin (approved):
Amazing. Heroic, even. 😻🥹

In the end, I don't think this multi-threaded implementation is _that_ complex, and it feels understandable-ish.

spec/furniture/marketplace/buying_products_system_spec.rb
zspencer: I am thinking of pulling it out into `spec/support/stripe_cli.rb` and giving it a bit more documentation.

zspencer:
I am waiting to merge this until https://github.com/zinc-collective/convene/pull/1685 is approved.

PR #1682, opened 2023-07-20
🌸✨ `Marketplace`: `Management` includes ` Payment Settings` screen

rosschapman:
This changeset is more the result of developer onboarding (for me) as I prototyped (in code) consolidating configuration for payment related things. This is merely a piece of a partial solve for requirements of https://github.com/zinc-collective/convene/discussions/1605 or https://github.com/zinc-collective/convene/discussions/1622. But was invaluable to help me understand _how to frontend_ in the Marketplace domain. I think this is going to be a boon for my 🍐 session with @KellyAH next week. 

### Notable details: 
- The CardComponent styling refactor does force automatically bottom-aligned CTAs. 

After: 
image
image

### DIFF
--- a/app/components/marketplace/stripe_overview_component.rb
+++ b/app/components/marketplace/stripe_overview_component.rb
@@ -1,7 +1,7 @@
 # Renders a tile with a link to view the Marketplace Stripe account page
 class Marketplace::StripeOverviewComponent < ApplicationComponent
   attr_accessor :marketplace
-  attr_accessor :marketplace_stripe_utility
+  delegate :stripe_utility, to: :marketplace, prefix: true
 
   def initialize(marketplace:, **kwargs)
     super(**kwargs)
--- a/app/components/marketplace/stripe_overview_component.rb
+++ b/app/components/marketplace/stripe_overview_component.rb
@@ -1,0 +1,12 @@
+# Renders a tile with a link to view the Marketplace Stripe account page
+class Marketplace::StripeOverviewComponent < ApplicationComponent
+  attr_accessor :marketplace
+  attr_accessor :marketplace_stripe_utility
+
+  def initialize(marketplace:, **kwargs)
+    super(**kwargs)
+
+    @marketplace = marketplace
+    @marketplace_stripe_utility = marketplace.stripe_utility
+  end
+end
--- a/app/components/marketplace/stripe_overview_component/stripe_overview_component.html.erb
+++ b/app/components/marketplace/stripe_overview_component/stripe_overview_component.html.erb
@@ -1,0 +1,14 @@
+<%= render CardComponent.new(classes: "flex flex-col h-full
+") do %>
+  <header class="flex font-bold">
+    <%= marketplace_stripe_utility.name %>
+  </header>
+  <footer class="flex flex-row justify-between mt-3">
+    <%= render ButtonComponent.new(
+      label: "View #{t('marketplace.stripe_accounts.show.link_to')}",
+      title: t('marketplace.stripe_accounts.show.link_to'),
+      href: marketplace.location(:index, child: :stripe_account),
+      method: :get,
+      scheme: :secondary) %>
+  </footer>
+<%- end %>
--- a/app/furniture/marketplace/breadcrumbs.rb
+++ b/app/furniture/marketplace/breadcrumbs.rb
@@ -108,3 +108,8 @@ crumb :edit_tax_rate do |tax_rate|
   parent :marketplace_tax_rates, tax_rate.marketplace
   link "Edit Tax Rate '#{tax_rate.label}'", marketplace.location(:new, child: :tax_rate)
 end
+
+crumb :payment_settings do |marketplace|
+  parent :edit_marketplace, marketplace
+  link "Manage Payment Settings", marketplace.location(:index, child: :payment_settings)
+end
--- a/app/furniture/marketplace/management_component.html.erb
+++ b/app/furniture/marketplace/management_component.html.erb
@@ -12,7 +12,7 @@
     <%= content %>
     <% card.with_footer do %>
       <nav class="grid grid-cols-1 sm:grid-cols-2 md:grid-cols-3 gap-3">
-        <%= render button({child: :stripe_account}, icon: :money) if policy(marketplace).edit? %>
+        <%= render button({child: :payment_settings}, icon: :money) if policy(marketplace).edit? %>
         <%= render button({child: :products}, icon: :tag)  if policy(marketplace.products).index? %>
         <%= render button({child: :delivery_areas}, icon: :map)  if policy(marketplace.delivery_areas).index? %>
         <%= render button({child: :tax_rates}, icon: :receipt_percent)  if policy(marketplace.tax_rates).index? %>
--- a/app/furniture/marketplace/marketplace.rb
+++ b/app/furniture/marketplace/marketplace.rb
@@ -41,9 +41,13 @@ class Marketplace
       products.present? && stripe_api_key? && delivery_areas.present? && stripe_account_connected?
     end
 
+    def stripe_utility
+      space.utilities.find_by!(utility_slug: :stripe).utility
+    end
+
     # The Secret Stripe API key belonging to the owner of the Marketplace
     def stripe_api_key
-      space.utilities.find_by!(utility_slug: :stripe).utility.api_token
+      stripe_utility.api_token
     end
 
     def stripe_api_key?
--- a/app/furniture/marketplace/marketplace.rb
+++ b/app/furniture/marketplace/marketplace.rb
@@ -42,7 +42,7 @@ class Marketplace
     end
 
     def stripe_utility
-      space.utilities.find_by!(utility_slug: :stripe).utility
+      @stripe_utility ||= space.utilities.find_by!(utility_slug: :stripe).utility
     end
 
     # The Secret Stripe API key belonging to the owner of the Marketplace
--- a/app/furniture/marketplace/payment_settings/index.html.erb
+++ b/app/furniture/marketplace/payment_settings/index.html.erb
@@ -1,0 +1,24 @@
+<%- breadcrumb :payment_settings, marketplace %>
+<%= render Marketplace::ManagementComponent.new(marketplace: marketplace) do %>
+  <h1>Payment Settings</h1>
+  <main>
+    <div class="grid grid-cols-1 gap-3 sm:gap-5 sm:grid-cols-3">
+      <%= render Marketplace::StripeOverviewComponent.new(marketplace: marketplace) %>
+      <%= render CardComponent.new(classes: "flex flex-col h-full") do %>
+        <header class="flex font-bold">
+          Did we miss something?
+        </header>
+        <div class="grow flex flex-col justify-between">
+          Let us know about other payment service options that will help your marketplace to thrive.
+        </div>
+        <footer class="flex flex-row justify-between mt-3">
+          <%= render ButtonComponent.new(
+            label: "Contact Us",
+            title: "Contact Us",
+            href: "mailto:#{ApplicationMailer::DEFAULT_FROM}",
+            scheme: :secondary) %>
+        </footer>
+      <% end %>
+    </div>
+  </main>
+<% end %>
--- a/app/furniture/marketplace/routes.rb
+++ b/app/furniture/marketplace/routes.rb
@@ -16,6 +16,7 @@ class Marketplace
         router.resource :stripe_account, only: [:show, :new, :create]
         router.resources :stripe_events
         router.resources :tax_rates
+        router.resources :payment_settings, only: [:index]
       end
     end
   end
--- a/spec/components/marketplace/stripe_overview_component_spec.rb
+++ b/spec/components/marketplace/stripe_overview_component_spec.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Marketplace::StripeOverviewComponent, type: :component do
+  pending "add some examples to (or delete) #{__FILE__}"
+
+  # it "renders something useful" do
+  #   expect(
+  #     render_inline(described_class.new(attr: "value")) { "Hello, components!" }.css("p").to_html
+  #   ).to include(
+  #     "Hello, components!"
+  #   )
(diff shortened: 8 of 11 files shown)


app/components/card_component.html.erb
```diff
@@ -1,9 +1,12 @@
-<div <%== attributes(classes: card_classes) %>>
-  <%= content %>
+<div class="<%== card_classes_wrapper %>">
+  <% unless content.blank? %>
+    <div <%== attributes(classes: card_classes_content) %>">
```
rosschapman, line 3: If I've interpreted the code correctly, we actually want to merge the classes declared on init into the content wrapper.

app/components/card_component.rb
rosschapman, line 1: Again, hopefully I've interpreted the intent of the code. But I believe separating the class definitions this way makes styling simpler.

app/furniture/marketplace/management_component.html.erb
```diff
@@ -6,18 +6,19 @@
...
-  <% card.with_footer do %>
-    <nav class="grid grid-cols-1 sm:grid-cols-2 md:grid-cols-3 gap-3">
-      <%= render button({child: :stripe_account}, icon: :money) if policy(marketplace).edit? %>
-      <%= render button({child: :products}, icon: :tag)  if policy(marketplace.products).index? %>
-      <%= render button({child: :delivery_areas}, icon: :map)  if policy(marketplace.delivery_areas).index? %>
-      <%= render button({child: :tax_rates}, icon: :receipt_percent)  if policy(marketplace.tax_rates).index? %>
-      <%= render button({child: :orders}, icon: :cart)  if policy(marketplace.orders).index? %>
-      <%= render button({child: :notification_methods}, icon: :bell)  if policy(marketplace.notification_methods).index? %>
-    </nav>
+<div class="flex flex-col space-y-3">
```
rosschapman, line 9: This is more verbose but inverts responsibility for _layout_ styling to the parent, which generally preferred since it's easier to reason about child(ren) layout from the top level.

app/furniture/marketplace/marketplace.rb
```diff
@@ -30,9 +30,13 @@ def ready_for_shopping?
       products.present? && stripe_api_key? && delivery_areas.present? && stripe_account_connected?
     end
 
+    def stripe_utility
+      space.utilities.find_by!(utility_slug: :stripe).utility
+    end
```
rosschapman, line 35: Then I can also reuse this in the new overview tile component: https://github.com/zinc-collective/convene/pull/1682/files?diff=split&w=1#diff-49b1666c6a6edea9fcc01d928b068af8cb31fbf259e376289098a3c6d2a9379bR9

app/furniture/marketplace/payment_settings/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        </div>
+        <footer class="flex flex-row justify-between mt-3">
+          <%= render ButtonComponent.new(
+            label: "Contact Us",
+            title: "Contact Us",
+            href: "mailto:hello@zinc.coop",
+            scheme: :secondary) %>
+        </footer>
+      <% end %>
+    </div>
```
rosschapman, line 22: Useful onboarding? Was very trivial to add and makes the tiling look less lonely with just Stripe there.

app/furniture/marketplace/routes.rb
```diff
@@ -15,6 +15,7 @@ def self.append_routes(router)
         router.resource :stripe_account, only: [:show, :new, :create]
         router.resources :stripe_events
         router.resources :tax_rates
+        router.resources :payment_settings
```
rosschapman, line 18: ❓ Should I do `only: [:index]` or it doesn't really matter?

spec/components/marketplace/stripe_overview_component_spec.rb
rosschapman, line 1: ❓ Do we write individual component render tests?

app/furniture/marketplace/routes.rb
zspencer: `only: [:index]` is a bit tighter; but it's a six-in-one-half-dozen-in-the-other kind of thing.

app/components/card_component.html.erb
```diff
@@ -1,9 +1,12 @@
-<div <%== attributes(classes: card_classes) %>>
-  <%= content %>
+<div class="<%== card_classes_wrapper %>">
+  <% unless content.blank? %>
```
zspencer, line 2: Can this use [`if content?`](https://viewcomponent.org/api.html#content--boolean)?

app/components/card_component.html.erb
zspencer, line 3: I know there's a reason we use `attributes` instead of `class="#{card_content_classes}"` but for the life of me I cannot remember why...

Also, minor preference for `card_content_classes` vs `card_classes_content`, just because it feels more like `content` is the branch in the "what method am I reaching for?!" decision tree and `classes` is the leaf.

app/furniture/marketplace/management_component.html.erb
zspencer: Does this mean we need to find the places we're using the `CardComponent` and make sure we add a section or div that determines their layout?

If so, can we make that an independent PR?

app/furniture/marketplace/marketplace.rb
zspencer: Nice! I do continue to worry that the way we find the `stripe_utility` is fraught with peril; but that's out of scope.

app/furniture/marketplace/payment_settings/index.html.erb
zspencer: I like the idea; but would rather we did it outside of this PR. A few other things:

- Can we leverage or adjust the `EMAIL_DEFAULT_FROM` environment variable for the contact email?
- Do we want actively solicit guidance and feedback through the application? Or do we want to keep it flowing through the people who are interacting with our Client Owners?

app/furniture/marketplace/breadcrumbs.rb
```diff
@@ -98,3 +98,8 @@
   parent :marketplace_tax_rates, tax_rate.marketplace
   link "Edit Tax Rate '#{tax_rate.label}'", marketplace.location(:new, child: :tax_rate)
 end
+
+crumb :payment_settings do |marketplace|
+  parent :edit_marketplace, marketplace
+  link "Manage Payment Settings", marketplace.location(:index, child: :payment_settings)
```
zspencer, line 104: Can we use the value stored in `I18n.t("marketplace.payment_settings.index.link_to")` so that the text is consistent in the breadcrumb and the links to the breadcrumb?

zspencer (approved):
Once you get the stuff passing, I think this is "Good enough" to merge; but some things that would make me very happy would be:

- Adding some request and component specs for the new components and endpoints
- A little bit tighter focus in the PR. I see three+ changes:
  - `CardComponent` is getting refactored 
  - Adding a solicitation for feedback
  - Adding a brand new page (`Marketplace::PaymentSettings#index`)
  - Maybe some amount of restyling is going on?

app/furniture/marketplace/payment_settings/index.html.erb
anaulin: I like this a lot. Given that it is 20 lines of code and it feels relevant to this PR, I don't know that it is worth pulling out into a separate PR.

The email in the mailto should definitely be `ApplicationMailer::DEFAULT_FROM`, which in our prod installation is the `convene-support@` email, but in other installations could be configured to something else. (Keep in mind that in theory someone who isn't us might be running their own Convene installation.)

spec/components/marketplace/stripe_overview_component_spec.rb
anaulin: Only if you feel that the complexity of the component warrants it.

app/components/marketplace/stripe_overview_component/stripe_overview_component.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= render CardComponent.new(classes: "flex flex-col h-full
+") do %>
+  <header class="flex font-bold">
+    <%= marketplace_stripe_utility.name %>
+  </header>
+  <div class="grow flex flex-col justify-between">
+    <%# [TODO] %>
```
anaulin, line 7: What is meant to come here? If there is a plan to do this in a follow-up, I would either put a summary of what the TODO is in this comment, or remove this TODO part from the code and make an issue instead.

spec/components/marketplace/stripe_overview_component_spec.rb
anaulin: Looking at your `StripeOverviewComponent`, to me it feels like it doesn't have much complex logic, and it would not be worth writing much of a render spec, beyond just testing that it renders without exploding.

app/components/card_component.rb
```diff
@@ -3,14 +3,27 @@ class CardComponent < ApplicationComponent
...
+    ].compact.join(" ")
+  end
+
+  def card_classes_footer
+    [
+      "bg-purple-50",
       "p-4",
       "sm:p-6",
-      ("rounded-b-none" if footer?)
+      ("rounded-t-none" unless content.blank?),
```
anaulin: I note that this changes the logic from checking for footer content vs checking for content overall. I trust you that there was a reason for that (I don't remember the original reason why we were checking for footer content).

Also, reminder that in a ViewComponent you can always use `content?`, instead of content.blank? (https://viewcomponent.org/api.html#content--boolean).

app/furniture/marketplace/management_component.html.erb
anaulin, line 9: This makes me think of the Evil Martian's "wrapped component" helper for ViewComponents, from their `view-component_contrib` gem. It basically lets you do exactly this, but as a tidier helper that you we could add to our `ApplicationComponent`:
https://github.com/palkan/view_component-contrib#wrapped-components

(This gem has some other goodies that we might benefit from, and they are actively adding new features to it as they work more with ViewComponent. This feels like maybe a good excuse to add it to our codebase, but in that case it definitely feels like a separate PR.)

anaulin (approved):
I love to see a useful, non-empty landing page for "Manage Marketplace"! 🌷

anaulin:
FYI, since you have lint errors, we have some scripts under `bin` that you can use to quickly autoformat your changed files, e.g. for Ruby: https://github.com/zinc-collective/convene/blob/main/bin/lint-changed-ruby

app/furniture/marketplace/management_component.html.erb
rosschapman: Interesting, I'll check it out. Seems like this could be useful. Although the `render?` api is somewhat of a reversal from the way I've built frontend. Generally I've worked from the perspective that higher level components define conditions for child rendering.

rosschapman:
@zspencer @anaulin thanks for the feedback. I'll break up this work into separate PRs so the separate things I'm addressing are clearer. I did my due diligence and manually QA'd everywhere we use `CardComponent`, so I think my changes are safe. I won't let this hard block Square work, though, since that is higher prio.

app/components/card_component.html.erb
rosschapman: Ok I remember now. I was having some trouble with `content?` actually working right? I didn't investigate at the time. But check this out. I threw a _debugger_ in the template and when I print out the value of `content?` and `content` respectively I get _a Proc_ and _nil._ 

image

Whereas if I print out `footer?` the value is `true`

rosschapman: Ohhh I see what's happening. `content?` is technically truthy and returning a proc, because that proc is actually the nested `CardComponent` block in the `ManagementComponent`.

zspencer: That is weird as heck. I wonder if we're doing anything (like, defining a `content` or `content?` method). I found [the implementation](https://github.com/ViewComponent/view_component/blob/3487283035bbcb868e46f93fcfb6abadfeed650a/lib/view_component/base.rb#L269-L276) but it didn't shine much light for me.

rosschapman: My goal was to avoid this empty white space above the Manage Marketplace navigation when there's no content. 
image

rosschapman:
Closing in favor of smaller, separate PRs.

app/furniture/marketplace/payment_settings/index.html.erb
rosschapman: Chat about this strategy in Ensemble? My thought was that 1) as a skeleton crew, and 2) with delegates for our client owners inconsistently participating over time (part-timers, some-timers, etc...), that an evergreen solicitation in the product (a virtual extension of our "product" people) could be light-weight mechanism for feedback intake that would fill in the gaps.

anaulin: Yep, let's talk about this on Wednesday. I do like it as an idea, and a PoC for a type of component we could have in more places ("missing a Gizmo, let us know...", etc).

rosschapman: Ohhhh, I like the idea of a generic component for that.

app/components/marketplace/stripe_overview_component.rb
```diff
@@ -0,0 +1,12 @@
+# Renders a tile with a link to view the Marketplace Stripe account page
+class Marketplace::StripeOverviewComponent < ApplicationComponent
+  attr_accessor :marketplace
+  attr_accessor :marketplace_stripe_utility
```
zspencer, line 4: It's slightly more ruby-esque to do `delegate :stripe_utility, to: :marketplace, prefix: true`.

app/furniture/marketplace/marketplace.rb
```diff
@@ -41,9 +41,13 @@ def ready_for_shopping?
       products.present? && stripe_api_key? && delivery_areas.present? && stripe_account_connected?
     end
 
+    def stripe_utility
+      space.utilities.find_by!(utility_slug: :stripe).utility
```
zspencer, line 45: We may want to memoize this (`@stripe_utility ||= ...`)

zspencer (approved):
Ship it!

app/components/marketplace/stripe_overview_component.rb
rosschapman: Oh, nice tip!

PR #1683, opened 2023-07-20
🛠️ `Neighborhood`: Bump node to latest v16

zspencer:
- https://github.com/zinc-collective/convene/pull/1678#issuecomment-1643024009

It appears that local yarn installs are broken due to a minimum version requirement from a package.

Unfortunately, I don't know how to mirror `.nvmrc` and `package.json` to respect the same version; so I have bumped the nvmrc in particular.


rosschapman (approved):

anaulin:
Thanks for fixing this @zspencer 💗

zspencer:
> Hmm. Do we need or use the nvmrc? Is there no automatic way to keep the version in the package.json in sync with this file?

I haven't the foggiest. My inclination is towards dropping it; that way people can set their local versions as they see fit.

Issue #1657, opened 2023-07-13
🐞 `Marketplace`: `Guest` `Shopper` is shown a 404 after `Checkout`


anaulin:
- https://github.com/zinc-collective/convene/issues/1326
- https://github.com/zinc-collective/convene/issues/1661

After a successful purchase from the Marketplace, a shopper got redirected to their order page (https://marketplace.piikup.com/spaces/piikup/rooms/crumbleandwhisk/marketplaces/4cad5db3-562a-4630-b31d-fc6a253ea5a9/orders/45f6c244-4e32-43c1-aa4b-8e002e278d81?stripe_session_id=cs_live_b1GZZOiJS3UGIqMKQdk5PMAPiwtdA3Me3ZXIYkePdKHoGnGYXRo4mJdtMr), but this page is only visible to logged-in users, so instead they got the standard Rails 404 page.

We should make it so that a successful order confirmation page is visible even if the user is logged-out of Convene, which is going to be the most common case. One way to do this would be to create a separate "order confirmed" page that is visible for logged out users and directs users to log into Convene to see their complete order history.

zspencer:
I'm planning to work on this either between now and Wednesday or on Wednesday.

zspencer:
I have been able to confirm this through the automated system test; are there any @zinc-collective/contributors interested in double-checking that the fix is "good enough?"

If so, assign this issue to yourself and run through placing an order! (Note: This is... not trivial locally due to the need to be connected to `Stripe`, so it may not be an ideal thing to tackel unless you want to work through that molasses or have already done so)

If not, I'll close this out next time I'm looking at my assigned issues.

PR #1685, opened 2023-07-20
🐞🔨 `Marketplace`: `Guest` may see recent `Order`s

zspencer:
- Alternative to https://github.com/zinc-collective/convene/pull/1688

This allows `Orders` placed by `Guests` to be seen by any `Guest` for 1 week; after which they will 404.

There's probably something smerter to do around the `Marketplace::Policy` to use the `Shopper`, which *is* both present and persisted in all cases...

### DIFF
--- a/app/furniture/marketplace/order_policy.rb
+++ b/app/furniture/marketplace/order_policy.rb
@@ -6,8 +6,13 @@ class Marketplace
       def resolve
         return scope.all if person.operator?
 
-        scope.joins(marketplace: [:room]).where(rooms: {space_id: person.spaces})
-          .or(scope.where(shopper: Shopper.where(person: person)))
+        if person.authenticated?
+          scope.where(rooms: {space_id: person.spaces})
+            .or(scope.where(shopper: Shopper.where(person: person)))
+        else
+          scope.where(shopper: Shopper.where(person: nil))
+            .where("marketplace_orders.created_at > ?", 1.week.ago)
+        end.joins(marketplace: [:room])
       end
     end
   end
--- a/spec/furniture/marketplace/orders_controller_request_spec.rb
+++ b/spec/furniture/marketplace/orders_controller_request_spec.rb
@@ -24,8 +24,9 @@ RSpec.describe Marketplace::OrdersController, type: :request do
     end
 
     context "when the order is not viewable by the current person" do
+      before { sign_in(space, create(:person)) }
+
       specify { expect { perform_request }.to raise_error(ActiveRecord::RecordNotFound) }
-      # it { is_expected.to be_not_found}
     end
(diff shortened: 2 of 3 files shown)


app/furniture/marketplace/order_policy.rb
```diff
@@ -6,8 +6,13 @@ class Scope < ApplicationScope
...
 
-        scope.joins(marketplace: [:room]).where(rooms: {space_id: person.spaces})
-          .or(scope.where(shopper: Shopper.where(person: person)))
+        if person.authenticated?
+          scope.where(rooms: {space_id: person.spaces})
+            .or(scope.where(shopper: Shopper.where(person: person)))
+        else
+          scope.where(shopper: Shopper.where(person: nil))
+            .where("marketplace_orders.created_at > ?", 1.week.ago)
+        end.joins(marketplace: [:room])
```
rosschapman, line 15: I'm still a bit fuzzy on how authorization is wired up, but theoretically does this open access to any order by any Guest that might have any link to a recently created order?

spec/furniture/marketplace/orders_controller_request_spec.rb
```diff
@@ -24,8 +24,9 @@
     end
 
     context "when the order is not viewable by the current person" do
+      before { sign_in(space, create(:person)) }
```
rosschapman, line 27: For my learning, this `create(:person)` call is just creating a guest not associated with a space (as _member_, per se), right?

app/furniture/marketplace/order_policy.rb
zspencer: Yes, that is what happens. Order links are highly-unique; and a better solution would be to go through the Shopper.

spec/furniture/marketplace/orders_controller_request_spec.rb
zspencer: Kinda! This is creating a `Person` record in the database; which is not a `Guest` (`Guests` are unpersisted). You can see the factory definition in [spec/factories/people.rb](https://github.com/zinc-collective/convene/blob/main/spec/factories/person.rb)

If we want to create a `Person` who is `Member` of a `Space`, we would use the [`create(:membership)` factory](https://github.com/zinc-collective/convene/blob/main/spec/factories/membership.rb)

rosschapman: Oh oh I get it. You aren't trying to simulate a "guest" as I thought, because you're persisting a record. This is testing that a person cannot view another person's order. 👍🏻

zspencer:
Closed in favor of https://github.com/zinc-collective/convene/pull/1688

PR #1687, opened 2023-07-21
Removes unused instance variable in `AuthenticatedSessionMailer`

rosschapman:
Was I correct in noticing this instance var is unused (and not implicitly used anywhere magically 😅)?


anaulin (commented):
I think it is used in the email template here:
https://github.com/zinc-collective/convene/blob/5f5b0f805ac6496883ab3e57fc3dd6a2042f6dc4/app/views/authenticated_session_mailer/one_time_password_email.text.erb#L3

rosschapman:
Darn, I didn't see it. Okie dokie. Thanks for the look.

PR #1688, opened 2023-07-21
🐞🔨 `Marketplace`: `Guest` can see only their `Orders`

zspencer:
- Alternative to https://github.com/zinc-collective/convene/pull/1685

This adjusts the `OrdersController` to use the `Shopper` for the `OrderPolicy::Scope`; which ensures
when `Guests` place an order, only the device used to place the `Order` will be able to see the `Order`

Pros:
- If the `Order` url  is intercepted by a nefarious actor, it won't leak the `Shopper`s information

Cons:
- `Shopper` for `Guests` is device-dependent, so if I order something on my laptop I can't open the `Order` on my phone.

### DIFF
--- a/app/furniture/marketplace/order_policy.rb
+++ b/app/furniture/marketplace/order_policy.rb
@@ -2,12 +2,20 @@
 
 class Marketplace
   class OrderPolicy < Policy
-    class Scope < ApplicationScope
-      def resolve
-        return scope.all if person.operator?
+    class Scope
+      attr_accessor :scope, :shopper
+      delegate :person, to: :shopper
 
-        scope.joins(marketplace: [:room]).where(rooms: {space_id: person.spaces})
-          .or(scope.where(shopper: Shopper.where(person: person)))
+      def initialize(shopper, scope)
+        self.scope = scope
+        self.shopper = shopper
+      end
+
+      def resolve
+        return scope.all if person&.operator?
+
+        scope.joins(marketplace: [:room]).where(rooms: {space_id: person&.spaces})
+          .or(scope.where(shopper: shopper))
       end
     end
   end
--- a/spec/furniture/marketplace/order_policy_spec.rb
+++ b/spec/furniture/marketplace/order_policy_spec.rb
@@ -39,25 +39,30 @@ RSpec.describe Marketplace::OrderPolicy, type: :policy do
     let!(:neighbor_order) { create(:marketplace_order, marketplace: marketplace, shopper: create(:marketplace_shopper, person: neighbor)) }
 
     context "when an operator" do
-      let(:actor) { operator }
+      let(:actor) { create(:marketplace_shopper, person: operator) }
 
       it { is_expected.to contain_exactly(guest_order, neighbor_order) }
     end
 
     context "when the neighbor who placed the order" do
-      let(:actor) { neighbor }
+      let(:actor) { neighbor_order.shopper }
 
       it { is_expected.to contain_exactly(neighbor_order) }
     end
 
     context "when a guest" do
-      let(:actor) { guest }
+      before do
+        # Creates another Guest order to prove we only load their own
+        create(:marketplace_order, marketplace: marketplace, shopper: create(:marketplace_shopper))
+      end
 
-      it { is_expected.to be_empty }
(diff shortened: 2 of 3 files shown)


app/furniture/marketplace/order_policy.rb
```diff
@@ -2,12 +2,20 @@
 
 class Marketplace
   class OrderPolicy < Policy
-    class Scope < ApplicationScope
+    class Scope
+      attr_accessor :scope, :shopper
+      delegate :person, to: :shopper
+
+      def initialize(shopper, scope)
```
rosschapman, line 9: The first argument -- `shopper` in this case -- is always just `current_user`, controlled by Pundit right? (Getting more familiar with pundit.)

spec/furniture/marketplace/order_policy_spec.rb
```diff
@@ -39,25 +39,30 @@
     let!(:neighbor_order) { create(:marketplace_order, marketplace: marketplace, shopper: create(:marketplace_shopper, person: neighbor)) }
 
     context "when an operator" do
-      let(:actor) { operator }
+      let(:actor) { create(:marketplace_shopper, person: operator) }
```
rosschapman, line 42: Nice explicitness

rosschapman (approved):

app/furniture/marketplace/order_policy.rb
zspencer: Well, it's a little more nuanced. 

We `includePundit::Authorization` in our `ApplicationController` this grants us the `Pundit::Authorization#policy_for` and  `Pundit::Authorization#policy_scope` methods.

Both of these methods call `#pundit_user` when providing `Policy` and `Scope` objects with their `user`. We define `#pundit_user` in our `ApplicationController`as an alias `#current_person`; which returns either a `Person` or a `Guest`, depending on if the person on the other end of the browser has gone through the Sign In flow.

So what I've done is said "OK `OrderPolicy::Scope`... Screw `#pundit_user`, `Person`, and `Guest`! Take a `Shopper` instead!" and instantiated the `OrderPolicy::Scope` directly rather through the `policy_scope` helper method.

In my first try, I had attempted to override `#pundit_user` in the `Marketplace::OrdersController`; which failed miserably  because it started passing `Shopper`s everywhere, including the `RoomPolicy` and `RoomPolicy::Scope` that gets invoked in `ApplicationController#current_room`...

TL/DR: While there is no `current_user` in Convene; there is a `pundit_user`, which `Pundit::Authorization#policy_for` and `Pundit::Authorization::policy_scope` use; which we define in `ApplicationController` as the `current_person

anaulin: I guess this works for this specific bugfix, but I think it is a bad idea to break the expected Pundit interfaces like this. This will fall apart as soon as someone tries to do a standard `policy_scope(Order).where(...)` somewhere, and this will need to be reworked.

zspencer:
That gave me an idea! I think a better seam is to have a Marketplace::Guest that knows it’s shopper and is returned from Marketplace::Controller#current_person It lets us keep the shopper based scoping, while not creating a one off razor wire in the bathtub in our permissions codeOn Jul 21, 2023, at 20:08, Ana Ulin ***@***.***> wrote:﻿
@anaulin commented on this pull request.



In app/furniture/marketplace/order_policy.rb:
> @@ -2,12 +2,20 @@
 
 class Marketplace
   class OrderPolicy < Policy
-    class Scope < ApplicationScope
+    class Scope
+      attr_accessor :scope, :shopper
+      delegate :person, to: :shopper
+
+      def initialize(shopper, scope)

I guess this works for this specific bugfix, but I think it is a bad idea to break the expected Pundit interfaces like this. This will fall apart as soon as someone tries to do a standard policy_scope(Order).where(...) somewhere, and this will need to be reworked.

anaulin:
@zspencer I like that! (lol at "razorwire in the bathtub" 🛁 🩸)

zspencer:
Here's a try at that: https://github.com/zinc-collective/convene/pull/1692

Not quite sure how I feel about it; but it gets rid of the gross-ness of a weird `OrderPolicy::Scope`...

PR #1692, opened 2023-07-23
🧹 `Marketplace`: Use domain-specific `Person` and `Guest`

zspencer:

This is a _very weird_ PR; but I think gives us a better interface for building out features in the `Marketplace` because now a `Person` and `Guest` can know about their `Shopper`.

There's some stuff that gave me ?!?! while making this work, which I'll do my best to enumerate so we can at least discuss during the review process and reference them in some future git-blame

1. The `Marketplace::CurrentPerson` module feels very weird to me.
2. The `Marketplace::Policy::SpecFactories` really does not want to live where it lives IMO; and it feels like a bit of an obstraction... But it was also very convenient that all the specs that used it were switched to a `Marketplace::Person` and `Marketplace::Guest` at the same time...


app/furniture/marketplace/current_person.rb
```diff
@@ -0,0 +1,17 @@
+class Marketplace
+  module CurrentPerson
+    def current_person
+      super
+
+      if @current_person.is_a?(Person) || @current_person.is_a?(Guest)
+        return @marketplace_current_person = @current_person
+      end
+
+      @current_person = if @current_person.authenticated?
```
anaulin, line 10: This method is confusing to me. In which cases might a @current_person _not_ be a `Person` nor a `Guest`, but still be `authenticated?`

app/furniture/marketplace/guest.rb
```diff
@@ -0,0 +1,16 @@
...
+    attr_writer :shopper
+    def shopper
+      @shopper ||= Shopper.find_by(id: session[:guest_shopper_id])
+    end
+
+    def find_or_create_shopper
+      @shopper ||= Shopper.find_or_create_by(id: session[:guest_shopper_id] ||= SecureRandom.uuid)
+    end
+
+    private def guest_shopper_id
```
anaulin, line 12: Is this method used anywhere? I can't find any references to it.

app/furniture/marketplace/current_person.rb
```diff
@@ -0,0 +1,17 @@
+class Marketplace
+  module CurrentPerson
+    def current_person
+      super
+
+      if @current_person.is_a?(Person) || @current_person.is_a?(Guest)
+        return @marketplace_current_person = @current_person
```
anaulin, line 7: I this this instance variable being set, `@marketplace_current_person`, but I don't ever see it referenced in this PR. What is it used for?

anaulin (commented):
This feels like a lot of complexity that I'm not immediately grokking (see some initial questions below), and I don't have the spoons to think about more carefully, so I'll be abstaining from an actual review at this time.

app/furniture/marketplace/current_person.rb
zspencer: Nope, that's some leftover that should probably be removed.

app/furniture/marketplace/current_person.rb
zspencer: Yea, I think there's some... weirdness here. I'll take a quick refactor pass to get rid of some of the stumbling blocks; and we can sit on it since I think the bug is fixed already.

app/furniture/marketplace/guest.rb
zspencer: Nope, I defined it but didn't use it...

app/furniture/marketplace/current_person.rb
anaulin: That feels like a good call. Hopefully next week I'll have a bit more energy to dig into this.

app/furniture/marketplace/orders_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Marketplace
   class OrdersController < Controller
     expose :order, scope: -> { orders }, model: Order
-    expose :orders, -> { OrderPolicy::Scope.new(shopper, marketplace.orders).resolve.order(created_at: :desc) }
+    expose :orders, -> { policy_scope(marketplace.orders).order(created_at: :desc) }
```
zspencer, line 4: This is the entire point of this PR... 😭

app/furniture/marketplace/current_person.rb
zspencer: The pass is done; but it doesn't really reduce complexity more than remove cruft.

Anyway; I don't see this as urgent; so we can close it out if you don't get to it int he next week or so.

app/furniture/marketplace/orders_controller.rb
anaulin: I know. 🤗

anaulin:
@zspencer it feels like we're not feeling it on this PR. Do you still want to push it to merge, or are you ok closing this?

zspencer:
I'm just gonna close it. I'll come up with something simpler when we actually run into problems.

PR #1693, opened 2023-07-23
🧹 `Marketplace`: Move the `Policy::Factories` to `Spec`

zspencer:

It felt really weird to me to have a module calling `let` from within the `app` directory; since that's very much an `rspec`-thing.

This pulls it out to it's own place; ensures `solargraph` can find the class definition, comments it, and uses it almost everywhere `let(:marketplace)` is used.

Again, not sure it's a great / good idea...


anaulin (approved):
Somehow I had missed that we had these "shared lets" for our specs. Seems like a bad idea to me, since it obscures how is the test data set up and will create some exciting surprises for folks writing tests.

But I guess this PR makes that no worse.

anaulin:
Ah, I see now that this is merging into another pending PR, so it is not in `main` yet. I see how it makes things convenient in this one refactor, but I don't love it. That said, I won't stand in your way.

zspencer:
> Somehow I had missed that we had these "shared lets" for our specs. Seems like a bad idea to me, since it obscures how is the test data set up and will create some exciting surprises for folks writing tests.

Agree; I was like "Wait, what is this? Ehhh, I guess?"

The other option is to inline it; rather than double-down; which I think is the more prudent choice.

> Ah, I see now that this is merging into another pending PR, so it is not in main yet. I see how it makes things convenient in this one refactor, but I don't love it. That said, I won't stand in your way.

Well, it's more that the other PR adds to the shared lets; and this moves them out of the application code.

Issue #1656, opened 2023-07-13
🐞 `Marketplace`: `Shopper` and `Vendor` received duplicated `Order` `Notification`


anaulin:
* https://github.com/zinc-collective/convene/issues/1331

When a new order was placed on one of the Piikup marketplaces, both the shopper and the convene-support@ email address received the order confirmation notification twice. Only one notification should be sent to each address, however.

zspencer:
Gonna tilt at this one next, now that https://github.com/zinc-collective/convene/issues/1657 has two possible fixes in open PRs

zspencer:
Good news: I wrote a test to prove only a single `OrderPlaced` and `OrderReceived` email is sent ! Yay! Testing!

Bad news:  The test passes

So I started digging through the logs, and what appears to be happening is:

- Stripe is raising an `InsufficientBalance` error (https://zinc-collective.sentry.io/issues/4312299857/events/941add6ad6634bfd9a7cd45246c1e3ef/?project=6052513&referrer=previous-event); 
- Which is causing the `Webhook::Event` to be resent over and over again: https://dashboard.stripe.com/events/evt_1NTLcvGi42bwJzizOPg7k6EH

This indicates to me a few things:

- The transfer can't be made immediately since payments don't clear into the balance for ~3 days
- We probably should just... capture and store the stripe event rather than attempt to perform all this work at once.

Another thing I noticed is that stripe webhooks fires events for *every webhook* registered on the account; which makes me think that the place we *really* want `StripeEvents` coming in to is the `StripeUtility`, not the `Marketplace`; so that we only get one event per transaction rather than 1 events per-transaction-per-marketplace-that-has-been-connected-to-stripe.

zspencer:
OK, once #1699  gets merged, we should be able to re-test on a Space with multiple Marketplaces and confirm only one email gets sent.

PR #1699, opened 2023-07-25
🥗🐞🔨 `Marketplace`: Prevent duplicate `Order::Notification`

zspencer:
- https://github.com/zinc-collective/convene/issues/1656

Stripe sends every event to every `Stripe::Webhook` listening for it's type; which makes a lot of sense but does not play well with our current design where every Marketplace creates their own `Stripe::Webhook`.

In the short-term, discarding events for orders that do not belong to the given `Marketplace` gets us away from accidental double-mailing (or worse, double-transfering!)


anaulin (approved):
Thanks for the test and fix!

Issue #1327, opened 2023-04-08
🥔✨ `Marketplace`: `Split` `Payment` between `Vendor` and `Distributor`


zspencer:
- https://github.com/zinc-collective/convene/issues/831
- https://github.com/zinc-collective/convene/issues/1002 
- https://github.com/zinc-collective/convene/issues/1622

## Use Cases
  - [x] 🥔 `Seller` receives payment for `Order`'s `Product`s  in their Stripe Account
  - [x] https://github.com/zinc-collective/convene/issues/1510
  - [x] 🥔 `Distributor` retains `Order` Payment Processor Fees
  - [x] #1700
  
  ## Bugs
  - [ ] #1508

PR #1701, opened 2023-07-25
🧹 `Marketplace`: Cache the fee amount on the `Order`

zspencer:
- https://github.com/zinc-collective/convene/issues/1327

In order to pull the splitting of the payment to a retriable background job; we need to cache the payment processor fee information.

### DIFF
--- a/app/furniture/marketplace/order.rb
+++ b/app/furniture/marketplace/order.rb
@@ -22,6 +22,11 @@ class Marketplace
       paid: "paid"
     }
 
+    monetize :payment_processor_fee_cents
+    def vendors_share
+      product_total - payment_processor_fee
+    end
+
     def product_total
       ordered_products.sum(0, &:price_total)
     end
--- a/app/furniture/marketplace/stripe_events_controller.rb
+++ b/app/furniture/marketplace/stripe_events_controller.rb
@@ -21,7 +21,7 @@ class Marketplace
         latest_charge = Stripe::Charge.retrieve(payment_intent.latest_charge, api_key: marketplace.stripe_api_key)
         balance_transaction = Stripe::BalanceTransaction.retrieve(latest_charge.balance_transaction, api_key: marketplace.stripe_api_key)
 
-        order.update(status: :paid, placed_at: DateTime.now)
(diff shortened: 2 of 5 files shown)


zspencer:
See https://github.com/zinc-collective/convene/pull/1702 for the usage of this cached value

app/furniture/marketplace/order.rb
```diff
@@ -22,6 +22,11 @@ class Order < Record
       paid: "paid"
     }
 
+    monetize :payment_processor_fee_cents
+    def vendors_share
+      product_total - payment_processor_fee
+    end
+
```
anaulin, line 29: Should `product_total` be `monetize`d as well? It seems odd to be substracting two values, one of which is monetized and the other one isn't.

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -21,7 +21,7 @@ def create
         latest_charge = Stripe::Charge.retrieve(payment_intent.latest_charge, api_key: marketplace.stripe_api_key)
         balance_transaction = Stripe::BalanceTransaction.retrieve(latest_charge.balance_transaction, api_key: marketplace.stripe_api_key)
 
-        order.update(status: :paid, placed_at: DateTime.now)
+        order.update(status: :paid, placed_at: DateTime.now, payment_processor_fee_cents: balance_transaction.fee)
```
anaulin, line 24: Not from this PR, but I'm noticing that we do this `.update` and don't seem to be checking for success, opening us up to the possibility that this update will fail, the order won't be marked as `paid`, but notifications and other parts of the "order pipeline" would continue as if everything was fine.

One quick fix would be to throw a `!` onto the `order.update` call.

anaulin (approved):
Minor comments below.

app/furniture/marketplace/order.rb
zspencer: product_total is monetized, in that it sums the monetized product_price * quantity

app/furniture/marketplace/stripe_events_controller.rb
zspencer: Good point! Will do!

Issue #1700, opened 2023-07-25
🥔✨ `Marketplace`: `Split` `Payments` are performed automatically


zspencer:
- https://github.com/zinc-collective/convene/issues/1327

## Use Cases

- [x] 🥔 ✨ `Marketplace`: `Splits` are triggered when an `Order` is paid
- [ ] ✨ `Marketplace`: Failed `Splits` try again

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #1702, opened 2023-07-25
🥗🥔✨ `Marketplace`: Failed Automatic `Split` payments try again

zspencer:
- https://github.com/zinc-collective/convene/issues/1700

This may not be the right call, so I'm just tossing it up for conversation. In particular:

- Sidekiq does *not* guarantee once-and-only-once execution of jobs; which means if we have two workers and they both grab the job off the queue we might get two transfers. Not a good look.
- It would probably be better for us to explicitely record on our end that a Transfer has been created (or perhaps query the Stripe API?
- While automatic splitting seems like the right call; I don't know that splitting immediately upon payment is the right call. Stripe holds funds for 3 days; which indicates we should not pull it from the float until it's cleared?

Anyway, it should at least fix the bug where initial payment splits will fail if there is an insufficient balance...

### DIFF
--- a/app/furniture/marketplace/split_job.rb
+++ b/app/furniture/marketplace/split_job.rb
@@ -9,14 +9,14 @@ class Marketplace
 
     def perform(order:)
       self.order = order
-
+      order.events.create(description: "Payment Split Attempted")
       Stripe::Transfer.create({
         amount: vendors_share,
         currency: "usd",
         destination: vendor_stripe_account,
         transfer_group: order.id
       }, {api_key: stripe_api_key})
-      order.events.create(description: "Payment Split")
+      order.events.create(description: "Payment Split Completed")
     end
   end
 end
--- a/app/furniture/marketplace/split_job.rb
+++ b/app/furniture/marketplace/split_job.rb
@@ -1,0 +1,22 @@
+class Marketplace
+  class SplitJob < ApplicationJob
+    attr_accessor :order
+    delegate :vendors_share, :marketplace, to: :order
+    delegate :stripe_api_key, :vendor_stripe_account, to: :marketplace
+
+    sidekiq_options retry: 5
+    sidekiq_retry_in { |count, exception, jobhash| 1.day }
+
+    def perform(order:)
+      self.order = order
+
+      Stripe::Transfer.create({
+        amount: vendors_share,
+        currency: "usd",
+        destination: vendor_stripe_account,
+        transfer_group: order.id
+      }, {api_key: stripe_api_key})
+      order.events.create(description: "Payment Split")
+    end
+  end
+end
--- a/app/furniture/marketplace/split_job.rb
+++ b/app/furniture/marketplace/split_job.rb
@@ -10,12 +10,19 @@ class Marketplace
     def perform(order:)
       self.order = order
       order.events.create(description: "Payment Split Attempted")
-      Stripe::Transfer.create({
+      if order.stripe_transfer_id.present?
+        order.events.create(description: "Payment Split Detected Existing Transfer")
+        return
+      end
+
+      transfer = Stripe::Transfer.create({
         amount: vendors_share.to_i,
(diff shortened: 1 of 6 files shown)


zspencer:
During Wednesday session, we discussed adding an audit-log (i.e. a `Marketplace::Event` or something) prior to pulling this out of the main thread.

app/furniture/marketplace/split_job.rb
```diff
@@ -0,0 +1,22 @@
+class Marketplace
+  class SplitJob < ApplicationJob
```
zspencer, line 2: There is a world where I would have a `Marketplace::SplitPayment` class which does everything but the Sidekiqy-bits; so the database munging can be handled independently of the background jarb-bits.

This encapsulation makes the database / api bits far easier to test... But I didn't do that because we test this through both the `request` and `system` specs covering the `Marketaplace`

app/furniture/marketplace/split_job.rb
```diff
@@ -0,0 +1,22 @@
...
+    def perform(order:)
+      self.order = order
+      order.events.create(description: "Payment Split Attempted")
+      Stripe::Transfer.create({
+        amount: vendors_share.to_i,
+        currency: "usd",
+        destination: vendor_stripe_account,
+        transfer_group: order.id
+      }, {api_key: stripe_api_key})
+      order.events.create(description: "Payment Split Completed")
```
anaulin, line 19: I think we might want to have a flag on the order `split_completed` that we set to `true`, and before attempting the transfer we check that flag. It makes it easier to avoid double-transfers, and also makes it possible to find Orders with / without the split, for debugging.

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -29,14 +29,7 @@ def create
...
 
-        Stripe::Transfer.create({
-          # Leave the Stripe Fees in the `Distributor`'s account
-          amount: order.product_total.cents - balance_transaction.fee,
-          currency: "usd",
-          destination: marketplace.vendor_stripe_account,
-          transfer_group: order.id
-        }, {api_key: marketplace.stripe_api_key})
-        order.events.create(description: "Payment Split")
+        SplitJob.perform_later(order: order)
```
anaulin, line 32: Your point about Stripe holding the float for 3 days is an important one. What about setting this job to perform in 3 or 4 days? That way if the order gets cancelled or reversed or whatever, we don't do the transfer later.

```suggestion
        SplitJob.set(wait: 4.days).perform_later(order: order)
```

app/furniture/marketplace/split_job.rb
anaulin: Are there any other conditions we should check here before trying to transfer the split? For example, check that the order status is still `paid`.

anaulin (approved):
This looks ok, but I would prefer adding a bit more guardrails to make sure that we don't double-split (see my comment below about adding a flag to mark an order as already-split).

app/furniture/marketplace/split_job.rb
zspencer: I am a bit hesitant to add checking for `paid?` since we could add `delivered` or `ready_for_pickup` or `in_transit` all of which would no longer count as `paid?`...

But I do think we should have a flag of some kind. I went with an id-and-also-flag since keeping the context of which transfer did the split doesn't seem like a bad idea?

app/furniture/marketplace/stripe_events_controller.rb
zspencer: I'm torn. On the one hand, waiting 3~4 days is not a bad idea since we don't want to transfer money out if we don't have it... On the other, forcing them to wait 7~10 days to receive payment for their goods feels not-great; since it will take 3~4 days from transfer to enter their accounts...

For now; I'd prefer to just try immediately; but I could be convinced otherwise.

Issue #1691, opened 2023-07-22
🐞 `Marketplace`: order history for Marketplace managers contains incomplete orders


anaulin:
I noticed in the production Piikup marketplace that the Order History is showing every single shopping cart that was ever created for that Marketplace:
![image](https://github.com/zinc-collective/convene/assets/6729309/a55dc94a-fd5a-4d8e-95e0-4e7dc3ba2bdd)

I think this Order History page for the Marketplace should only show orders that were completed (paid). (Maybe one day we want to show "abandoned orders" somewhere else, but that would be a separate view.)

zspencer:
Yikes; would you prefer we fix this by:

- Filtering off of current data
- Remucking the domain model so that an `Order` and a `Cart` are separate
- Using Rails built-in Polymorphism to make sure `Order` and `Cart` are queried indepenetly?

anaulin:
I think an immediate fix to show only "paid" orders is a good idea.

Mid/long-term we might re-think the domain model, but we shouldn't block fixing this bug on that re-think.

zspencer:
Alright, I'll do a quick fix for filtering by paid!

zspencer:
Bam: https://github.com/zinc-collective/convene/pull/1703

PR #1703, opened 2023-07-25
🐞🔨 `Marketplace`: Only list `paid` `Orders`

zspencer:
- https://github.com/zinc-collective/convene/issues/1691

This just makes sure the orders index isn't overwhelming with all the un-checked-out-carts


anaulin (approved):
Thanks for the quick fix!

Issue #1660, opened 2023-07-13
🐞 `Marketplace`: order time is displayed in UTC


anaulin:
- #1331 

This order was placed at 2:15 AM local time, but in Convene it shows as:
![image](https://github.com/zinc-collective/convene/assets/6729309/d92ca546-065c-424c-a93c-605e078496cc)

We should try to display this timestamp in the user's timezone (as inferred from the browser).

zspencer:
Do you have thoughts on how we can determine the appropriate time zone? Should we just presume PT as the application default for now? Or should we let it be set on the `Person` and `Space` level?

anaulin:
I was thinking we could use something like https://github.com/kbaum/browser-timezone-rails to set the timezone for each request.

anaulin:
Also, we could one or all of:
* let a Space configure a default timezone
* let a Person configure their default timezone (and fallback to their browser)

rosschapman:
Converting the value automatically using the browser timezone seems sufficient to me unless Piikup mentioned they wanted configurability to handle some use case? 

I tend to like an abstraction around 3rd-party lib helpers like `humanized_money_with_symbol`. Like, creating our own ViewComponent called `DateTime` or something. It cold parameterize TZ in case it does become dynamically set later.

anaulin:
@rosschapman agreed on starting simple, since nobody has asked for anything special, to the best of my knowledge.

I also like the idea of encapsulating the whole thing, although a helper method might be more lightweight than a ViewComponent, just so that in views we can do something like:
```
Order placed at <%= in_user_tz(date: date) %>
```

instead of:
```
Order placed at <%= render DateComponent.new(date: date) %>
```

Although if we use something like the gem I suggested above (which is used by one of the clients I work for), then we don't need to do anything special in views, because it sets the timezone on a per-request basis at the Rails level, so when you display a date in the Rails timezone it shows up in the correct timezone.

zspencer:
I believe [`localize(date, :format)`](http://api.rubyonrails.org/classes/ActionView/Helpers/TranslationHelper.html#method-i-localize) is what you're looking for from a "how to format dates?" perspective; as this calls [`#to_formatted_s`](http://api.rubyonrails.org/classes/DateTime.html#method-i-to_formatted_s) under the covers.

[...]

PR #1706, opened 2023-07-26
🐞🔨 `Marketplace`: Use the default `TimeZone` when showing dates and times

zspencer:
- https://github.com/zinc-collective/convene/issues/1660

This is a small step towards a better way for handling time; since we currently only operate out of the SF Bay Area; it seems like a reasonable place to start.

At some point we'll want to add in time zone support for browsers, users, spaces, etc. but this gets us to a place where things are a little better sooner rather than later

## Before
Screenshot 2023-07-26 at 12 46 01 PM
Screenshot 2023-07-26 at 12 45 53 PM

## After
Screenshot 2023-07-26 at 12 46 27 PM
Screenshot 2023-07-26 at 12 46 10 PM


anaulin (approved):
Thank you! ⏲️

PR #1684, opened 2023-07-20
🥔✨ `Journal`: `Entries` link to their `Keywords`

zspencer:
- https://github.com/zinc-collective/convene/issues/1566
- https://github.com/zinc-collective/convene/issues/1662

Screenshot 2023-07-20 at 12 41 16 PM

PR #1708, opened 2023-07-27
🧹 `Journal`: Rename the method that turns Keywords into Links

zspencer:
- https://github.com/zinc-collective/convene/pull/1684#discussion_r1270924805

Better name thanks to @RossChapman


anaulin (approved):
🔗

PR #1710, opened 2023-07-27
🥔✨ `Marketplace`: Record `Order` and `Cart` events

zspencer:

Start recording an audit trail of some of the key events, so that we can review what happened when.

Starting with:

- Cart Created
- Checkout Started
- Payment Received
- Notifications to Vendor and Distributor Sent
- Notification to Buyer Sent
- Payment Split


anaulin (approved):
🪵

PR #1712, opened 2023-07-27
🌸🥗🥔✨ `Marketplace`: `Orders` expose their `Events`

zspencer:

Next steps:

- [x] Tidy the UI! It gross
- [x] Add Specs! It needs them!

### After?

Screenshot 2023-08-09 at 4 52 08 PM


app/furniture/marketplace/orders_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Marketplace
   class OrdersController < Controller
-    expose :order, scope: -> { orders }, model: Order
+    expose :order, scope: -> { orders.includes(:events, ordered_products: [:products]) }, model: Order
```
zspencer, line 3: I am writing N+1 queries 😢. This gets rid of them.

app/furniture/marketplace/orders/_order.html.erb
```diff
@@ -26,4 +26,18 @@
   <hr class="col-span-6 mb-2" />
   <div class="col-span-4">Total</div>
   <div class="col-span-2 text-right"><%= humanized_money_with_symbol(order.price_total) %></div>
+
+  <%- if policy(order.events).index? %>
```
zspencer, line 30: Do we even need this?

app/furniture/marketplace/orders_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Marketplace
   class OrdersController < Controller
-    expose :order, scope: -> { orders }, model: Order
+    expose :order, scope: -> { orders.includes(:events, ordered_products: [:product]) }, model: Order
```
zspencer, line 3: Lilliputian Zee would do `events` independently as part of this PR, and then the `ordered_products` after; but Gilligan Zee is just stomp stomp stomping on the orders

daltonrpruitt (approved):

app/furniture/marketplace/orders/_order.html.erb
rosschapman: Because the user is already authorized at the Order level?

app/furniture/marketplace/orders/_order.html.erb
```diff
@@ -26,4 +26,18 @@
...
+    <hr class="col-span-6 my-4" />
+    <div class="col-span-6 text-xs">
+
+      <h3 class="mb-2 text-sm font-bold">Events</h3>
```
rosschapman, line 34: Is "Events" a friendly term? What about "Timeline"?

app/furniture/marketplace/orders/_order.html.erb
zspencer: More that `Policy#index?` returns "true" by default, so it feels kind of extraneous....

app/furniture/marketplace/orders/_order.html.erb
zspencer: I like that!

Issue #1714, opened 2023-07-27
🐞 `Journal`: `Entry` with overlapping `Keywords` ignore the longer `Keyword`


zspencer:
- https://github.com/zinc-collective/convene/issues/1713

For https://slipvector.games, I noticed that one of the worldbuilding notes was not correctly extracting keywords.

```markdown
\#SarinsGarden is a large #WorldGarden on the planet #Sarin in the #DarnikSystem. 
```

<img width="431" alt="Screenshot 2023-07-27 at 1 52 26 PM" src="https://github.com/zinc-collective/convene/assets/50284/5934b194-d083-4fef-8bf3-349a5551fac0">

I would expect `#SarinsGarden` to be it's own keyword.

zspencer:
Fixed!

PR #1715, opened 2023-07-27
🐞🔨 `Journal`: `Entry` supports overlapping `Keywords`

zspencer:
- https://github.com/zinc-collective/convene/issues/1714

Turns out, it was a renderer-layer thing; where the shorter `Keyword` was getting linkfied prior to the longer `Keyword`, and when I switched to linkify the longer `Keyword` first it would double-down on the links putting a link to the shorter keyword within the longer one!

So now I'm matching the `#` as part of the `Keyword`, and replacing it with the html entity for `#`, (aka `&#35;`)

This makes it so when we linkify keywords long-to-short; the longer keywords no longer contain the shorter keyword; eliminating the nested `a` tags.

### DIFF
--- a/app/furniture/journal/entry_component.rb
+++ b/app/furniture/journal/entry_component.rb
@@ -28,7 +28,7 @@ class Journal
 
     # We sort the variants longest to shortest because regex matches groups left-to-right
     private def keywords_regex(keyword)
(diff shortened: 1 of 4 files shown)


zspencer:
I want my bug fixxxeeeeed so I am merging without review. YOLO!

Still curious about any questions that come-up from review by @zinc-collective/convene-contributors!

PR #1717, opened 2023-07-28
🧹 `Marketplace`: Cause Stripe to Retry if we can't mark an `Order` as `paid`

zspencer:
- Recommended by @anaulin https://github.com/zinc-collective/convene/pull/1701#discussion_r1275566331


anaulin (approved):
👍🏼

PR #1720, opened 2023-07-28
🥗✨ `Journal`: `Keyword`s can be merged into one another

zspencer:
- https://github.com/zinc-collective/convene/issues/1713

There's a number of things that can and will go wrong with this code eventually, including (but not limited to)

1. What happens when a `Keyword` has 1000s of entries?!
2. If this happens on-request-thread it will probably be very very slow; but if we do it *off* request-thread; the `Keyword#show` UI will be out-of-date; and we haven't implemented `ActionCable` yet

But hey, at least I will have a Heroku-CLI affordance for merging some of the `Keywords` that are synonyms...

I'll probably want to build out a WUI as well; but I wanted to get the basic logic written down and tested before I tilted at the WUI.

### DIFF
--- a/spec/furniture/journal/keyword_spec.rb
+++ b/spec/furniture/journal/keyword_spec.rb
@@ -8,19 +8,31 @@ RSpec.describe Journal::Keyword, type: :model do
   it { is_expected.to belong_to(:journal).inverse_of(:keywords) }
 
   describe "#merge" do
+    let(:journal) { create(:journal) }
+    let(:entry) { create(:journal_entry, body: "#GoodTime and #GoodTimes", journal: journal) }
+    let(:other_entry) { create(:journal_entry, body: "#GoodTime", journal: journal) }
+    let(:good_time_keyword) do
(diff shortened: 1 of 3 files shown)


spec/furniture/journal/keyword_spec.rb
```diff
@@ -7,6 +7,24 @@
   it { is_expected.to validate_uniqueness_of(:canonical_keyword).case_insensitive.scoped_to(:journal_id) }
   it { is_expected.to belong_to(:journal).inverse_of(:keywords) }
 
+  describe "#merge" do
+    it "adds aliases for the other `Keywords` canonical and aliases, deletes the other Keyword, and re-detects the entries keywords" do
```
anaulin, line 11: This might be overly paranoid, but I wonder if it would be worth adding an assertion that if the `from` keyword is associated to any entries, those don't get deleted when the keyword is destroyed.

anaulin (approved):

spec/furniture/journal/keyword_spec.rb
zspencer: Not too paranoid!

Issue #1658, opened 2023-07-13
🐞 `Space`: 404 pages should give people a way back to the `Space` they are in


anaulin:
- https://github.com/zinc-collective/convene/issues/1154

When a person lands on a page that they don't have access to, or that doesn't exist, we are showing a Rails-standard 404 page, which looks like this:
![image](https://github.com/zinc-collective/convene/assets/6729309/22d000a6-c525-40d8-a579-bb0f763adcf7)

We should replace this standard page with something that:
* shows the Convene top/bottom navigation elements, so that people can try navigating back to what they were doing
* in addition to the "not found" text, have some text about "maybe you need to log in?" if they are logged out

anaulin:
I'm taking a look at this.

PR #1727, opened 2023-08-03
🥔✨ `Neighborhood`: Error pages use the Convene layout

anaulin:
- https://github.com/zinc-collective/convene/issues/1658

Fixes https://github.com/zinc-collective/convene/issues/1658 by adding basic error pages to replacethe standard Rails error pages. These new pages will show the standard Convene UI chrome, and will include things like the Space name in the breadcrumbs, if one is available.

It's pretty barebones for now, but at least it gives a way back to the app.

It also proved surprisingly hard to test this controller in a request or Capybara-based test, so instead I wrapped the less straightforward part of this code in a `rescue` block.

#### 404 error
image

#### Some other error
image

#### 404 error
image

#### Some other error
image

For my and others' reference, to test this locally you need to set `config.consider_all_requests_local = false` in your `development.rb`.


anaulin:
There is some wonky interaction happening between my custom error pages and that one spec, sigh.

zspencer (approved):
This is great! Thank you!

anaulin:
Thank you for helping push this over the finish line, @zspencer !

zspencer:
cats-teamwork

PR #1729, opened 2023-08-03
Adds .gitignore directives for vscode users

rosschapman:


zspencer:
Hey @rosschapman, 

This looks like it will ignore changes for the shared ltex configuration as well. 

I think there's a set of things that we do want to share; and some things we don't in this folder. My gut says we should start with sharing and then prune out what is not working well to share?

Thoughts?

rosschapman:
@zspencer makes sense. I didn't notice those ltex files!

.gitignore
```diff
@@ -31,3 +31,6 @@ app/assets/builds/
 .yardoc
 
 .devcontainer/output
+
+# Bust A Gem VSCode extension
+TAGS
```
rosschapman, line 36: I really like this extension which allows _go to definition_ for symbols/etc... for bundled gems. It generates a ctags file locally.

.vscode/launch.json
rosschapman, line 1: @zspencer In projects I've worked on we don't typically share launch configs from the `.vscode` dir, but rather put these in documentation. But I wouldn't be against it either if this is a common Ruby/Rails vscode extension.

rosschapman, line 1: @zspencer In projects I've worked on we don't typically share launch configs from the `.vscode` dir, but rather put these in documentation. But I wouldn't be against it either if this is a common Ruby/Rails vscode extension.

.gitignore
zspencer: Nice! I use solargraph which is able to leverage the Gems yardoc; but we should definitely support folks who prefer CTAGS

.vscode/launch.json
zspencer: Normally this is the kind of thing I would be like "Ehh, keep it local; keep it safe" as well. Especially if the launch configs are machine-specific (like, it has absolute paths on a system directory)

zspencer (approved):
Ship it!

Issue #1500, opened 2023-05-25
🌱✨ `Marketplace`: `Seller` receives `Order` in their `Square` Point of Sale


zspencer:
- https://github.com/zinc-collective/convene/issues/1331
- [Engineering Document](https://docs.google.com/document/d/1Z-d1i5AEeFPDS44vyWSESd1p-1hFEuDAjsXZuhKzn1Y/edit)

## Bounties:
- $109.52 from https://github.com/zinc-collective/convene/issues/832

## Requested By
- https://github.com/zinc-collective/convene/issues/832 on behalf of [Crumble and Whisk](http://crumbleandwhisk.com/)

## Prerequisites
- [ ] https://github.com/zinc-collective/convene/issues/1501

zspencer:
This was requested by Piikup on behalf of [Crumble and Whisk](http://crumbleandwhisk.com/)

rosschapman:
Spec document: https://docs.google.com/document/d/1Z-d1i5AEeFPDS44vyWSESd1p-1hFEuDAjsXZuhKzn1Y. Ask @rosschapman for access.

zspencer:
- Bounty created by https://github.com/zinc-collective/convene/issues/832 

Invoice: https://app.qbo.intuit.com/app/invoice?txnId=588

rosschapman:
Things are moving along! Allow me to share a sneak peak of a Square order programmatically created from a `marketplace_order` in my local Convene environment in response to a simulated _checkout workflow_ manually triggered by a `checkout.session.completed` webhook event generated by the Stripe CLI.  

<img width="1275" alt="image" src="https://github.com/zinc-collective/convene/assets/5185/d5b5d0d2-8dab-42fb-a6e1-d6d5626f10aa">

anaulin:
Exciting! 🎉

rosschapman:
🚧🚧🚧WIP🚧🚧🚧

Recap from @rosschapman's  8/15/23 ride-along with Piikup:

Firstly, it should be noted that it was atypical for April to be driving the cargo truck. Normally she would be running dispatch from home/office. I'd like to first thank her and the Piikup team for accommodating us! April  displayed a heroic effort in multi-tasking driving, executing a pickup route with multilpe, talking to me and keeping me informed along the way, continuing to manage dispatch, and handling other impromptu business issues during the ride-along.

We spent nearly 2 hours together and drove a somewhat typical route: 
[...]

PR #1731, opened 2023-08-04
🥔✨ `Marketplace`:  Encrypt and Store Square Access Token

rosschapman:
- https://github.com/zinc-collective/convene/issues/1735

I borrowed this pattern for storing encrypted data from `Utility`, and I think it works well for Furniture-level private keys as well. In order not to conflict with the more general `settings` attribute, I opted for `secrets`.


rosschapman:
@zspencer @anaulin The rspec test seems to be failing because the new migration is not applied in the `Setup Rails` CI step. At least, in what I assume is the list of migration IDs that get dumped to output, the last one is `20230727010141` which precedes mine. 

Does this seem abnormal?

zspencer (approved):
This looks great! I did some digging and [figured out *why* it was not working](https://github.com/zinc-collective/convene/pull/1734).

The TL/DR  is that `RankedModel` uses `Furniture.select(:id, :room_id, :slot)` under the covers at some point; which retrieves only those three attributes.

By adding an `if has_attribute?("secrets_ciphertext")` we skip setting the default in cass where the secrets are not loaded.

rosschapman:
Thanks for the assist @zspencer. Whew. I'm now more well-read on Ranked model, too.

app/models/furniture.rb
```diff
@@ -24,6 +24,20 @@ class Furniture < ApplicationRecord
 
   validates :furniture_kind, presence: true
 
+  # Used by child classes that require storage of secrets, like Square API keys
+  # for a Marketplace
+  has_encrypted :secrets, type: :json
+
+  # Forces consistent access, rather than having to work with a nil value
+  after_initialize do
```
rosschapman, line 32:
```suggestion
  # Setting `secrets` to an empty hash after initialization will force consistent access for developers, rather than having to code around a potential nil value
  after_initialize do
```

app/models/furniture.rb
```diff
@@ -24,6 +24,20 @@ class Furniture < ApplicationRecord
...
+  has_encrypted :secrets, type: :json
+
+  # Setting `secrets` to an empty hash after initialization will force consistent access for developers, rather than having to code around a potential nil value
+  after_initialize do
+    # The `RankedModel` gem uses
+    #   [`ActiveRecord::QueryMethods.select`](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-select)
+    # to do some stuff (Validation?) under the covers while keeping memory usage light... Which is good!
+    # But it also means that we don't have the ciphertext!
+    # So we can't set the secrets in that case (nor would we want to)
+    self.secrets ||= {} if has_attribute?("secrets_ciphertext")
```
rosschapman, line 38: @zspencer Makes me think this will not be the last time we try to access data on models returned ranked queries with surprising results 🤣

PR #1733, opened 2023-08-05
🥔✨👷 `Marketplace`: Transmits orders to `Square`

rosschapman:

## TDR (Technical Decision Records)
- Put new Square logic on `Marketplace::Order`. See https://github.com/zinc-collective/convene/pull/1733#discussion_r1289360021. 
- Won't implement an ersatz feature flag. See https://github.com/zinc-collective/convene/pull/1733/files#r1302266674

### Local QA Trials
- [x] 8/24/23 - Created multi-product, multi-quantity order 
  image

- [x] 8/23/23 - Successfully created a  Square order and payment with the correct vendor share price total with the Stripe processing fee subtracted:
    image

- [x] 8/18/23 - Successfully created a  Square order and payment by invoking the call sequence in with this Stripe CLI command: `$ stripe trigger checkout.session.completed`. Order shoes in my Sandbox Seller Dashboard: 

	image


app/furniture/marketplace/marketplace.rb
```diff
@@ -22,6 +22,9 @@ class Marketplace < Furniture
     setting :stripe_webhook_endpoint
     setting :stripe_webhook_endpoint_secret
 
+    # Square order notifications integration
+    setting :square_location_id
```
zspencer, line 26: Is `location` a term-of-art in Square land?

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -24,6 +24,19 @@ def create
...
+        )
+
+        order_body = build_square_create_order_body(order)
+        square_client.orders.create_order(order_body)
+        payment_body = build_square_create_order_payment_body(order, marketplace, balance_transaction)
+
+        # DEV NOTE: Square requires that orders are paid in order to show up in the Seller Dashboard
+        square_client.payments.create_payment(payment_body)
+
+        order_to_square_order
```
zspencer, line 39: This controller is not a great example; because does way too much and all that stuff is Very Important.

Would you be available to extract this to a method on the `Order` class for now? (Or, alternatively, a `Marketplace::SquareOrder` or `Marketplace::Order::InSquare` class?)

zspencer (commented):
This looks like it's on the right track! I also added [Support for Bust a Gem](https://github.com/zinc-collective/convene/pull/1744) so that you don't have to manually un-stage your `TAGS` file.

app/furniture/marketplace/marketplace.rb
rosschapman: Yes. Business _location_ is a key aspect of a merchant account. It's required for account activation. There's a good overview here: https://developer.squareup.com/docs/locations-api.

app/furniture/marketplace/stripe_events_controller.rb
rosschapman: Definitely. I was prepared to go in that route after sketching out this call sequence. There's a few ways I can think of to create a better abstraction here. Allow me to enumerate out loud as a reference: 

1. Move logic to the existing `Marketplace::Order` model (as you suggest)
1. Move logic to a new model/class to encapsulate the `SquareOrder` behavior (as you suggest)
1. Similar to the above, move logic to a new single-purpose `SquareOrderCreateService` "service object" type class which executes this workflow and could be invoked like `SquareOrderCreateService.new(...).call` (inspired by some recent reading I've done at https://jardo.dev/rails-service-objects)
1. (More avant-garde for the moment): Rejiggle this controller to pass the event on to  an marketplace-level event bus (implemented as a _class_ with similar case/switch reducer) and inline the logic (or new class call) there.

At this nascent stage I like the idea of moving some business logic onto the existing `Marketplace::Order` class, and continue monitoring what types of behavior accumulate there. Rather than give into the seduction of decoupling prematurely. Then let's see how the resulting _specs as documentation_ turn out. Will it be easy to test? Will the tests feel like a satisfactory observation of the behavior in action with the model as the subject?

.env.example
```diff
@@ -31,3 +31,6 @@ OPERATOR_API_KEY=a-secret-that-should-get-changed
 STRIPE_API_KEY=get-this-from-your-stripe-account
 
 NEIGHBORHOOD_TIME_ZONE="Pacific Time (US & Canada)"
+
+# Configure Square environment
+# SQUARE_ENV="production"|"sandbox"
```
zspencer, line 36: This probably should be set in the Marketplace for now, rather than a global variable.

app/furniture/marketplace/order.rb
```diff
@@ -52,5 +52,126 @@ def marketplace_name
...
+    def send_to_square_seller_dashboard(stripe_balance_transaction)
+      square_create_order_response = create_square_order
+      square_order_id = square_create_order_response.body.order[:id]
+      create_square_order_payment(square_order_id)
+
+      # TODO: What to return here?
+      true
+    end
+
+    def create_square_order
```
zspencer, line 72: Can this be a private method since it's only called from the `send_to_square_seller_dashboard` method?

app/furniture/marketplace/order.rb
```diff
@@ -52,5 +52,126 @@ def marketplace_name
...
+    def square_client
+      environment = ENV.fetch("SQUARE_ENV")
+      access_token = marketplace.square_access_token
+
+      @square_client ||= Square::Client.new(access_token:, environment:)
+    end
+
+    def send_to_square_seller_dashboard(stripe_balance_transaction)
+      square_create_order_response = create_square_order
+      square_order_id = square_create_order_response.body.order[:id]
```
zspencer, line 65: I would probably inline this assignment as well as the ones on 57 and 58; since they don't add anything directly.

app/furniture/marketplace/order.rb
```diff
@@ -52,5 +52,126 @@ def marketplace_name
...
+      create_square_order_payment(square_order_id)
+
+      # TODO: What to return here?
+      true
```
zspencer, line 69: True is fine?

app/furniture/marketplace/order.rb
```diff
@@ -52,5 +52,126 @@ def marketplace_name
...
+    # NOTE: Square requires that orders are paid in order to show up in the Seller
+    # Dashboard
+    def create_square_order_payment(square_order_id)
+      square_location_id = marketplace.settings["square_location_id"]
+      space_id = marketplace.space.id
+      square_create_payment_body = build_square_create_order_payment_body(
+        square_order_id,
+        square_location_id,
+        space_id,
+        # TODO: what should be canonical pricing data here: stripe_balance_transaction or order data?
```
zspencer, line 86: I would expect this to be the `sellers_share` of the order; not the full balance or order amount.

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -24,14 +24,20 @@ def create
         order.update!(status: :paid, placed_at: DateTime.now, payment_processor_fee_cents: balance_transaction.fee)
         order.events.create(description: "Payment Received")
 
+        # TODO: Does this suffice as a feature flag for the time being?
+        # TODO: Need better conditional abstraction here?
+        if marketplace.square_location_id? && Rails.env.development?
+          order.send_to_square_seller_dashboard(balance_transaction)
+        end
```
zspencer, line 31: I don't know that we need to hide it behind a feature flag

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -24,14 +24,20 @@ def create
...
+
         Order::ReceivedMailer.notification(order).deliver_later
         order.events.create(description: "Notifications to Vendor and Distributor Sent")
         Order::PlacedMailer.notification(order).deliver_later
         order.events.create(description: "Notification to Buyer Sent")
 
         SplitJob.perform_later(order: order)
       else
-        raise UnexpectedStripeEventTypeError, event.type
+        # raise UnexpectedStripeEventTypeError, event.type
```
zspencer, line 40: This probably should still be there, no?

.vscode/launch.json
```diff
@@ -4,14 +4,6 @@
...
   "version": "0.2.0",
   "configurations": [
-    {
-      "type": "rdbg",
-      "name": "Debug current file with rdbg",
-      "request": "launch",
-      "script": "${file}",
-      "args": [],
-      "askParameters": true
-    },
```
zspencer, line 14: Is removing this part of doing Square stuff? Or is it incidental?

zspencer (commented):
It's on a good path; I would be happy for us to merge this if:

- We get rid of the global variable storing sandbox vs production
- It did not include the TAGS or the change to launch.json
- It had some amount of tests. I'm happy to pair on writing the tests; or do a branch-off-your-branch and submit a patch with some tests.

app/furniture/marketplace/stripe_events_controller.rb
rosschapman: Oops yes.

app/furniture/marketplace/order.rb
rosschapman: Definitely! And other methods as well. `send_to_square_seller_dashboard` was meant to be the only interface exposed to consumers.

.vscode/launch.json
rosschapman: Oh not sure what happened here. Oops.

app/furniture/marketplace/marketplace.rb
```diff
@@ -91,5 +94,13 @@ def create_stripe_webhook_endpoint(webhook_url:)
...
       @_model_name ||= ActiveModel::Name.new(self, ::Marketplace)
     end
+
+    def square_environment
+      if Rails.env.development?
+        "sandbox"
+      end
+
+      "production"
+    end
```
rosschapman, line 104:
```suggestion
      if Rails.env.production?
        "production"
      end

      "sandbox"
    end
```

app/furniture/marketplace/stripe_events_controller.rb
rosschapman: Agree, I'll write a boolean for this.

app/furniture/marketplace/order.rb
```diff
@@ -52,5 +52,135 @@ def marketplace_name
...
+
+    # NOTE: Square requires that orders include fulfillments in order to show up
+    # in the Seller Dashboard
+    # See: https://developer.squareup.com/docs/orders-api/create-orders
+    private def build_square_create_order_body(marketplace)
+      idempotency_key = "#{id}_#{8.times.map { rand(10) }.join}"
+      location_id = marketplace.settings["square_location_id"]
+      customer_id = shopper.id
+
+      stripe_fee_per_line_item = (payment_processor_fee_cents / ordered_products.sum(&:quantity))
```
rosschapman, line 100: @zspencer @anaulin I can't tell if this is tacky or not, but it quickly felt like the best way to make sure line items are separated in the Square order and the _vendor share_ total is accurately calculated.

app/furniture/marketplace/order.rb
```diff
@@ -52,5 +52,135 @@ def marketplace_name
...
+
+      line_items = ordered_products.map { |ordered_product|
+        {
+          name: ordered_product.name,
+          quantity: ordered_product.quantity.to_s,
+          item_type: "ITEM", # ITEM|CUSTOM_AMOUNT|CGI
+          base_price_money: {
+            amount: (ordered_product.price_cents - stripe_fee_per_line_item),
```
rosschapman, line 108: See: https://github.com/zinc-collective/convene/pull/1733/files#r1302335489

app/furniture/marketplace/marketplace.rb
```diff
@@ -91,5 +94,17 @@ def create_stripe_webhook_endpoint(webhook_url:)
     def self.model_name
       @_model_name ||= ActiveModel::Name.new(self, ::Marketplace)
     end
+
+    def square_environment
+      if Rails.env.production?
+        "production"
+      end
```
zspencer, line 101: I would prefer if this was a value stored in the `Marketplace`; since we do need to be able to use the sandbox environment in a Rails production context.

Even if we don't make it to the UI; it does not seem appropriate to flag based upon deployment environment.

app/furniture/marketplace/order.rb
zspencer: This calculation feels like a dangerous idea from a book-keeping perspective. They are selling the product for $10, not $10 - a_fraction_of_the_stripe_fees.

app/furniture/marketplace/order.rb
```diff
@@ -52,5 +52,136 @@ def marketplace_name
...
+      # This data is intended for use in debugging, etc... until we further
+      # the Square integration productize
+      {
+        order_id: square_create_order_response.body.order[:id],
+        payment_id: square_create_payment_response.body.payment[:id]
+      }
+    end
+
+    private def square_idemp_key
+      "#{id}_#{8.times.map { rand(10) }.join}"
```
zspencer, line 73: Does Square want this to be per-order impotency or is it per-api request impotency?

zspencer (commented):
This is looking pretty good!

A few things:

1. Why is `TAGS` checked in here?
2. My gut says we should be passing the full amounts directly through; and not doing any kind of fee-deductions. The PoS doesn't need to know about payment-processor fees; just how much money we collected on their behalf. It's not like a sandwhich that costs $10 is registered as sold at "$9.08" by the PoS because someone used MasterCard and $9.15 because someone else used Visa...

That said, I am down to approve this once the TAGS are deleted because I could be 100% wrong and I would rather we move towards prod and learn.

app/furniture/marketplace/order.rb
rosschapman: Yeah I figured as much. There's gotta be a better way around this. Square seems to compute and validate the line item total generated during order creation against the payment total during payment creation. If they don't match, the payment fails.

The trick is to have the vendor/seller line items appear with the correct amounts. Need to think on this more. It could be we add a negative service charge with the order creation data for the Stripe fee. But I'm not sure sellers are supposed to be aware of the Stripe fee?

app/furniture/marketplace/order.rb
rosschapman: Per request.

app/furniture/marketplace/marketplace.rb
rosschapman: I may have misinterpreted your earlier feedback as _put on marketplace model_ for some reason. Oof, tough getting used to jumping in and out of a codebase on limited time!

zspencer: Well, this is definitely the right *place* for it in our current domain model; it's just not quite how I would decide whether it's prod or not.

And yea, the "put work down and  pick it back up in narrow time windows" is a hard skill!

app/furniture/marketplace/order.rb
```diff
@@ -52,5 +52,133 @@ def marketplace_name
     def price_total
       [product_total, delivery_fee, tax_total].compact.sum
     end
+
+    def square_client
```
rosschapman, line 56:
```suggestion
	# TODO: Add `square_environment` attribute to database/model
    def square_client
```

app/furniture/marketplace/order.rb
zspencer: Can we tell Square that the Payment is for the total?

app/furniture/marketplace/order.rb
```diff
@@ -52,5 +52,135 @@ def marketplace_name
     def price_total
       [product_total, delivery_fee, tax_total].compact.sum
     end
+
+    # TODO: Add `square_environment` attribute to database/model
+    def square_client
+      @square_client ||= Square::Client.new(access_token: marketplace.square_access_token, environment: marketplace.square_environment)
```
zspencer, line 58: Since we're reading data from the `marketplace` to create the `square_client`, that implies to me that it maybe wants to live on the `marketplace`, and the `Order` class can `delegate :square_client, to: :marketplace`?

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -24,6 +23,10 @@ def create
         order.update!(status: :paid, placed_at: DateTime.now, payment_processor_fee_cents: balance_transaction.fee)
         order.events.create(description: "Payment Received")
 
+        if marketplace.square_order_notifications_enabled?
+          order.send_to_square_seller_dashboard
```
zspencer, line 27: Thoughts on encapsulating this conditional within the method? Basically, `Order#send_to_square_seller_dashboard` is a no-op if `marketplace.square_order_notifications_enabled?` is false.

PR #1742, opened 2023-08-08
🧹 `Neighborhood`: Apparently `MiniTest` is no more

zspencer:

I noticed this Sunday with Dalton, but don't know if I actually made any PRs with it...

Anyway, apparently the `MiniTest` namespace is no more; and it is only `Minitest`. It had been deprecated for a while but I am old and set in my ways.


anaulin (approved):
Weird, I don't think I ever noticed the switchover. tomaTo poTato.

PR #1744, opened 2023-08-09
🥔✨ `Contributor`: Support `Bust A Gem` VSCode Extension

zspencer:
- https://github.com/zinc-collective/convene/issues/1737

[Bust A Gem](https://marketplace.visualstudio.com/items?itemName=gurgeous.bust-a-gem) was recommended by @RossChapman as a way to provide jump-to-definition support for gems.

It seems neat, and seems worth adding to our recommended extensions / `.gitignore` so that folks who want/like to use it can do so.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ app/assets/builds/
 .yardoc
 
 .devcontainer/output
+
+TAGS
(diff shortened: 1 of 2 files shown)


anaulin (approved):
🏷️

rosschapman:
Nice, thanks! I'm actually finding it hard to get Solargraph working on my machine. Which means it's good to have options.

anaulin:
I've never been able to get Solargraph working well. Including once during an ensemble with Zee (unpaid Solargraph salesperson) to help me configure it properly. 🤷🏼‍♀️

zspencer:
It's not been working that great for me recently either. There appears to be fragmentation in Ruby Language Server land; for better or for worse.

PR #1755, opened 2023-08-14
🐞🔨 `Journal`: Fix `Keywords#show` erroring instead of 404ing

zspencer:
- https://github.com/zinc-collective/convene/issues/1754

This makes `Keyword`  browsing a bit tighter; in that:

- It's more generous in accepting aliases keywords
- It redirects to the Canonical location for the Keyword
- It 404s when a Keyword is not found.


anaulin (approved):
Nice.

PR #1756, opened 2023-08-14
`Neighborhood`: Handle `RecordNotFound` exceptions

zspencer:
- https://github.com/zinc-collective/convene/issues/1754

Apparently we never made it so our `ApplicationController` rescued from `ActiveRecord::RecordNotFound` errors.

Now we do!


anaulin (approved):
Heh, nice find.

PR #1770, opened 2023-08-16
🌸 `Marketplace`: Rename `Events` section to `Timeline`

zspencer:
- https://github.com/zinc-collective/convene/pull/1712#discussion_r1294632293

While either `History` or `Timeline` are way better than `Events`; `Timeline` felt better than `History`; since `Order History` feels like the history of all `Orders`; and `Order Timeline` feels like the `Timeline` for a single `Order`.


anaulin (approved):
Thanks for the improvement!

Issue #1763, opened 2023-08-15
`Marketplace`: `Order#delivery_address`


zspencer:
- When https://github.com/zinc-collective/convene/issues/1326 for https://github.com/zinc-collective/convene/issues/1325, we require a `DeliveryAddress`

## Use Cases
- [ ] `Shopper` fills in `DeliveryAddress` by hand
- [ ] `DeliveryAddress` may be auto-completed
- [ ] `DeliveryAddress` is validated

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

Issue #1672, opened 2023-07-14
EPIC: `Marketplace` - Improve the checkout experience


KellyAH:
- improvements to https://github.com/zinc-collective/convene/issues/831

This is an attempt to focus our efforts on improving/fixing the currently most valuable area of the app, the `Marketplace` which is used by [Piikup](https://github.com/zinc-collective/convene/issues/832) and other clients.

# Tasks 
## 🐛 Fix bugs
Anyone can pick up fixing a bug 😄 
- [x] https://github.com/zinc-collective/convene/issues/1657
- [x] https://github.com/zinc-collective/convene/issues/1656
- [x] https://github.com/zinc-collective/convene/issues/1658
- [x] https://github.com/zinc-collective/convene/issues/1660

## Feature

**Marketplace buyers can see when their order will arrive**
See - [Project doc](https://docs.google.com/document/d/1wbXO82OeEkoIOj8zL6B24fiCC2Wg-w7949fo6wg3qhc/edit?usp=sharing)
- [x] MVP  - @daltonrpruitt and @KellyAH will pair on 7/25 https://github.com/zinc-collective/convene/pull/1705
- [x] https://github.com/zinc-collective/convene/issues/1509
- [ ] V1 - **not doing right now - out of scope!**

anaulin:
Catching up with this project plan today (it's been that kind of week...). Thanks for putting this together, @KellyAH, I found it very helpful to see it all written up and also the screenshots with examples of how other apps do it. 😻

I wanted to offer some thoughts on the MVP, in case these are helpful for the implementation:
> Always Display marketplace delivery window value on the marketplace page. Either at the top or bottom where total price is displayed. 

In our current architecture, the delivery window depends on the delivery area chosen by the shopper. So we don't know the delivery window until the shopper chooses a delivery area -- that's why at the moment we don't show it until the "delivery details" are entered by the shopper.

The way I see other delivery sites dealing with this is that either they just have one delivery area for the entire marketplace, or they ask you upfront before you start shopping to enter a zip code or address or otherwise choose a delivery area. We could do something similar in Convene:
* if Marketplace has only one delivery area
  * all orders will default to this delivery are, we don't need shoppers to pick one 
  * we can show the delivery windows and delivery information on the Marketplace without any need for shopper input
* if Marketplace has more than one delivery area
  * we will need the shopper to pick a delivery area before we can show them the delivery windows
[...]

Issue #1509, opened 2023-05-26
🌸 `Marketplace`: Streamline `Shopper` => `Delivery` => `Cart` => `Checkout` flow


zspencer:
- https://github.com/zinc-collective/convene/issues/1325

> April voiced a preference for setting the delivery time and location more prominently since during her demos with potential clients and shoppers she's noticed bumps in setting that straight.

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #1772, opened 2023-08-17
🌸✨ `Marketplace`: Move `Checkout` to it's own page

zspencer:
- https://github.com/zinc-collective/convene/issues/1509
- https://github.com/zinc-collective/convene/issues/1672
- https://github.com/zinc-collective/convene/issues/1763

This starts to split out the Checkout flow to it's own page; so we can do things like:

- Collect the Delivery Address and Phone Number
- Let the Shopper pick a Delivery Time
- Confirm prices
- ???

## After

https://github.com/zinc-collective/convene/assets/50284/e6a3a3fc-ae27-49a5-b47c-dbd4d5a0d5e0


app/furniture/marketplace/checkouts/show.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= turbo_frame_tag(checkout.marketplace) do %>
```
rosschapman, line 1: Just read about how `turbo_frame_tag` creates a unique id from an object. Very rails-y!

spec/furniture/marketplace/buying_products_system_spec.rb
```diff
@@ -26,16 +26,19 @@ def url_options
...
-      contact_phone_number: "1234567890")
+    select(marketplace.delivery_areas.first.label, from: "cart[delivery_area_id]")
+    click_on("Save changes")
 
     add_product_to_cart(marketplace.products.first)
 
     expect(page).to have_content("Total: #{humanized_money_with_symbol(marketplace.products.first.price + marketplace.delivery_areas.first.price)}")
 
-    click_link_or_button("Checkout")
+    click_on("Checkout")
```
rosschapman, line 36: I like that this helper is more agnostic. Have we ever discussed special data attributes for test targets in UI code? This is a pattern I particular like for even further separating the markup from behavior. For example, Assuming rspec helpers allow for css selectors: 

```ruby
click_on('[data-test="checkout"]')
```

rosschapman (approved):

rosschapman:
@zspencer I think all this makes sense! I like where it's going 🚀

app/furniture/marketplace/cart/deliveries/_delivery.html.erb
```diff
@@ -1,33 +1,28 @@
...
-        label: t("marketplace.cart.deliveries.edit.details_missing"),
-        href: delivery.location(:edit),
-        method: :get,
-        turbo_stream: true,
-        scheme: :primary,
-      ) %>
-  <% end %>
-</div>
+    <h2>Payment Details</h2>
+    <%= button_to("Make Payment", delivery.cart.location(child: :checkout), data: { turbo: false }) %>
```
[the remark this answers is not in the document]
zspencer: It's because we redirect to Stripe; rather than rendering something on our own.

app/furniture/marketplace/checkouts/show.html.erb
zspencer: Yep! It uses `dom_id` under the covers; which I leverage quite a bit to encourage model / view consistency. Theoretically it means that drilling down from `inspect element` should be a lot easier...

spec/furniture/marketplace/buying_products_system_spec.rb
zspencer: Let's wait until we start experiencing brittleness that requires that. I'm of two minds:

1. "The UI *copy* should not be part of system tests" (which would encourage usage of structural selectors rather than textual selectors)
2. "The less attributes we *need* to add to make a system test, the better, since it's less code"

Right now I'm leaning pretty heavily towards 2; but I'm not opposed to using test ids in contexts where they are useful.

app/furniture/marketplace/cart/deliveries/_delivery.html.erb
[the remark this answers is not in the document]
rosschapman: Ah right. So every descendant link or button inside a turbo frame will be turbo-enabled by default.

spec/furniture/marketplace/buying_products_system_spec.rb
rosschapman: Yeah I agree there's not enough copy/code churn that makes a strong case at this point. It's a noop at this time. 

Although, regarding your first point, in my experience structural selectors can actually be more brittle than UI copy since devs often futz with the markup structure more often in response to capability changes than the user-facing verbiage.

zspencer: Yea, I am thinking more like `#order_12345_checkout_form input[type="submit"]` than `div.foo .bar ...`

Issue #1659, opened 2023-07-13
🐞 `Space`: Members page isn't listing any members


anaulin:
I should be able at the very least to see myself in this list:
![image](https://github.com/zinc-collective/convene/assets/6729309/1c73b9df-4e6f-4fb6-931e-f8a1fe8db60a)

anaulin:
Debugging this I realized that this is happening because I was logged in as a guest, and not a member of the Space. The real bug here is that as a guest I should not be shown the ⚙️ icon to manage the space.

PR #1778, opened 2023-08-24
🐞🔨 `Operators` always see all `Space` `Members`

anaulin:
Resolves https://github.com/zinc-collective/convene/issues/1659
- https://github.com/zinc-collective/convene/issues/1659

This was the root cause of why I was getting such weird behavior in production, not seeing any members in the Piikup Marketplace Space: it is because I was logged in with an account that is an operator, but not a member of the Piikup Space.

### DIFF
--- a/app/policies/membership_policy.rb
+++ b/app/policies/membership_policy.rb
@@ -23,7 +23,11 @@ class MembershipPolicy < ApplicationPolicy
 
   class Scope < ApplicationScope
     def resolve
-      scope.where(space: person.spaces)
+      if person.operator?
+        scope.all
+      else
+        scope.where(space: person.spaces)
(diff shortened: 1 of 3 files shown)


zspencer (approved):
Thank you! This makes sense since we didn't have an operator at the time.

Invitations may have a similar bug?

anaulin:
@zspencer I looked over all the other policy scopes, and didn't see any that would have this bug. But that's because for several objects we either don't have a policy scope, or for some the policy scope is defined `scope.all` for all users -- this latter case is what we have for Invitations.

Maybe in a follow-up PR I'll go make all the policies "proper", in terms of the spaces that are visible for a Person.

PR #1791, opened 2023-08-31
🛠️🔐 `Security`: Review Policy objects and constrain default access levels

anaulin:
Working on https://github.com/zinc-collective/convene/pull/1778, I noticed that we have a few Scopes that are defined as `scope.all` for all cases, and probably shouldn't be.

Some of these scopes might not be in use at the moment, but I figured I could fix them all up to prevent future issues.

### DIFF
--- a/app/policies/person_policy.rb
+++ b/app/policies/person_policy.rb
@@ -19,7 +19,13 @@ class PersonPolicy < ApplicationPolicy
 
   class Scope < ApplicationScope
     def resolve
-      scope.all
+      return scope.none if person.blank?
+
+      if person.operator?
+        scope.all
+      else
+        scope.where(id: person.id)
+      end
     end
   end
 end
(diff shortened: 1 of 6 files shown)


app/policies/person_policy.rb
```diff
@@ -19,7 +19,13 @@ def permitted_attributes(_params)
 
   class Scope < ApplicationScope
     def resolve
-      scope.all
+      return scope.none if person.blank?
+
+      if person.operator?
+        scope.all
+      else
+        scope.where(id: person.id)
```
zspencer, line 27: So I can only see myself?

zspencer (approved):

app/policies/person_policy.rb
anaulin: Yep! But you can see the `Memberships` for all of your Spaces (defined in the `MembershipPolicy`).

anaulin: (Also, the `show?` policy method above matches this: you can only `show?` yourself if you're not an Operator.)

zspencer: Makes sense! Thank you!

PR #1793, opened 2023-08-31
🧹🛠️ `Marketplace`: Wrap the whole thing in Turbo Frames

zspencer:

This may turn out to be a terrrrrrible idea; but it may also be fine.

Essentially, the `Marketplace` now lives within a turbo-frame; which means it's going to be replacing itself inline. Cool? Also, horrifying.

Basically, now every `Marketplace::*Controller` will want to inherit from `Marketplace::Controller`, so when we click a link it will auto-magically swap out the content without requiring us to do anything at all! Cool, huh?

### DIFF
--- a/app/furniture/marketplace/payment_settings_controller.rb
+++ b/app/furniture/marketplace/payment_settings_controller.rb
@@ -1,11 +1,7 @@
 class Marketplace
-  class PaymentSettingsController < FurnitureController
+  class PaymentSettingsController < Controller
     def index
       authorize(marketplace, :edit?)
-    end
-
-    helper_method def marketplace
-      @marketplace ||= policy_scope(Marketplace).find(params[:marketplace_id])
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/furniture/marketplace/payment_settings_controller.rb
```diff
@@ -1,11 +1,7 @@
 class Marketplace
-  class PaymentSettingsController < FurnitureController
+  class PaymentSettingsController < Controller
     def index
       authorize(marketplace, :edit?)
     end
-
-    helper_method def marketplace
-      @marketplace ||= policy_scope(Marketplace).find(params[:marketplace_id])
-    end
```
anaulin, line 9: Nice tightening.

anaulin (approved):
It does sound cool. And a little scary!

I wonder if this means that we should be more consistent with adding more browser-based tests for Marketplace functionality, to make sure the content is always replaced as intended. Or are there other types of tests that would be good enough to ensure that we don't forget this new layout in some controller?

zspencer:
Yea, I think right now we have only the Buying Products half of the core-use-case tested with browser tests; so Selling Products, Taxes, etc. all still want browser/system level tests.

app/furniture/marketplace/payment_settings_controller.rb
rosschapman: Oh that's helpful to see that. When I implemented this I forgot about the existence of `Marketplace::Controller`.

PR #1824, opened 2023-09-14
🧹 `C&M`: Squash the migrations!

zspencer:
- https://github.com/zinc-collective/convene/issues/1520

We have... so many migrations. But they don't need to exist!

So this scoops them down and checks in the resulting schema


anaulin (approved):
Thank you for squashing these!

db/schema.rb
```diff
@@ -77,7 +77,7 @@
     t.integer "sluggable_id", null: false
     t.string "sluggable_type", limit: 50
     t.string "scope"
-    t.datetime "created_at", precision: nil
```
rosschapman, line 80: @zspencer assuming changing precision of this field won't have any impact, ie if we were comparing any date strings that came from this table? Or doesn't matter since db default is ok for this field?  Also why did this change?

db/schema.rb
```diff
@@ -244,7 +244,7 @@
     t.string "email", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.boolean "operator", default: false
+    t.boolean "operator", default: false, null: false
```
rosschapman, line 247: Similarly curious why this changed?

zspencer: My best guess is that the default behavior for boolean columns changed between Rails 6 and 7; but I don't think it's significant enough to worry about.

db/schema.rb
zspencer: I presume this is because `precision: nil` is the same as leaving precision off; and between Rails 6 and 7 they made that implicit instead of explicit.

Issue #1810, opened 2023-09-07
🐛 `Marketplace`: payment_settings page fails to load


KellyAH:
- https://github.com/zinc-collective/convene/issues/1576

# problem
Given a section exists w/ a fresh marketplace gizmo (missing a Stripe Account/API key)

1. Go to "Manage Marketplace" e.g prod sandbox: https://convene.zinc.coop/spaces/sandbox/rooms/fake-shop/marketplaces/fac9f78c-23bd-423a-bf5a-4fc7a74a9685/edit

![Screenshot 2023-09-06 at 6 10 18 PM](https://github.com/zinc-collective/convene/assets/16140873/c098b343-8f95-4c34-bfde-8b3127bfc403)

3. click "payment settings" button

**expected behavior**
payment settings loads

**actual behavior**
404
![Screenshot 2023-09-06 at 6 56 14 PM](https://github.com/zinc-collective/convene/assets/16140873/402ae471-aed2-4ea0-8d60-e409f74332b4)

Replicated beak on local and prod sandbox.

Preview for component is also broken
http://127.0.0.1:3000/components/inspect/marketplace/stripe_overview/default
![Screenshot 2023-09-06 at 7 02 04 PM](https://github.com/zinc-collective/convene/assets/16140873/981614d6-0399-4211-aec1-73e217f68ee2)

# Fix
process of elimination revealed that something about the app/furniture/marketplace/stripe_overview_component.rb is causing the page to 404.

zspencer:
Thank you! I've linked this bug the `Marketplace: Accepting Payments feature: https://github.com/zinc-collective/convene/issues/1576

zspencer:
I've deployed a fix! @KellyAH when you get a chance can you confirm that my fix is great?

KellyAH:
✅ confirmed fixed* on prod. confirmed the payment settings section loads on the "Manage Marketplace" page (`/spaces/<space_name>/rooms/<room_name>/marketplaces/<marketplace_id>/edit`) *if a stripe utility exists. (add it via Sandbox › Configure Space › Add Utility)  

Tested on:
Chrome Version 117.0.5938.88
User: k_ahong@yahoo.com 
Space: "Sandbox"
https://convene.zinc.coop/spaces/sandbox/rooms/cake-cabana/marketplaces/3a73864c-e725-4694-aa3c-97f68d48fd0d/edit

[...]

PR #1825, opened 2023-09-14
🥗🐞🔨`Marketplace`: Fix broken Payment Settings page!

zspencer:
- https://github.com/zinc-collective/convene/issues/1810

Technically I think there is still an issue; but this adds a test!

In particular, I think there may still be a bug where *Operators* can't go to these pages.

I think it's worth merging this as is, so that it fixes the page for April

### DIFF
--- a/app/furniture/marketplace/stripe_accounts_controller.rb
+++ b/app/furniture/marketplace/stripe_accounts_controller.rb
@@ -1,5 +1,5 @@
 class Marketplace
-  class StripeAccountsController < FurnitureController
+  class StripeAccountsController < Controller
     def create
       # TODO: Should we be using https://stripe.com/docs/connect/oauth-express-accounts instead?!
       authorize(marketplace, :edit?)
@@ -22,10 +22,6 @@ class Marketplace
 
     def show
       authorize(marketplace, :edit?)
-    end
-
-    helper_method def marketplace
-      @marketplace ||= policy_scope(Marketplace).find(params[:marketplace_id])
     end
   end
 end
--- a/app/furniture/marketplace/stripe_accounts_controller.rb
+++ b/app/furniture/marketplace/stripe_accounts_controller.rb
@@ -1,7 +1,6 @@
 class Marketplace
   class StripeAccountsController < Controller
     def create
-      # TODO: Should we be using https://stripe.com/docs/connect/oauth-express-accounts instead?!
       authorize(marketplace, :edit?)
       stripe_account_link = marketplace.stripe_account_link(
         refresh_url: polymorphic_url(marketplace.location(:edit)),
--- a/spec/furniture/marketplace/collecting_payments_system_spec.rb
+++ b/spec/furniture/marketplace/collecting_payments_system_spec.rb
@@ -1,0 +1,27 @@
+require "rails_helper"
+
+# @see https://github.com/zinc-collective/convene/issues/1576
+describe "Marketplace: Collecting Payments", type: :system do
+  # @see https://github.com/zinc-collective/convene/issues/1622
+  describe "via Stripe" do
+    it "Connecting a Vendor Stripe Account" do
+      space = create(:space, :with_entrance, :with_members)
+      marketplace = create(:marketplace, :with_stripe_utility, room: space.entrance)
+      sign_in(space.members.first, space)
+
+      visit polymorphic_path(marketplace.room.location)
+
+      within("##{dom_id(marketplace, :onboarding)}") do
+        click_on("Manage Marketplace")
+      end
+      click_on("Payment Settings")
+      click_on("View Stripe Account")
+      expect(page).to have_content("Connect to Stripe")
+    end
+  end
+
+  def url_options
+    server_uri = URI(page.server_url)
+    {host: server_uri.host, port: server_uri.port}
+  end
+end
--- a/spec/furniture/marketplace/collecting_payments_system_spec.rb
+++ b/spec/furniture/marketplace/collecting_payments_system_spec.rb
@@ -4,7 +4,7 @@ require "rails_helper"
 describe "Marketplace: Collecting Payments", type: :system do
   # @see https://github.com/zinc-collective/convene/issues/1622
   describe "via Stripe" do
-    it "Connecting a Vendor Stripe Account" do
+    it "connects the Vendor's Stripe Account" do
       space = create(:space, :with_entrance, :with_members)
       marketplace = create(:marketplace, :with_stripe_utility, room: space.entrance)
       sign_in(space.members.first, space)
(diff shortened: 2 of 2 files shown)


app/furniture/marketplace/stripe_accounts_controller.rb
```diff
@@ -1,5 +1,5 @@
 class Marketplace
-  class StripeAccountsController < FurnitureController
+  class StripeAccountsController < Controller
     def create
       # TODO: Should we be using https://stripe.com/docs/connect/oauth-express-accounts instead?!
```
anaulin, line 4: Not from this PR, but we probably can remove this TODO for now.

spec/furniture/marketplace/collecting_payments_system_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      within("##{dom_id(marketplace, :onboarding)}") do
+        click_on("Manage Marketplace")
+      end
+      click_on("Payment Settings")
+      click_on("View Stripe Account")
+      expect(page).to have_content("Connect to Stripe")
+    end
+  end
+
+  def url_options
```
anaulin, line 23: I don't see this method used anywhere obvious. Is it used somehow implicitly? (If so, maybe a comment explaining that might be worth adding.)

spec/furniture/marketplace/collecting_payments_system_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+# @see https://github.com/zinc-collective/convene/issues/1576
+describe "Marketplace: Collecting Payments", type: :system do
+  # @see https://github.com/zinc-collective/convene/issues/1622
+  describe "via Stripe" do
+    it "Connecting a Vendor Stripe Account" do
```
anaulin, line 7: Nit: Slightly better naming here could be something like:
```suggestion
  describe "when using Stripe" do
    it "connects a Vendor Stripe Account" do
```

anaulin, line 7: Nit: Possibly slightly better test case naming:
```suggestion
  describe "with Stripe" do
    it "connects a Vendor Stripe Account" do
```

anaulin (approved):

spec/furniture/marketplace/collecting_payments_system_spec.rb
anaulin: I guess the test right now actually doesn't test that it "connects" the Stripe account. More like it "prompts to connect" or "gives option to connect" or something.

app/furniture/marketplace/stripe_accounts_controller.rb
```diff
@@ -23,9 +23,5 @@ def new
     def show
       authorize(marketplace, :edit?)
     end
-
-    helper_method def marketplace
-      @marketplace ||= policy_scope(Marketplace).find(params[:marketplace_id])
-    end
```
rosschapman, line 29: Nice thanks for the cleanup again.

spec/furniture/marketplace/collecting_payments_system_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+# @see https://github.com/zinc-collective/convene/issues/1576
```
rosschapman, line 3: Why are these back references here? I was assuming it might point to a specific comment or something in the issue related to the test. New documentation pattern?

spec/furniture/marketplace/collecting_payments_system_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    it "Connecting a Vendor Stripe Account" do
+      space = create(:space, :with_entrance, :with_members)
+      marketplace = create(:marketplace, :with_stripe_utility, room: space.entrance)
+      sign_in(space.members.first, space)
+
+      visit polymorphic_path(marketplace.room.location)
+
+      within("##{dom_id(marketplace, :onboarding)}") do
+        click_on("Manage Marketplace")
+      end
```
rosschapman, line 16: 🤔 Here I'm wishing for a single test selector target, so there's no need to scope the dom selection.

spec/furniture/marketplace/collecting_payments_system_spec.rb
rosschapman: Potentially copy pasta from https://github.com/zinc-collective/convene/blob/f3c78d6dc4eea3fbb3b7cd011ad0dbb3bbc35e89/spec/furniture/marketplace/buying_products_system_spec.rb#L22?

rosschapman (approved):

spec/furniture/marketplace/collecting_payments_system_spec.rb
zspencer: So, the way I am using issues is a little different than most teams; where all issues are short lived. Instead, I'm using long-lived issues to describe the overall feature definition. The backlink brings you to the long-lived issue describing the "`Marketplace`: Collecting Payments" value proposition.

The issue itself then links out to whatever other evergreen/long-lived documentation we want; like product marketing copy, interaction design wireframes, etc. 

If this test gets to be too big, we could sprout a `marketplace/collecting_payments/via_stripe_system_spec.rb` which would link to the issue  that describes that particular dip down into the fractal of super-powers we're providing.

spec/furniture/marketplace/collecting_payments_system_spec.rb
zspencer: I tend to prefer scoping to dropping test ids; but I don't block folks who lean into test ids.

app/furniture/marketplace/stripe_accounts_controller.rb
zspencer: True

spec/furniture/marketplace/collecting_payments_system_spec.rb
zspencer: It was a copy-pasta when I was trying to figure out why I couldn't run the test independently; I'm not sure it's necessary tho!

spec/furniture/marketplace/collecting_payments_system_spec.rb
zspencer: Yea, I was trying to figure out how to do the "actual test connection" bit but I ran out of spoons. Maybe someday!

spec/furniture/marketplace/collecting_payments_system_spec.rb
zspencer: I've remofed it and things seem fine; so I think it was just for the test that dealt with mailers.

spec/furniture/marketplace/collecting_payments_system_spec.rb
rosschapman: 👍🏻 Cool. Yeah, I get that we work with issues as a combination of living spec. So basically you're just pointing back to design docs. Which I think makes sense. I guess I was wondering if you would, then, point back to issues on all unit tests, or just feature specs, etc....

zspencer: I am planning to map system specs back to the issue definitions; since trying to do it on a class-structure level feels liek a bit too tight of coupling to keep consistent; whereas the system specs should be exercising a single use-case.

PR #1834, opened 2023-09-20
🥗 `Marketplace`: Test `Seller` receives `Order` in their `Square` Point of Sale

rosschapman:

### DIFF
--- a/app/furniture/marketplace/order.rb
+++ b/app/furniture/marketplace/order.rb
@@ -55,14 +55,19 @@ class Marketplace
 
     def send_to_square_seller_dashboard
       square_create_order_response = create_square_order
-      square_create_payment_response = create_square_order_payment(square_create_order_response.body.order[:id])
 
-      # This data is intended for use in debugging, etc... until we further
-      # the Square integration productize
-      {
-        order_id: square_create_order_response.body.order[:id],
-        payment_id: square_create_payment_response.body.payment[:id]
-      }
+      if square_create_order_response
+        square_create_payment_response = create_square_order_payment(square_create_order_response.body.order[:id])
+
+        # This data is intended for use in debugging, etc... until we further
+        # the Square integration productize
+        {
+          order_id: square_create_order_response.body.order[:id],
+          payment_id: square_create_payment_response.body.payment[:id]
+        }
+      else
+        # TODO: Noop for now
+      end
     end
 
     # Square sets max of 43 chars
--- a/spec/furniture/marketplace/stripe_events_controller_request_spec.rb
+++ b/spec/furniture/marketplace/stripe_events_controller_request_spec.rb
@@ -3,9 +3,13 @@ require "rails_helper"
 # rubocop:disable RSpec/VerifiedDoubles
 RSpec.describe Marketplace::StripeEventsController, type: :request do
   let(:marketplace) { create(:marketplace, :with_stripe_utility, stripe_account: "sa_1234", stripe_webhook_endpoint_secret: "whsec_1234") }
+  let(:marketplace_with_square) { create(:marketplace, :with_stripe_utility, stripe_account: "sa_1234", stripe_webhook_endpoint_secret: "whsec_1234", square_location_id: "123456789", square_environment: "sandbox", secrets: {square_access_token: "fake_square_access_token"}) }
   let(:space) { marketplace.space }
   let(:member) { create(:membership, space: space).member }
   let(:order) { create(:marketplace_order, :with_products, status: :pre_checkout, marketplace: marketplace) }
+  let(:person) { create(:person) }
+  let(:shopper) { create(:marketplace_shopper, person: person) }
+  let(:order_for_square) { create(:marketplace_order, :with_products, status: :pre_checkout, marketplace: marketplace_with_square, shopper: shopper) }
 
   let(:stripe_event) do
     double(Stripe::Event, type: "checkout.session.completed",
@@ -22,6 +26,14 @@ RSpec.describe Marketplace::StripeEventsController, type: :request do
     double(Stripe::Charge, balance_transaction: "btx_2234")
   }
 
+  let(:square_client) {
+    double(Square::Client)
+  }
+
+  let(:square_orders_api) {
+    double(Square::OrdersApi)
+  }
+
   before do
     allow(Stripe::Webhook).to receive(:construct_event).with(anything, "sig_1234", marketplace.stripe_webhook_endpoint_secret).and_return(stripe_event)
 
@@ -29,11 +41,15 @@ RSpec.describe Marketplace::StripeEventsController, type: :request do
     allow(Stripe::Transfer).to receive(:create).and_return(double(Stripe::Transfer, id: "st_fake_1234"))
     allow(Stripe::BalanceTransaction).to receive(:retrieve).with("btx_2234", anything).and_return(balance_transaction)
     allow(Stripe::Charge).to receive(:retrieve).with("ch_1234", anything).and_return(charge)
+    allow(Square::Client).to receive(:new).with(access_token: anything, environment: anything).and_return(square_client)
+    allow(square_client).to receive(:orders).and_return(square_orders_api)
+    allow(square_client).to receive(:payments).and_return(square_payments_api)
+    allow(square_orders_api).to receive(:create_order)
   end
 
   describe "#create" do
     subject(:call) do
-      post polymorphic_path(marketplace.location(child: :stripe_events)), headers: {HTTP_STRIPE_SIGNATURE: "sig_1234"}
+      post polymorphic_path(order.marketplace.location(child: :stripe_events)), headers: {HTTP_STRIPE_SIGNATURE: "sig_1234"}
       order.reload
       response
     end
@@ -79,6 +95,22 @@ RSpec.describe Marketplace::StripeEventsController, type: :request do
         expect(order.reload.placed_at).to be_nil
 
         expect(Stripe::Transfer).not_to(have_received(:create))
+      end
+    end
+
+    context "when Square notifications are not enabled" do
+      it "does not attempt to transfer the order to seller's Square dashboard" do
+        call
+        expect(square_client).not_to(have_received(:orders))
+      end
+    end
+
+    context "when Square notifications are enabled" do
+      let(:order) { order_for_square }
+
+      it "attempts to transfer the order to seller's Square dashboard" do
+        call
+        expect(square_client).to have_received(:orders)
       end
     end
   end
--- a/spec/furniture/marketplace/stripe_events_controller_request_spec.rb
+++ b/spec/furniture/marketplace/stripe_events_controller_request_spec.rb
@@ -43,7 +43,6 @@ RSpec.describe Marketplace::StripeEventsController, type: :request do
     allow(Stripe::Charge).to receive(:retrieve).with("ch_1234", anything).and_return(charge)
     allow(Square::Client).to receive(:new).with(access_token: anything, environment: anything).and_return(square_client)
     allow(square_client).to receive(:orders).and_return(square_orders_api)
-    allow(square_client).to receive(:payments).and_return(square_payments_api)
     allow(square_orders_api).to receive(:create_order)
   end
 
--- a/app/furniture/marketplace/marketplace.rb
+++ b/app/furniture/marketplace/marketplace.rb
@@ -21,6 +21,7 @@ class Marketplace
     alias_method :vendor_stripe_account, :stripe_account
     setting :stripe_webhook_endpoint
     setting :stripe_webhook_endpoint_secret
+    setting :square_environment
 
     # Square order notifications integration
     setting :square_location_id


app/furniture/marketplace/order.rb
```diff
@@ -55,14 +55,19 @@ def price_total
...
       square_create_order_response = create_square_order
-      square_create_payment_response = create_square_order_payment(square_create_order_response.body.order[:id])
 
-      # This data is intended for use in debugging, etc... until we further
-      # the Square integration productize
-      {
-        order_id: square_create_order_response.body.order[:id],
-        payment_id: square_create_payment_response.body.payment[:id]
-      }
+      if square_create_order_response
```
rosschapman, line 59: 🤔 The conditionality here actually makes testing simpler for the time being without having to mock responses.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -3,9 +3,13 @@
 # rubocop:disable RSpec/VerifiedDoubles
 RSpec.describe Marketplace::StripeEventsController, type: :request do
   let(:marketplace) { create(:marketplace, :with_stripe_utility, stripe_account: "sa_1234", stripe_webhook_endpoint_secret: "whsec_1234") }
+  let(:marketplace_with_square) { create(:marketplace, :with_stripe_utility, stripe_account: "sa_1234", stripe_webhook_endpoint_secret: "whsec_1234", square_location_id: "123456789", square_environment: "sandbox", secrets: {square_access_token: "fake_square_access_token"}) }
   let(:space) { marketplace.space }
   let(:member) { create(:membership, space: space).member }
   let(:order) { create(:marketplace_order, :with_products, status: :pre_checkout, marketplace: marketplace) }
+  let(:person) { create(:person) }
+  let(:shopper) { create(:marketplace_shopper, person: person) }
+  let(:order_for_square) { create(:marketplace_order, :with_products, status: :pre_checkout, marketplace: marketplace_with_square, shopper: shopper) }
```
rosschapman, line 12: FYI I did play with updating the existing `order` record at first but realized it would be simpler to consider a totally new object for the new case.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -22,18 +26,30 @@
     double(Stripe::Charge, balance_transaction: "btx_2234")
   }
 
+  let(:square_client) {
+    double(Square::Client)
+  }
+
+  let(:square_orders_api) {
+    double(Square::OrdersApi)
+  }
```
rosschapman, line 35: I decided to stop at mocking these apis. Although you could mock all the way down to the HTTP responses. Although, what would make mocking the Square connection simpler would be hiding all of the square stuff behind it's own module.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -81,6 +97,22 @@
         expect(Stripe::Transfer).not_to(have_received(:create))
       end
     end
+
+    context "when Square notifications are not enabled" do
+      it "does not attempt to transfer the order to seller's Square dashboard" do
```
rosschapman, line 102: FYI: Unless we record real http interactions with Square (do we want to now, ever, using `vcr` or something else?) I like using the term "attempt" to more accurately describe what the test covers.

rosschapman:
@zspencer @anaulin initial thought on integration testing the new Square code. I played around for a few hours to familiarize myself with rspec mocking and thought about a few different approaches here. I'm even ok with throwing all of this out if we want to test agains the actual sandbox, record interactions (eg, `vcr`), or mock deeper into the Square stack. Thanks for taking a look at your convenience!

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
zspencer: In rspec, you can use [`context` blocks](https://rspec.rubystyle.guide/#use-contexts) to group test setup. Example:


```rb
context "when the marketplace has a Square integration" do
  let(:marketplace) {  create(:marketplace, :with_stripe_utility, stripe_account: "sa_1234", stripe_webhook_endpoint_secret: "whsec_1234", square_location_id: "123456789", square_environment: "sandbox", secrets: {square_access_token: "fake_square_access_token"})  }
  # ...
end
```

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
zspencer: I tend to mock at the ruby level except in system tests; and in system tests I prefer something like how we did the Stripe integration, so it winds up actually hitting a Stripe account.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -22,18 +26,30 @@
...
+  }
+
   before do
     allow(Stripe::Webhook).to receive(:construct_event).with(anything, "sig_1234", marketplace.stripe_webhook_endpoint_secret).and_return(stripe_event)
 
     allow(Stripe::PaymentIntent).to receive(:retrieve).with("pi_1234", anything).and_return(payment_intent)
     allow(Stripe::Transfer).to receive(:create).and_return(double(Stripe::Transfer, id: "st_fake_1234"))
     allow(Stripe::BalanceTransaction).to receive(:retrieve).with("btx_2234", anything).and_return(balance_transaction)
     allow(Stripe::Charge).to receive(:retrieve).with("ch_1234", anything).and_return(charge)
+    allow(Square::Client).to receive(:new).with(access_token: anything, environment: anything).and_return(square_client)
```
zspencer, line 44: I prefer to use more specific values than `anything` so that tests will fail if we start passing in nonsense (i.e. missing an access token or environment)

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
zspencer: I don't find usages of `attempt` or similar particularly useful. The test describes what it does. Whether or not it is mocked feels incidental.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -81,6 +97,22 @@
...
+        expect(square_client).not_to(have_received(:orders))
+      end
+    end
+
+    context "when Square notifications are enabled" do
+      let(:order) { order_for_square }
+
+      it "attempts to transfer the order to seller's Square dashboard" do
+        call
+        expect(square_client).to have_received(:orders)
```
zspencer, line 113: I tend to prefer more explicit expectations; as it stands we could just call `orders` and the tests would continue to pass. If we include the actual data structure it would be a bit more tightly coupled; but would also fail in the event we accidentally start munging about the data.

Once we start adding more permutations, we'll want to push the test down below the request spec level; and probably add a `Marketplace::Square::Order` class to encapsulate that logic.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -81,6 +97,22 @@
...
+
+    context "when Square notifications are not enabled" do
+      it "does not attempt to transfer the order to seller's Square dashboard" do
+        call
+        expect(square_client).not_to(have_received(:orders))
+      end
+    end
+
+    context "when Square notifications are enabled" do
+      let(:order) { order_for_square }
```
zspencer, line 109: If you use `let(:marketplace) { ... }` you can get rid of the `order_for_square` bits entirely.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -3,9 +3,13 @@
 # rubocop:disable RSpec/VerifiedDoubles
 RSpec.describe Marketplace::StripeEventsController, type: :request do
   let(:marketplace) { create(:marketplace, :with_stripe_utility, stripe_account: "sa_1234", stripe_webhook_endpoint_secret: "whsec_1234") }
+  let(:marketplace_with_square) { create(:marketplace, :with_stripe_utility, stripe_account: "sa_1234", stripe_webhook_endpoint_secret: "whsec_1234", square_location_id: "123456789", square_environment: "sandbox", secrets: {square_access_token: "fake_square_access_token"}) }
   let(:space) { marketplace.space }
   let(:member) { create(:membership, space: space).member }
   let(:order) { create(:marketplace_order, :with_products, status: :pre_checkout, marketplace: marketplace) }
+  let(:person) { create(:person) }
+  let(:shopper) { create(:marketplace_shopper, person: person) }
```
zspencer, line 11: Why do we need a `Shopper` with a `Person`?

zspencer (approved):
Overall, this looks really good!

I think it could be merged as is, since it does add test coverage; but I would love it if:

- The request spec was more explicit about what data structure and data was being passed to the Square gem
- The code for talking to Square was pulled into an independent class (I would probably go with `Marketplace::SquareOrder` or `Marketplace::Square::Order`, depending on how much of a butt the Ruby constant lookup is being)
- We kept the `let` statements that are specific for particular context blocks in the context block that uses them; so we don't have to jump around the spec to figure out what data is being set.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman: Yeah agree. I believe I may have used this as a placeholder. Good catch.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman: Because it's in the data: https://github.com/zinc-collective/convene/blob/adds-square-notifications-tests/app/furniture/marketplace/order.rb#L136.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman: Nice. Yeah, I like the power of the lazy evaluation offered by ruby here. Although, when possible, I prefer to put data in one place (ie the top of file) _together with_ more descriptive variable names. This will: 
- allow the developer to focus more reasily on the intent of the test in terms of it's description and commands (less code to read in the test assertion itself)
- make data easier to understand at a glance -- oh I have this _type_ of order here, and this _type_ of order there
Does that make sense?

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman: I don't have a strong preference either way, though I like the embellishment because it reads more like natural language for tossing data into a void. Though I'm privvy to a more literary approach to any documentary writing. I can remove it at least so it's consistent with other writing.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman: Yeah I considered this might be a stepping stone to that while I got my bearings writing system tests in our ruby codebase. What do you think about leaving as is for now while I follow-up with a full integration with a Sandbox account?

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman: That said, I also sometimes rewrite larger tests suites like this that live in a single file with context blocks for every method. Then all data setup is contextual to their proper case/usage -- this improves the developer experience per what I listed above. So I'd be into making that adjustment!

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman: Yeah, I agree. I didn't want to get toooo detailed until I got some feedback first from y'all. I'll update with a `.with(...)` clause.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
zspencer: You may prefer to group the data in one place; but it is not what most people who read and write rspec will expect. In general, when I am learning a new language or framework; I try to blend in with what the community conventions are; since I've been bitten time and time again by thinking I know best 🤣

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
zspencer: Ah, I should be more clear; yes, this is where I would stub at the Request Spec level; and I would make assertions against the messages sent; including arguments.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
zspencer: Ahhh, thank you that makes sense. What happens if the `Shopper` does not have a `Person`? `Guest` users for example do not have a `Person` record.

zspencer: @rosschapman - I am sitll a bit worried about what happens when a Guest checks out at a Marketplace with Square. I know it's not the most important thing to deal with now-right-now; since no one has Square set up yet; but a test for that use case would put my mind at ease.

rosschapman: @zspencer sure thing. I've got some updates coming soon so I'll add this. Thanks for reminding me about the _Guest_ use case.

Issue #1806, opened 2023-09-07
🐛 `Marketplace`: "Content missing" displays when click "Add Stripe Account"


KellyAH:
- https://github.com/zinc-collective/convene/issues/1622

# problem
Given a section exists w/ a fresh marketplace gizmo (missing a Stripe Account/API key)

1. Go to "Manage Marketplace" e.g prod sandbox: https://convene.zinc.coop/spaces/sandbox/rooms/fake-shop/marketplaces/fac9f78c-23bd-423a-bf5a-4fc7a74a9685/edit

![Screenshot 2023-09-06 at 6 10 18 PM](https://github.com/zinc-collective/convene/assets/16140873/c098b343-8f95-4c34-bfde-8b3127bfc403)

3. click "Add Stripe Account" link

**expected behavior**
See page to Add Stripe Account
https://convene.zinc.coop/spaces/sandbox/rooms/fake-shop/marketplaces/fac9f78c-23bd-423a-bf5a-4fc7a74a9685/stripe_account
![Screenshot 2023-09-06 at 6 32 16 PM](https://github.com/zinc-collective/convene/assets/16140873/4bc073dc-6f4f-4e48-b842-f4d8ad068580)

**actual behavior**
See  "Content missing" display on the page
![Screenshot 2023-09-06 at 6 10 24 PM](https://github.com/zinc-collective/convene/assets/16140873/ad5af0d5-be0b-4a86-8166-f66b13dd3a96)

Replicated beak on local and prod sandbox.

anaulin:
See associated error reported by Sentry here: https://github.com/zinc-collective/convene/issues/1807

zspencer:
Thank you Kelly and Ana!

zspencer:
Gonna tackle this next!

zspencer:
Patch is in! Turns out there were a number of sad things happening here. Once I get code review; I'll merge and deploy and assign to @KellyAH for testing!

zspencer:
This has been fixed and merged; the deploy should be complete by 6:45 PT on 9/20; at which point if you could re-test it for me @KellyAH and close the issue / provide feedback I would be stoked!

KellyAH:
✅ confirmed fixed on prod.  

Tested on https://sandbox.zinc.coop/spaces/sandbox/rooms/fake-shop/marketplaces/fac9f78c-23bd-423a-bf5a-4fc7a74a9685/edit

PR #1840, opened 2023-09-21
🥗🐞🔨 `Marketplace`: Test & Fix Setting `Stripe` API Key

zspencer:
- https://github.com/zinc-collective/convene/issues/1806

Turns out, there were a number of small bugs here:

- Both `Marketplace::Marketplace` and `Marketplace::StripeOverviewComponent` presumed that there would be a `StripeUtility` for the `Space`; when there may not be.
- The link to `Space#edit` does not have a turbo-frame for the marketplace; so it needed to have turbo disabled
- The `ButtonComponent` did not have a way to disable turbo!

This fixes all of that, and allows `Operators` and `Members` to set the Stripe API key via the `Marketplace#edit` flow.

### DIFF
--- a/app/components/marketplace/stripe_overview_component.rb
+++ b/app/components/marketplace/stripe_overview_component.rb
@@ -7,6 +7,5 @@ class Marketplace::StripeOverviewComponent < ApplicationComponent
     super(**kwargs)
 
     @marketplace = marketplace
-    @marketplace_stripe_utility = marketplace.stripe_utility
   end
 end
--- a/app/furniture/marketplace/marketplace.rb
+++ b/app/furniture/marketplace/marketplace.rb
@@ -42,12 +42,12 @@ class Marketplace
     end
 
     def stripe_utility
-      @stripe_utility ||= space.utilities.find_by!(utility_slug: :stripe).utility
+      @stripe_utility ||= space.utilities.find_by(utility_slug: :stripe)&.utility
     end
 
     # The Secret Stripe API key belonging to the owner of the Marketplace
     def stripe_api_key
-      stripe_utility.api_token
+      stripe_utility&.api_token
     end
 
     def stripe_api_key?
--- a/spec/furniture/marketplace/collecting_payments_system_spec.rb
+++ b/spec/furniture/marketplace/collecting_payments_system_spec.rb
@@ -4,11 +4,43 @@ require "rails_helper"
 describe "Marketplace: Collecting Payments", type: :system do
   # @see https://github.com/zinc-collective/convene/issues/1622
   describe "via Stripe" do
-    it "connects the Vendor's Stripe Account" do
-      space = create(:space, :with_entrance, :with_members)
-      marketplace = create(:marketplace, :with_stripe_utility, room: space.entrance)
-      sign_in(space.members.first, space)
+    let(:space) { create(:space, :with_entrance, :with_members) }
 
+    before do
+      sign_in(space.members.first, space)
+    end
+
+    it "sets the Distributor's Stripe API Key" do # rubocop:disable RSpec/ExampleLength
+      marketplace = create(:marketplace, room: space.entrance)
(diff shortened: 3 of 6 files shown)


app/components/marketplace/stripe_overview_component.rb
```diff
@@ -7,6 +7,5 @@ def initialize(marketplace:, **kwargs)
     super(**kwargs)
 
     @marketplace = marketplace
-    @marketplace_stripe_utility = marketplace.stripe_utility
```
zspencer, line 10: This was raising the `RecordNotFound` on instantiation; which was causing the 404!

KellyAH (approved):
LGTM

rosschapman:
@zspencer @KellyAH  Ah yes, here's to the tests that were [never written](https://github.com/zinc-collective/convene/blob/5ff60865670e3e7062e05733b27167db8bce9a20/spec/components/marketplace/stripe_overview_component_spec.rb#L8)! 🍻

I owe us a view spec for the `stripe_overview_component`. I also wonder if a small unit test of `Marketplace::Marketplace#stripe_utility` is warranted. Typically I _would_ write a small test for code that performs anything more trivial than property access or a boolean. I can work on that tomorrow during a plane ride to Mexico.

app/furniture/marketplace/marketplace.rb
```diff
@@ -42,12 +42,12 @@ def ready_for_shopping?
     end
 
     def stripe_utility
-      @stripe_utility ||= space.utilities.find_by!(utility_slug: :stripe).utility
+      @stripe_utility ||= space.utilities.find_by(utility_slug: :stripe)&.utility
```
rosschapman, line 45: @zspencer `nil` seems very reasonable as return value here!

app/furniture/marketplace/marketplace.rb
```diff
@@ -42,12 +42,12 @@ def ready_for_shopping?
...
     end
 
     # The Secret Stripe API key belonging to the owner of the Marketplace
     def stripe_api_key
-      stripe_utility.api_token
+      stripe_utility&.api_token
```
rosschapman, line 50: Same, I like returning `nil` here as well.

spec/furniture/marketplace/collecting_payments_system_spec.rb
```diff
@@ -4,11 +4,43 @@
...
+      expect(space.utilities).to exist(utility_slug: "stripe")
+      expect(space.utilities.find_by(utility_slug: "stripe").utility.api_token).to eq(ENV.fetch("STRIPE_API_KEY", "not-a-real-key"))
+
+      visit polymorphic_path(marketplace.location(:edit))
+      click_on("Payment Settings")
+      click_on("View Stripe Account")
+      expect(page).to have_content("Connect to Stripe")
+    end
+
+    it "connects the Vendor's Stripe Account" do
```
rosschapman, line 42: 👏🏻

spec/furniture/marketplace/collecting_payments_system_spec.rb
```diff
@@ -4,11 +4,43 @@
...
-    it "connects the Vendor's Stripe Account" do
-      space = create(:space, :with_entrance, :with_members)
-      marketplace = create(:marketplace, :with_stripe_utility, room: space.entrance)
+    let(:space) { create(:space, :with_entrance, :with_members) }
+
+    before do
       sign_in(space.members.first, space)
+    end
+
+    it "sets the Distributor's Stripe API Key" do # rubocop:disable RSpec/ExampleLength
```
rosschapman, line 13: 👏🏻 Great test.

zspencer:
> I also wonder if a small unit test of `Marketplace::Marketplace#stripe_utility` is warranted. Typically I _would_ write a small test for code that performs anything more trivial than property access or a boolean. I can work on that tomorrow during a plane ride to Mexico.

Yea, I was thinking it would be good to add those in!

PR #1842, opened 2023-09-21
Adds content for Stripe Account settings component when Stripe utility missing

rosschapman:
Writing a component view spec encouraged me to "find" the actual conditional cases present here. Generally I prefer unwrapping these cases with separate rendering paths rather than doing more inline cleverness switching on value presence.


zspencer (approved):
I dig it! Thank you!

PR #1856, opened 2023-09-28
Match capy click selectors with underlying html

rosschapman:
Regarding https://github.com/zinc-collective/convene/pull/1848#discussion_r1339158584

The difference between buttons and links are syntactically _and_ semantically confusing. I would love to better understand the accessibility standards around this. Having some expertise here would be helpful. But until we dig into that, what I was thinking is that we could force the developer into a syntactically correct choice to remove any ambiguity in the code about whether an element is a `<button>` or `<a>`.


anaulin (approved):
I think this is a good change, thank you!

You're going to have to fix a bunch more Capybara-based specs for this to go green tho.

anaulin:
You can add exceptions for the newly-failing files if you want, and check this in. That way you are making some progress, but don't need to sink a bunch of time into fixing all these other occurrences.

zspencer:
Yea, we have some work to do to clean up how we're using links and buttons. It was my second project using ViewComponent and Turbo; and I basically brute-forced things to work. 

But getting to the point where buttons == change data and display and links == change display will be a huge win. Probably not the most urgent project but something to look out for as we make other changes.

rosschapman:
@anaulin @zspencer thanks for the review and notes. Only took me ~10 mins to cleanup the rest of the style check failures. 🚀

zspencer:
Wooo! Thank you!

Issue #833, opened 2022-09-15
🌫️✨ `Space`: Download as an App on Mobile App Stores


zspencer:
- https://github.com/zinc-collective/convene/issues/1154

It would be cool if each Space could be packaged as an App for distribution on the iOS or Android App Store(s)

## Bounties:
- $109.52 from https://github.com/zinc-collective/convene/issues/832

## Use Cases

- [ ] `Space` is Installable via iOS App Store
- [ ] `Space` is Installable via Google Play Store

zspencer:
- $109.52 Bounty offered by https://github.com/zinc-collective/convene/issues/832 

Invoice: https://app.qbo.intuit.com/app/invoice?txnId=588

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

Issue #832, opened 2022-09-15
Client: Piikup


zspencer:
- [ ] [Furniture: Marketplace](https://github.com/zinc-collective/convene/issues/831)
- [ ] Utility: ToastTab
- [ ] Utility: CashApp 
- [ ] Utility: Venmo

zspencer:
April (CEO of Piikup) is interested in a demo sometime! I've asked if she'd be available during our monthly member meeting, or during our Wednesday sessions. If not, I'll set aside a separate time.

zspencer:
Demoed our [Marketplace: Furniture](https://github.com/zinc-collective/convene/issues/831) to April today, feedback was:

1. Stripe takes ~3.5% from a $50 transaction. How do we cover the Stripe fees? We discussed a "Solidarity Fee" option that Buyers could opt-in to cover the Vendor/Distributor's payment processor fees.
2. The goal for the Marketplace continues to be about driving purchases to Piikup's Bendors; so we want to keep costs born by the Vendors to a minimum.

zspencer:
Today, I onboarded Ivin and Mikaya as Account Managers for Piikup; who will be supporting Piikups Clients with their eCommerce platform.

I recorded the walk-through that Ivin did, and will be passing it on to them via email.

We've also asked that they do some practice account set up and orders to make sure they're comfortable with the flow; as well as to provide us with adjusting feedback.

zspencer:
Script I used to port data from the sandbox to the production site:

```rb
from = Space.find_by(slug: 'piikup-sandbox')
to = Space.find_by(slug: 'piikup')
from.rooms.map { |r| to.rooms.find_or_initialize_by(name: r.name, slug: r.slug).update!(access_level: :internal, publicity_level: :listed) }
frs = from.rooms
frs.map do |fr| 
  tr = to.rooms.find_or_initialize_by(name: fr.name, slug: fr.slug).update!(access_level: :internal, publicity_level: :unlisted);
  fr.furnitures.each  { |ff| tr.furnitures.find_or_initialize_by(slot: ff.slot).update(furniture_kind: ff.furniture_kind, settings: ff.settings) } }
end
tms = Marketplace::Marketplace.where(furniture_kind: "marketplace", room: to.rooms)
fms = Marketplace::Marketplace.where(furniture_kind: "marketplace", room: from.rooms)
tms.each do |tm| 
[...]

Issue #621, opened 2022-03-24
🌫️✨ `Space`: `Themes`


zspencer:
-  https://github.com/zinc-collective/convene/issues/1154

## Use Cases
- [ ] `Member` may set `Theme`
- [ ] `Member` may preview `Theme`

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424).

PR #1859, opened 2023-09-30
🌸 `Themes`/ `Mobile`:  Eliminate top navigation

zspencer:
- https://github.com/zinc-collective/convene/issues/832
- https://github.com/zinc-collective/convene/issues/833

The top navigation is a bit excessive, especially on mobile. 

This brings the UI to a much cleaner place; with breadcrumbs and sign in / out only on the bottom of the page.

## After
Screenshot 2023-10-04 at 17-34-42 Convene - The Oakland Community Food Network

## Before
Screenshot 2023-09-30 at 16-18-18 Convene - Zee's Space

### DIFF
--- a/app/assets/stylesheets/components/header.scss
+++ b/app/assets/stylesheets/components/header.scss
@@ -1,18 +1,4 @@
 @layer components {
-  body > header {
-    @apply flex justify-between px-2 py-2 bg-orange-500 text-white place-items-end;
-
-    .profile-menu {
-      @apply flex flex-row relative;
-      :first-child {
-        @apply pr-0;
-      }
-      > * {
-        @apply px-2;
-      }
-    }
-  }
-
   .flash-message {
     @apply px-4 py-4;
 
--- a/app/furniture/marketplace/marketplace_component.html.erb
+++ b/app/furniture/marketplace/marketplace_component.html.erb
@@ -1,3 +1,7 @@
+<h2 class="ml-2 my-2">
+  <%= t('.header', space: marketplace.space.name, room: marketplace.room.name) %>
+</h2>
+
 
 <%= render onboarding_component %>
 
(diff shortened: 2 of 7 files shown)


anaulin:
I agree that the current setup with Breadcrumbs and "Your account" and "Sign out" is a bit much ("Sign out" should be an option in the "Your Account" menu, and it should show your email address or name, instead of "Your Account"), but I think we need to replace it with some minimal framing or title for what the user is seeing. Otherwise it is not clear what am I looking at, it doesn't say "Marketplace" anywhere clearly, and it doesn't say "Piikup", I'm just dropped into "choose your delivery address", without confirmation of which site I'm on.

Showing the "room" name in the header was serving a bit of this purpose, because that way you can land on a page and see e.g. "Crumble & Whisk" at the top, but maybe we should instead have a header on the Gizmo itself, so it would say "Crumble & Whisk Marketplace" at the top (and maybe the content can be configurable).

zspencer:
>  It is not clear what am I looking at, it doesn't say "Marketplace" anywhere clearly, and it doesn't say "Piikup", I'm just dropped into "choose your delivery address", without confirmation of which site I'm on.
>
> Showing the "room" name in the header was serving a bit of this purpose, because that way you can land on a page and see e.g. "Crumble & Whisk" at the top, but maybe we should instead have a header on the Gizmo itself, so it would say "Crumble & Whisk Marketplace" at the top (and maybe the content can be configurable).

I like the idea of having the Marketplace include the name of both the Space and the Room (i.e. `Crumble & Whisk on the Piikup Marketplace`) and we could make that configurable!

zspencer:
@anaulin how do you like it now? I put in a header on the Marketplace; which I think it wanted anyway; and that should better ground people so they know where they're at.

anaulin (approved):
I like it!

PR #1860, opened 2023-10-01
🌸 `Themes`: Try a stone?

zspencer:

OK I'm just futzing about; but I *think* what we can do is pull in most of the places where we use colors into the `ViewComponent` layer; then we could put together pallettes with appropriately balanced colors across the built-in tailwind colors.

From there, we can define custom colors for each client with a particular brand, for example `piikup-orange` in tailwind landia!

Anyway, here's the stoney-version!
Screenshot 2023-09-30 at 20-13-02 Convene - Zee's Space

### DIFF
--- a/app/assets/stylesheets/components/button.scss
+++ b/app/assets/stylesheets/components/button.scss
@@ -19,12 +19,12 @@
     }
 
     &.--danger {
-      @apply bg-danger-500;
+      @apply bg-red-500;
       &:hover {
-        @apply bg-danger-700;
+        @apply bg-red-700;
       }
       &:active {
-        @apply bg-danger-200;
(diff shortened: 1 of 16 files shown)


anaulin:
I like the idea of having `color-scheme` in ViewComponents that carries color palette information. Do you want to try to re-implement this "stone" or the "orange" in your other PR as part of a ViewComponent scheme, so that different spaces can pick different palettes?

zspencer:
Yea, I was thinking of doing something like that! I'm going to close this PR for now tho; since it feels more like a learning spike towards that direction.

I am also trying to balance the palette/color-scheme bits with getting Space's working in a mobile app; but my instincts tell me that both will want to happen

PR #1862, opened 2023-10-01
✨ `Marketplace`: Show `DeliveryFee` when `DeliveryArea` is selected

zspencer:

I think this is leftover from when we collected both the address and the area at the same time.


anaulin (approved):
Good catch.

PR #1858, opened 2023-09-30
🌸 `Themes`: Make the primary theme Orange

zspencer:
- https://github.com/zinc-collective/convene/issues/832
- https://github.com/zinc-collective/convene/issues/621

While I would love to support multiple themes or palettes or something; I don't really have a solid enough idea for how to proceed. Meanwhile; Piikup's branding is orange; so it makes sense for Convene to reflect that.

## After

Screenshot 2023-09-30 at 14-50-54 Convene - Zee's Space

## Before
Screenshot 2023-09-30 at 14-53-21 Convene - Zee's Space

PR #1868, opened 2023-10-04
🌸 `Themes`: Violet begats violets

zspencer:
- https://github.com/zinc-collective/convene/pull/1858#issuecomment-1743768286

Now they're orange! Just like everything else! ORANGE ORANGE ORANGE

### DIFF
--- a/app/components/alert_component.rb
+++ b/app/components/alert_component.rb
@@ -25,7 +25,7 @@ class AlertComponent < ApplicationComponent
     when :success
       "bg-green-50"
     else
-      "bg-violet-50"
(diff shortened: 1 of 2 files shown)


anaulin (approved):
🍊

(I'm wondering a bit about the UX of having alerts that are orange and branding that is also orange; might it be confusing? But we can refine that as we go.)

zspencer:
Yea, I think there's some thinkin' we need to do to make sure things aren't... too wild... The success alerts are currently matching branding; but I would expect them to be a complementary color for the main brand...

PR #1870, opened 2023-10-05
🧹 `Section`: Drop the `publicity_level` column, since it is dead

zspencer:


anaulin (approved):
🧹 🧹

PR #1891, opened 2023-10-12
🐞🔨 `Authentication`: Don't explode when OTP or OTP secrets are nil

zspencer:
- https://github.com/zinc-collective/convene/issues/1809

There's a couple things happening, one of which is if an `AuthenticationMethod` does not have a One Time Password Secret, it can't actually do the verification.

So i've added a check to make sure it returns false in cases when the OTP secret has not been set yet; as well as when a nil OTP is provided.


anaulin (approved):
Thank you!

PR #1892, opened 2023-10-12
🐞🔨`Authentication` Prevent race condition when double-tapping sign-in

zspencer:
- https://github.com/zinc-collective/convene/issues/1809

OK, so my working assumption is that the bug is caused by two requests being processed at the same time for a new email address, and the first one creates the `AuthenticationMethod` and the second one *tries* to, and fails; resulting in the validation error.

This requeries the database and retries the rest of the work.

Would love to have names for methods to pull this out into, rather than copy-pasting the method body but oh well.

### DIFF
--- a/app/models/authenticated_session.rb
+++ b/app/models/authenticated_session.rb
@@ -45,6 +45,16 @@ class AuthenticatedSession
     end
 
     false
+  rescue ActiveRecord::RecordInvalid
+    self.authentication_method = nil
+    if one_time_password.nil?
+      authentication_method.send_one_time_password!(space)
+      false
+    elsif authentication_method.verify?(one_time_password)
+      session[:person_id] = authentication_method.person.id
+      authentication_method.confirm!
+      true
+    end
   end
 
   # If we don't have a OTP _or_ a way of issuing one, there's nothin' we can do.
--- a/app/models/authenticated_session.rb
+++ b/app/models/authenticated_session.rb
@@ -35,26 +35,10 @@ class AuthenticatedSession
   def save
     return false if !valid? || !actionable?
 
-    if one_time_password.nil?
-      authentication_method.send_one_time_password!(space)
-      return false
-    elsif authentication_method.verify?(one_time_password)
-      session[:person_id] = authentication_method.person.id
-      authentication_method.confirm!
-      return true
-    end
-
-    false
+    verify_or_resend_otp
   rescue ActiveRecord::RecordInvalid
     self.authentication_method = nil
-    if one_time_password.nil?
-      authentication_method.send_one_time_password!(space)
-      false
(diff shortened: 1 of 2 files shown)


app/models/authenticated_session.rb
```diff
@@ -45,6 +45,16 @@ def save
     end
 
     false
+  rescue ActiveRecord::RecordInvalid
+    self.authentication_method = nil
+    if one_time_password.nil?
```
anaulin, line 50: I suppose you could extract lines 38-44 into a method called, say `verify_or_resend_otp`, and then this `save` method would become something like:
```ruby
def save
  return false if !valid? || !actionable?
  verify_or_resend_otp
rescue ActiveRecord::RecordInvalid
  # Retry in case of race condition
  self.authentication_method = nil
  verify_or_resend_otp
end
```

app/models/authenticated_session.rb
```diff
@@ -45,6 +45,16 @@ def save
     end
 
     false
```
anaulin, line 47: I know this is not from this PR, but I think this `false` never did anything -- the if/else above will always return before this line is executed.

anaulin (approved):
Looks good!

This makes me wonder if it might be worth it adding a "disable after submit" thing to some (all?) of our submit buttons. It's a little tricky, though, to make sure that we don't end up with disabled buttons on e.g. a form with errors.

app/models/authenticated_session.rb
zspencer: Oh that's a good point!

app/models/authenticated_session.rb
zspencer: That's a better  name than what I came up with on my own; so I'll do that.

zspencer:
> And for completeness: a different approach, if we think this is due to simultaneous submissions, we could consider using some kind of locking mechanism. Seems overkill for now, though.

This is due to simultaneous *first-time* submissions. I.e. if you attempt to sign in as ana+new-user@example.com for the first time and double-click the button; the error will occur.

It does not appear to occur with second-times; as the authentication method is already in the database so double-clicking will just double-email.

> This makes me wonder if it might be worth it adding a "disable after submit" thing to some (all?) of our submit buttons. It's a little tricky, though, to make sure that we don't end up with disabled buttons on e.g. a form with errors.

I could have sworn there was something like that built into rails, but maybe it was part of ujs and isn't in turbo yet...

anaulin:
> I could have sworn there was something like that built into rails, but maybe it was part of ujs and isn't in turbo yet...

Me too, and I googled a bunch and couldn't find anything. Something about "disable with"??? But couldn't find it. 🤷🏼‍♀️

Issue #1841, opened 2023-09-21
🐞 `Section`: Cannot be removed


KellyAH:
- https://github.com/zinc-collective/convene/issues/1155

Replicated on Prod: https://sandbox.zinc.coop/spaces/sandbox/edit
and Local: `http://localhost:3000/spaces/<space_name>/edit`

Given a space has at least 1 section.
1. Go to any space edit page: `/spaces/<space_name>/edit`
2. Click on any section to go to the section details page `/spaces/<space_name>/rooms/<room-name>/edit`
3. click "Remove Section" button
4. see "You've updated Section '<room-name>' successfully! 🎉"

expected behavior
* Section is deleted and no longer displays on the space edit page

actual behavior
* Section fails to be deleted and still displays on the space edit page

![section fail to be deleted](https://github.com/zinc-collective/convene/assets/16140873/1100c8df-3e7e-448f-a7d1-d03f87cd6bb2)

zspencer:
Oof, this is not ideal. I will nab it 🔜

zspencer:
@KellyAH OK! This is fixed and the fix should be deployed! Can you give it a whirl?

KellyAH:
> @KellyAH OK! This is fixed and the fix should be deployed! Can you give it a whirl?

Confirmed the issues is fixed on prod. TY for fixing it!
Tested on Sandbox space: https://sandbox.zinc.coop/edit w/ user k_ahong@yahoo.com.
✅ can delete a section if it has gizmos or not.
✅ if a section has any gizmos, when clicking  the`Remove Section` button on the section details page, a confirmation pop-up displays.
✅ if no gizmos exist for a section, after clicking `Remove Section` button on the section details page, a confirmation pop-up does not display.
✅ can create new sections and delete them.

PR #1900, opened 2023-10-19
🐞🔨 `Section`: Deleting Sections is possible again

zspencer:
- https://github.com/zinc-collective/convene/issues/1841

So, this was caused by me being FANCY and putting the delete button in the same form as the edit and update. A bad idea.

Now, deleting a Section is handled in it's own independent form, rather than attempting to rewrite the forms hidden `_method` field to trick Rails into doing a destroy.

This is also better from a design perspective because:

A) omg, a delete button so close to the save button is DANGEROUS
B) Deleting a Section is kinda a big deal, and while it would be better  to bubble-wrap the heck out of it so people who *do* delete a  section don't wind up in a sad-mad state; we can at least give them  lots of "DANGER WILL ROBINSON DANGER" messaging. 

This is... irresponsible of us but what we can do now. If a motivated bystander or contributor wanted to design and implement a way for Sections to be Archived instead; I would be Very Happy :tm:

### DIFF
--- a/app/views/rooms/edit.html.erb
+++ b/app/views/rooms/edit.html.erb
@@ -20,3 +20,17 @@
     <%= render "furnitures/new", furniture: new_furniture %>
   <%- end %>
 <% end %>
+
+
+<%- if policy(room).destroy? && room.persisted? && !room.entrance? %>
+  <div class="m-5">
+    <h2>Remove this Section?</h2>
+    <%- data = {} %>
+    <%- if !room.gizmos.reload.empty? %>
+      <p>Warning! Deleting this Section will also delete all Gizmos and their related data. Please be very sure you want to do this! We recommend making the Section "internal" instead.</p>
+      <%- data[:turbo_confirm] = I18n.t("rooms.destroy.confirm", room_name: room.name) %>
+    <%- end %>
+
+    <%= button_to(t('rooms.destroy.link_to'), room.location, data: data, method: :delete, class: "--danger w-full") %>
+  </div>
+<%- end %>
--- a/spec/system/sections_spec.rb
+++ b/spec/system/sections_spec.rb
@@ -1,0 +1,53 @@
+require "rails_helper"
+
+RSpec.describe "Sections" do
+  describe "Removing a Section" do
+    let(:space) { create(:space, :with_entrance, :with_members, member_count: 1) }
+
+    before do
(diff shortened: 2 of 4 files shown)


app/views/rooms/edit.html.erb
```diff
@@ -20,3 +20,17 @@
     <%= render "furnitures/new", furniture: new_furniture %>
   <%- end %>
 <% end %>
+
+
```
anaulin, line 24: Nit: unnecessary blank line.

app/views/rooms/edit.html.erb
```diff
@@ -20,3 +20,17 @@
...
+<%- if policy(room).destroy? && room.persisted? && !room.entrance? %>
+  <div class="m-5">
+    <h2>Remove this Section?</h2>
+    <%- data = {} %>
+    <%- if !room.gizmos.reload.empty? %>
+      <p>Warning! Deleting this Section will also delete all Gizmos and their related data. Please be very sure you want to do this! We recommend making the Section "internal" instead.</p>
```
anaulin, line 30: Ideally these strings would be I18n-ed, but I don't think that's a blocker for merging this.

Good call on this warning and on having a `confirm` on the delete button.

spec/system/sections_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      context "when the section has Gizmos" do
+        before {
+          create(:furniture, room: section)
+          refresh
+        }
+
+        # Design note: It would be far better for us to have a way to safely undo
+        # the deletion of a Section, or even put the Gizmos into a holding space
+        # or something to be re-assigned; but that is out of scope for me at the
+        # moment - ZS 10/18/23
```
anaulin, line 41: Eventually: use the `discard` gem or similar?

anaulin (approved):
Thank you for the 🐛🔨!

spec/system/sections_spec.rb
zspencer: Yea!

app/views/rooms/edit.html.erb
zspencer: True, ugh

app/views/rooms/edit.html.erb
zspencer: I wish we had an erb linter that worked

PR #1909, opened 2023-10-26
👷 Refactors existing Square logic to a new `SquareOrder` service class

rosschapman:
## Issue

* https://github.com/zinc-collective/convene/issues/1500

## Goals
1. Add additional test coverage for a Guest checkout
1. The new class should expose a single public method for executing the workflow to create an order in square. For example, from inside `stripe_events_controller.rb` or elsewhere 

	```ruby
	square_order = SquareOrder.new(order, ???)
	square_order.send_to_square_seller_dashboard
	```

## Notes
- 🙋 What initial data does a `SquareOrder` instance need to do it's work? 
  - order, marketplace, ordered_products
  - just an `order` should suffice to query all necessary data


app/furniture/marketplace/order.rb
rosschapman, line 1: Relocates all Square related behavior to the new `Marketplace::SquareOrder` class.

app/furniture/marketplace/square_order.rb
rosschapman, line 1: Most of this code was previously reviewed in https://github.com/zinc-collective/convene/pull/1733. Reviewers should not some key differences: 
1. All membership data can be queried through the `order` passed in on initialization
2. Class encapsulation means wore work is now done in initialization to prepare data as instance variables for use in member methods

spec/factories/furniture/marketplace.rb
rosschapman, line 1: This sketches a theoretical use of a live sandbox for testing.

spec/furniture/marketplace/square_order_spec.rb
rosschapman, line 1: Sketches of unit tests for complex internal methods of the new `SquareOrder` service class. I'm thinking about this class in a way similar to an external library, therefore it feels justified to test complex _private_ methods that are critical to the correct transmission of data.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman, line 1: Because of additional coverage elsewhere, the tests here only have to concern themselves with the correct invocation of the _publicly_ exposed interface of the `SquareOrder` class.

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -24,7 +24,8 @@ def create
         order.events.create(description: "Payment Received")
 
         if marketplace.square_order_notifications_enabled?
-          order.send_to_square_seller_dashboard
+          square_order = SquareOrder.new(order)
```
rosschapman, line 27:
```suggestion
          square_order = Marketplace::SquareOrder.new(order)
```

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -24,7 +24,8 @@ def create
         order.events.create(description: "Payment Received")
 
         if marketplace.square_order_notifications_enabled?
-          order.send_to_square_seller_dashboard
+          square_order = Marketplace::SquareOrder.new(order)
```
zspencer, line 27: This controller action is getting nasty; and that's 99.999% my fault. To help reduce the sprawl, would you be available to encapsulate the square stuff a bit more tightly? For example `marketplace.square_connection&.send_order(order)`. 

We could make `marketplace.square_connection` return `nil` in cases where Square Order Notifications are not enabled and a tiny little adapter class (say `Marketplace::SquareConnection`) when there is; eliminating the need for an `if` in the controller and shaving off a few lines.

app/furniture/marketplace/square_order.rb
```diff
@@ -0,0 +1,149 @@
...
+        # TODO: Noop for now
+      end
+    end
+
+    # Square sets max of 43 chars
+    private def square_idemp_key
+      "#{@order.id}_#{8.times.map { rand(10) }.join}"
+    end
+
+    private def create_square_order
```
zspencer, line 47: Could we log an order event before and after attempting to send to Square so that we have a record of attempts to send the Order to Square?

app/furniture/marketplace/square_order.rb
```diff
@@ -0,0 +1,149 @@
...
+      square_create_order_body = prepare_square_create_order_body(@marketplace)
+      @marketplace.square_client.orders.create_order(body: square_create_order_body)
+    end
+
+    # NOTE: Square requires that orders must have be in a state of complete
+    # payment to show up in the Seller Dashboard
+    private def create_square_order_payment
+      square_create_payment_body = prepare_square_create_order_payment_body
+
+      marketplace.square_client.payments.create_payment(body: square_create_payment_body)
```
zspencer, line 57: Same here, just for shits and giggles.

app/furniture/marketplace/stripe_events_controller.rb
zspencer: Also, heads up but you may have noticed this; because this file opens with:

```
class Marketplace

end
```

Constant lookups will *start* by looking for constants within the `Marketplace` namespace. So `Marketplace::SquareOrder` will mostly likely resolve as `Marketplace::Marketplace::SquareOrder`. I don't think it will ever look for `Marketplace::SquareOrder` so it will probably raise a constant undefined error.

You can double check this with `rails zeitwerk:check` which will parse the constant lookup for everything in the Rails autoload path to make sure it's all wired in correctly.

Personally, I like how the Rails constant lookup means changing the namespace lets you "switch" the objects that will resolve; as it feels very smalltalk, but I also find it infuriating due to how inconsistent and opaque it is. Zeitwerk helped a lot but there are still moments where it's just facepalms into a door frame.

zspencer (commented):
I like where this is headed! Just not approving because it's still in draft.

Can't wait to see the tests!

app/furniture/marketplace/stripe_events_controller.rb
rosschapman: Thanks for the tips about constant lookups. I was even reading through https://noelrappin.com/blog/2023/10/better-know-a-ruby-thing-22-constants/ this week to get a better handle on this aspect of Ruby.

rosschapman: > eliminating the need for an if in the controller and shaving off a few lines.
Yes lets do this. I reminded of a Sandi Metz talk about programming before conditionals and how generally one should consider OO encapsulation vs this type of imperative programming. Good to be reminded!

spec/factories/furniture/marketplace.rb
```diff
@@ -65,6 +65,14 @@
       with_delivery_areas
       with_notification_methods
     end
+
+    trait :with_square do
```
zspencer, line 69: Is there supposed to be an Application ID? I'm adding secrets to Github and I've added:
 
1. `MARKETPLACE_VENDOR_SQUARE_ACCESS_TOKEN`
2. `MARKETPLACE_VENDOR_SQUARE_LOCATION`
3. `SQUARE_ENV`
4. `MARKETPLACE_DISTRIBUTOR_APPLICATION_ID`

zspencer: I think the application id is only important if we make it an OAUTH App; which I would love for us to do at some point (so that someone can just add "Delivery by Piikup" to their Square account and magic-pony complete!)

app/furniture/marketplace/square_order.rb
```diff
@@ -0,0 +1,167 @@
...
+        idempotency_key: square_idemp_key,
+        amount_money: {
+          amount: @order.product_total.cents,
+          currency: "USD"
+        },
+        order_id: @square_order_id,
+        location_id: @square_location_id,
+        external_details: {
+          type: "OTHER",
+          source: "CONVENE_SYSTEM_PAYMENT for Space #{@space.id}"
```
zspencer, line 157: I would say something like `Paid by Stripe (Charge #{order.stripe_payment_id}) via #{space.name} (#{space.id})` so we have more context than just the space id and we can use that to find the Stripe charge later.

app/furniture/marketplace/square_order.rb
zspencer: To clarify, I mean record an `order.events` object in the database; not just a Rails logger.

app/furniture/marketplace/square_order.rb
```diff
@@ -0,0 +1,167 @@
...
+      square_create_payment_body = prepare_square_create_order_payment_body
+
+      @square_client.payments.create_payment(body: square_create_payment_body)
+    end
+
+    # NOTE: Square requires that orders must include fulfillments to display
+    # in the Seller Dashboard
+    # See: https://developer.squareup.com/docs/orders-api/create-orders
+    private def prepare_square_create_order_body
+      location_id = @marketplace.settings["square_location_id"]
```
zspencer, line 82: I believe this is available on the `marketplace` as `marketplace.square_location_Id`

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -23,9 +23,7 @@ def create
         order.update!(status: :paid, placed_at: DateTime.now, payment_processor_fee_cents: balance_transaction.fee)
         order.events.create(description: "Payment Received")
 
-        if marketplace.square_order_notifications_enabled?
-          order.send_to_square_seller_dashboard
-        end
+        marketplace.square_connection&.send_order
```
zspencer, line 26: Very tiny!

zspencer (commented):
I've added the credentials for a Zinc Collective LLC Square Account and read through the code. 

I think this is a good enough step forward; and left some stylistic feedback around:

- Using `delegate` and `attr_accessor` to reduce the amount of setup on class instantiation
- Including more context when sending information to the Square POS
- Storing significant events (Square Order, etc) in the database so that they are visible to the Marketplace Vendor and Distributor; rather than just developers.

app/furniture/marketplace/square_order.rb
```diff
@@ -0,0 +1,167 @@
...
+      @order = order
+      @marketplace = order.marketplace
+      @shopper = order.shopper
+      @space = order.marketplace.space
+      @ordered_products = @order.ordered_products
+      @square_location_id = @marketplace.settings["square_location_id"]
+      @square_order_id = nil
+      @square_payment_id = nil
+      # @state = "READY"
+    end
```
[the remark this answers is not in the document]
rosschapman: TIL 🙏🏻

app/furniture/marketplace/square_order.rb
```diff
@@ -0,0 +1,167 @@
...
+
+    # Sends an order to Square. Will create both an order and payment in the
+    # receiving account which are requirements for the order to show up in the
+    # Seller Dashboard.
+    def send_order
+      Rails.logger.info("Start creating Square order")
```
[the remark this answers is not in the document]
rosschapman: Ah ok. A blend of misinterpreting your guidance and forgetting we had an event system implemented. Let's do it.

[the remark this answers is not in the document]
rosschapman: @zspencer 

What's the thinking around what strings go into the `public.marketplace_events.description` column since it's a "free" `text` type?  Understandably with a small org code review serves as a guardrail to maintain consistent event descriptions/actions. But would the list of accepted descriptions strings be something we would consider declaring in some kind of _enum_ somewhere so developers could easily consult an existing set? 

I already notice we use a typical pattern of Title Case, past tense, and Noun + Verb. What about adding a small style guide comment to the `Event` model?

app/furniture/marketplace/square_order.rb
rosschapman: Yeah, not sure why I did that.

spec/factories/furniture/marketplace.rb
rosschapman: It's not necessary for instantiating a Square Client class.

app/furniture/marketplace/stripe_events_controller.rb
rosschapman: It is. Although I'd still prefer a more object-oriented, explicit exposure and/or handling of `nil` cases. 😅

rosschapman:
@zspencer I'm going to follow up with a separate PR for the tests to reduce noise in this PR. See https://github.com/zinc-collective/convene/pull/1936. Therefore consider what's already here ready for any final review. I suspect it may not be perfect so we can also work out some details on the next PR if required.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -100,16 +88,20 @@
...
       it "does not attempt to transfer the order to seller's Square dashboard" do
         call
-        expect(square_client).not_to(have_received(:orders))
+        expect(square_order).not_to(have_received(:send_order))
       end
     end
 
     context "when Square notifications are enabled" do
-      let(:order) { order_for_square }
+      let(:marketplace) { create_marketplace_with_square }
```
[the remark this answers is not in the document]
rosschapman: Will follow up in https://github.com/zinc-collective/convene/pull/1936

app/furniture/marketplace/square_order.rb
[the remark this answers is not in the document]
zspencer: I think a style guide is probably better than an enum; but if you want to gather the usages I wouldn't be opposed.

zspencer (approved):
Once you get the tests passing let's shipit!

PR #1920, opened 2023-11-01
Removes blank option from Delivery Areas dropdown on Marketplace

rosschapman:
**UPDATE 11/7/23:** See discussions below for decision history on this issue, including copy updates. In brief, the `select` partial has been replaced by a new `SelectComponent. Also I have spot tested the application and dropdown selection seems to be working as expected. 

---
Not sure if we meant to add a prompt or not? 

image


anaulin:
Yeah, I noticed this too. I think the ideal solution here would be:
1. If there is only 1 delivery area option, don't show the dropdown at all. Instead, set the delivery area for the order automatically (e.g. maybe make it a hidden field with the one area id).
2. If there is more than one delivery area, show the dropdown with a prompt.

Since adding the "default to the area if there is only 1" is probably much bigger of a change that you were planning for, an intermediate step would be: if there is more than 1 option, show the prompt (maybe "-- Select delivery area --"), and if there is only 1 option, just show that one (and it will be "selected" by default).

anaulin (approved):
I think this is an improvement and safe to merge.

If you're feeling extra, see my previous comment on having different behaviors when we have 1 option versus multiple.

Thank you!

rosschapman:
> If you're feeling extra, see my previous comment on having different behaviors when we have 1 option versus multiple.

@anaulin I'm feeling extra for this one. I think your suggestion makes sense. What do you think about standardizing this behavior in our [select](https://github.com/zinc-collective/convene/blob/main/app/views/application/_select.html.erb#L6-L7) partial? Even more, refactoring to a View Component?

zspencer:
In another codebase; I've played around with using a custom form builder rather than ViewComponents or Partials for defining how forms work. The advantage of that is:

- You can use `form_with` and `form_for` without any nosense

rosschapman:
It could be the limit of my understanding of View Components but my mental model is that they are like partials on a syntactic sugar high, but ultimately render out some data into html yada yada. Therefore passing through the correct `FormBuilder` `form` object to a View Component like this

```ruby
<%= render(SelectDropdownComponent.new(options: cart.marketplace.delivery_areas.pluck(:label, :id), include_blank: true, attribute: :delivery_area_id, form: form, skip_label: true) ) %+>
```

would behave unsurprisingly like passing the same `form` object as we do for our `select` partial like

```ruby
<%= render "select", options: cart.marketplace.delivery_areas.pluck(:label, :id), include_blank: true, attribute: :delivery_area_id, form: form, skip_label: true %>
```

But the resulting benefit is increased reusability/standardization of select dropdown behavior with a View Component.

rosschapman:
> I don't know how well ViewComponent works when embedded within the FormBuilder, which is how I would prefer we write our more complex UI elements...

I'll take some time to play with nesting a View Component in a `form_with` block. In my mental model VCs are just partials on a syntactic sugar high, and passing through a `form` object should work similarly/unsurprisingly?

rosschapman:
@anaulin @zspencer the latest code adds a new `SelectComponent` to experiment with View Components nested in a `FormBuilder` helper. I also experimented with how to display the `DeliveryArea` to the shopper if the there is only a single `DeliveryArea` associated to the `Marketplace`. This seemed like a better approach to having a `select` tag wrapper component figure out display logics. 

image

app/components/select_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<div>
+  <% unless @skip_label %>
+    <%= @form.label @attribute %>
+  <% end %>
+  <%= @form.select @attribute, @choices, @options, @html_options %>
+  <%= render partial: "error", locals: { model: @form.object, attribute: @attribute } %>
```
rosschapman, line 6: 👍🏻 Partial lookup seems to work fine from within a View Component

zspencer:
I think it's fine to start with a ViewComponent; but where we want to go is to get away from needing people to go "off the rails" so-to-speak when making forms (either by instantiating a `ViewComponent` *or* using a `render partial: form/{field}`.)

The way to do that is to write a [Custom `Marketplace::FormBuilder`](https://medium.com/@mmigliore311/using-a-custom-formbuilder-ab7f1ae65f2b) and overload the `select` method definition. It's entirely possible the `Marketplace::FormBuilder#select` method body will literally be `render SelectComponent.new(...)`; since I think it's fine to use a `ViewComponent` for the html-building side of things. 

Then, we can define a `form_with` helper method (for now, isolated to the `Marketplace`, perhaps on the `Marketplace::Controller`) which inserts the `Marketplace::FormBuilder` for the `builder` argument.

That way whenever we want to make forms in the `Marketplace` (and eventually throughout Convene) we can do the Rails-default thing of `form_with`. 

It may make more sense to pull it out of the `Marketplace` namespace; but I figure keeping it tightly scoped initially; and then promoting it to the top-level once we want to use it elsewhere would be the safest, smallest step forward.


TL/DR: I would love it if we could ditch `render partial: "form/select"` and `render SelectComponent` from our templates; so that we stay "on the Rails".

app/components/select_component.rb
```diff
@@ -0,0 +1,12 @@
+class SelectComponent < ApplicationComponent
+  def initialize(config, **kwargs)
+    @form = config[:form]
+    @attribute = config[:attribute]
+    @choices = config[:choices] || []
+    @options = config[:options] || {}
+    @html_options = config[:html_options] || {}
+    @skip_label = config[:skip_label] || true
```
zspencer, line 8: I think we want this to default to `false`, rather than true. Also, make sure to use [`Hash#fetch`](https://ruby-doc.org/3.2.2/Hash.html#method-i-fetch), rather  than `||`; because if `config[:skip_label]` is `false` then it will become `true`; whereas with `fetch` it only returns the default value if no value is set.

Another option would be to `def initialize(form:, attribute:, choices: [], options: {}, html_options: {}, skip_label: false)` so that each argument's defaults are part of the method definition.

app/furniture/marketplace/cart/delivery_area_component.html.erb
```diff
@@ -4,18 +4,20 @@
...
       </div>
-
       <div class="text-right">
         <%= link_to(t('marketplace.cart.delivery_areas.edit.link_to'), cart.location(:edit, child: :delivery_area)) %>
       </div>
     <%- else %>
-      <%= render "marketplace/cart/delivery_areas/form", cart: cart %>
+      <%- if cart.marketplace.has_single_delivery_area? %>
+        <div><%= single_delivery_area_label %></div>
+        <small>If you are ordering from outside this area then 🤷</small>
```
zspencer, line 17: Perhaps "Orders with delivery addresses outside of our delivery area will be canceled." or something

app/furniture/marketplace/cart/delivery_area_component.html.erb
```diff
@@ -4,18 +4,20 @@
...
         </div>
-
       </div>
-
       <div class="text-right">
         <%= link_to(t('marketplace.cart.delivery_areas.edit.link_to'), cart.location(:edit, child: :delivery_area)) %>
       </div>
     <%- else %>
-      <%= render "marketplace/cart/delivery_areas/form", cart: cart %>
+      <%- if cart.marketplace.has_single_delivery_area? %>
```
zspencer, line 15: I'm curious about why you chose to define `has_single_delivery_area?` (note: Ruby prefers `single_delivery_area?` over `has_...` on the `marketplace` while the `single_delivery_area_label` lives on the component?

Since both are really only used here, I tend to keep my method definitions as close to their sole caller as possible, and then "promote" methods that want to be used elsewhere.

app/components/select_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<div>
```
zspencer, line 1: Something to think about is we now have *two* places that we are defining how a `select` field should work.

My preference would be to render the partial here directly; for now; and then inline that partial once we've either:

A) replaced all the `render partial: "select"` with `render SelectComponent`... _or_
B) Did the fancy-pants Custom Form Builder thing I described in the initial comment; and replaced all the `render partial: 'select'` with `f.select`.

zspencer (approved):
I think the removal of the blank option and the branch for marketplaces that have only one delivery area is a huge win!

I think getting a `SelectComponent` is a good step as well; and while I would like it if we didn't duplicate the `select` partial and invert the logic `skip_label` logic, I don't think those are blocking concerns.

Most of my other comments are "here are the steps I would take to make this really shine!

app/components/select_component.rb
rosschapman: Oh good call. I actually confused myself into thinking that those arguments might come in as strings (although I even defaulted to a `true` boolean. But I had actually checked their type and forgot to come back and change these.

app/furniture/marketplace/cart/delivery_area_component.html.erb
rosschapman: Good suggestion, I agree with putting this method on the component. I just didn't think through it very hard before comitting the code because I implemented the _idea_ on `marketplace` before I knew where I was putting it in the UI code. At the time `marketplace` felt "closer" in my brain. :)

app/components/select_component.html.erb
rosschapman: Yeah that's fair. @zspencer I'm always curious what your restrained, preferred incremental design approach will be and what diverging paths you will suggest for the code I'm madly slinging around. 😅 

Doing `A)` is not a heavy lift at the moment (though let's see how good our test coverage is to catch any mistakes I make!). I designed this select component as a similar base layer component to replace the current partial so that's a reasonable step forward. There aren't too many spots to find/replace.

rosschapman:
> TL/DR: I would love it if we could ditch render partial: "form/select" and render SelectComponent from our templates; so that we stay "on the Rails".

@zspencer Oh, custom form builders. That makes sense. Form libraries scare me coming from the JS ecosystem where there are too many choices and patterns. Also, it's almost rare that reuse works in the long run. For example, _lets use this shared form code for both "edit" and "create" forms_ -- cue eventual cumbersome refactor! This is why I'll always build up base form components first before stuffing them in a general solution. 

That said, FormBuilder is intrinsic to Rails so it's comforting to be able to build on top of that abstraction with confidence it's the blessed path.

app/furniture/marketplace/cart/delivery_area_component.html.erb
```diff
@@ -12,7 +12,7 @@
         <%= link_to(t('marketplace.cart.delivery_areas.edit.link_to'), cart.location(:edit, child: :delivery_area)) %>
       </div>
     <%- else %>
-      <%- if cart.marketplace.has_single_delivery_area? %>
+      <%- if single_delivery_area? %>
         <div><%= single_delivery_area_label %></div>
         <small>If you are ordering from outside this area then 🤷</small>
```
rosschapman, line 17:
```suggestion
        <small>Orders with delivery addresses outside of this delivery area will be subject to cancellation.</small>
```

app/components/select_component.html.erb
zspencer: Perfect! We could always go *full* ouroboros and replace the *body* of the render partial with `render SelectComponent` 🤣 😭

app/furniture/marketplace/cart/delivery_area_component.html.erb
zspencer: Now that I see it, "Orders outside of this location will be subject to cancellation." is even terser. Ugh writing for applications is hard.

app/furniture/livestreams/_form.html.erb
```diff
@@ -1,6 +1,19 @@
...
   <%= render "text_field", form: furniture_fields, attribute: :channel %>
-  <%= render "select", form: furniture_fields, attribute: :layout, options: ['video', 'video_with_chat'], include_blank: false %>
-
-  <%= render "select", form: furniture_fields, attribute: :provider, options: ['twitch'], include_blank: false %>
+  <%= render SelectComponent.new({
+      choices: ['video', 'video_with_chat'],
+      attribute: :layout,
+      form: furniture_fields,
+      options: {
+        include_blank: false,
```
zspencer, line 8: Heads up, [`include_blank` defaults to false](https://api.rubyonrails.org/classes/ActionView/Helpers/FormTagHelper.html#method-i-select_tag), so you don't need to set it explicitly.

app/components/select_component.rb
zspencer: Cool, does this mean you're going to lean into Ruby's assignment syntax rather than the `||` operator? As it currently stands, `SelectComponent.new(skip_label: false)` will become `@skip_label # true`.

zspencer (approved):
Great! Let's roll with it! And we should be able to ditch `the `app/views/application/_select.html.erb` now!

rosschapman:
Just discovered a bug because we rely on physically selecting a delivery area to apply the delivery area to the cart. My new UX has far reaching implications! Considering the new design, I will have to figure out a way to set the delivery area for carts when loading the page.

rosschapman:
@zspencer and I have figured out a path forward for creating the cart with a delivery area if only a single delivery area is associated to a marketplace: https://github.com/zinc-collective/convene/pull/1920/commits/629b53a0252d09af669159ee107b74b36c19e887.

rosschapman:
Btw @zspencer if the current state of this changeset suffices, great. I've also got a fast follow-up with some view specs for the new conditional logic [here](https://github.com/zinc-collective/convene/pull/1926/files?diff=unified)

app/components/select_component.rb
```diff
@@ -0,0 +1,12 @@
+class SelectComponent < ApplicationComponent
+  def initialize(config, **kwargs)
+    @form = config.fetch(:form)
+    @attribute = config.fetch(:attribute)
+    @choices = config.fetch(:choices, [])
+    @options = config.fetch(:options, {})
+    @html_options = config.fetch(:html_options, {})
+    @skip_label = config.fetch(:skip_label, true)
+
+    # super(**kwargs)
```
rosschapman, line 10:
```suggestion
    super(**kwargs)
```

PR #1936, opened 2023-11-13
🥗 `Marketplace`: Test `Seller` receives `Order` in their `Square` Point of Sale

rosschapman:

Follow up to https://github.com/zinc-collective/convene/pull/1909


app/furniture/marketplace/square_idempotency_key.rb
```diff
@@ -0,0 +1,7 @@
+class Marketplace
+  class SquareIdempotencyKey
+    def self.generate(order_id)
+      "#{order_id}_#{8.times.map { rand(10) }.join}"
```
zspencer, line 4: You can also use `SecureRandom.hex(8)` to generate 8 random hex characters

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -81,6 +82,113 @@
...
     end
+
+    context "when Square notifications are not enabled" do
+      let(:marketplace) { create(:marketplace) }
+      let(:person) { create(:person) }
+      let(:shopper) { create(:marketplace_shopper, person: person) }
+      let(:order) { create(:marketplace_order, :with_products, status: :pre_checkout, marketplace: marketplace, shopper: shopper) }
+
+      it "does not attempt to transfer the order to seller's Square dashboard" do
+        square_order_double = instance_double(Marketplace::SquareOrder, send_order: true)
```
zspencer, line 93: I would probably test the negative case against the `fake_square`; because right now there's not really any way for this test to fail; since the `square_order_double` is local to this block; and I'm not seeing a place where you're injecting this double into the underlying call stack.

For example:

```rb
fake_square = instance_double(Square::Client, :orders)`
allow(Square::Client).to receive(:new).and_return(:fake_square)
call
expect(fake_square).not_to have_received(:orders)
```

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
```diff
@@ -81,6 +82,113 @@
...
+              line_items: [{
+                base_price_money: {
+                  amount: order.ordered_products.first.product.price.cents,
+                  currency: "USD"
+                },
+                item_type: "ITEM",
+                name: order.ordered_products.first.name,
+                quantity: "1"
+              }],
+              location_id: ENV.fetch("MARKETPLACE_VENDOR_SQUARE_LOCATION", "square_location_id_1234"),
```
zspencer, line 169: Instead of coupling to `ENV`, can we get this value from the `marketplace`?

zspencer (approved):
This test looks great! I'd love more permutation tests deeper in the stack; but having a shot-across-the-bow in the request specs that confirms we're interacting with the square gem correctly is highly valuable in it's own right.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman: I'm not sure I understand the difference, mainly because I just care about testing the conditional logic in `marketplace.square_connection` and this test achieves that goal. Are you interested in also covering failure mode(s) in instantiation of the `SquareOrder`?

rosschapman:
> This test looks great! I'd love more permutation tests deeper in the stack; but having a shot-across-the-bow in the request specs that confirms we're interacting with the square gem correctly is highly valuable in it's own right.

Yep! This is definitely step 1 of N toward confident coverage. And it helped us clean up the interface between the controller/caller code and the Square subsystem of classes.

app/furniture/marketplace/square_idempotency_key.rb
rosschapman: Oh, me likey.

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
rosschapman: Of course!

spec/furniture/marketplace/stripe_events_controller_request_spec.rb
zspencer: If you try and make your current test fail; It can't.  At no point does the fake get accessed by the method-under-test.

If you don't want to switch to testing against the `Square::Client` (the "inner-most-outer-edge" of the seam between Convene and Square); you would still need to do something like:


```rb
square_order_double = instance_double(Marketplace::SquareOrder, send_order: true)
# Inject our double into the call stack
allow(Marketplace::SquareOrder).to receive(:new).and_return(square_order_double) 
call
expect(square_order_double).not_to have_received(:send_order)
```



Both are perfectly acceptable;

rosschapman: Ehh, I'm still getting things confused because of JavaScript and forgetting about needing to invoke `new` and _allowing_ that into the call stack (so to speak). 

I'm not as concerned with testing down to the `Square::Client` with this negative test since we don't even get that far.

anaulin:
It looks like the tests added in this PR are failing in CI on `main`:
image
https://github.com/zinc-collective/convene/actions/runs/7108075091/job/19350699926

zspencer:
Oh I think I know what's happening! Dependabot doesn't get the same secrets that the contributors have access to. I can't remember what I did with Stripe; but it was either:

A) I skipped the tests when the environment variables weren't set, or
B) I gave dependabot the secrets
C) I designed the tests to pass without Stripe (Which I don't think is true because the system test definitely depends on Stripe working? Maybe?)

zspencer:
I added the secrets from `actions` to `dependabot` and I think that fixes it! https://github.com/zinc-collective/convene/settings/secrets/dependabot

rosschapman:
Oh boy, thanks @zspencer for the rescue there.

zspencer:
Computaaaahs

PR #1901, opened 2023-10-19
Update Rails to 7.1.1

anaulin:
This brings us to the latest version of Rails.

A scan of the [release notes](https://github.com/rails/rails/releases) didn't surface any breaking changes.

PR #1957, opened 2023-11-23
Update `Gemfile` and `schema.rb` files to reflect the version of Rails we're using

anaulin:
We upgraded to Rails 7.1 in https://github.com/zinc-collective/convene/pull/1901, but the migrations file was not updated, so if you run all migrations on a clean branch you get this diff to update the schema version in `schema.rb`. This PR fixes things so that there are no schema changes on a clean app install.

Also took the opportunity to tweak the Rails version constraint in the Gemfile, for added clarity.


zspencer (approved):
:swoon: Thank you!!!!

Issue #1956, opened 2023-11-23
✨ `Marketplace`: Printable `Flyers`


zspencer:
- As part of https://github.com/zinc-collective/convene/issues/1955 it would be lovely if `Vendors` or `Distributor` could print something to stick to their windows that indicates that they offer delivery for online orders.

## Design

- A half-page design By [https://spanner.works/@bouncinglime](https://spanner.works/@bouncinglime)
  - [Print Quality Flyer (Example Piikup)(pdf)](https://github.com/zinc-collective/convene/files/13445435/piikup-flier.pdf)
  - [Print Quality Flyer (Example Piikup)(svg)](https://github.com/zinc-collective/convene/assets/50284/136382be-1dfb-46d0-91cb-460e1d473438)

PR #1958, opened 2023-11-23
`Marketplace`: Very Rough  `Flyer`

zspencer:
- https://github.com/zinc-collective/convene/issues/1956

This gives us a QR-Code Flyer, but we should probably ditch the footer if we can...

I sprouted a "nude" layout (Ok it needs a better name) that gives us an empty page and breaks us out of turbo-frame land.

Anyway, to make pretty we need to:

- Attach a `Vendor`'s logo to the `Space`
- Attach a `Distributor`s logo to the `Marketplace`
- Actually work towards the fancy-pants design that Mari made!

But in the meantime, this is something that could exist and maybe even could get better.

[Zee's Pizza.pdf](https://github.com/zinc-collective/convene/files/13445705/Zee.s.Pizza.pdf)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,8 @@ gem "gretel", "~> 4.6"
 # Better UI components
 gem "lookbook", ">= 2.0.0.beta.4"
 gem "view_component", "~> 3.7"
+# QR Code GeneratioN!
+gem "rqrcode", "~> 2.2"
 
 # Pagination!
 gem "pagy", "~> 6.2"
--- a/Gemfile
+++ b/Gemfile
@@ -43,7 +43,7 @@ gem "gretel", "~> 4.6"
 # Better UI components
 gem "lookbook", ">= 2.0.0.beta.4"
 gem "view_component", "~> 3.7"
-# QR Code GeneratioN!
+# QR Code Generation!
 gem "rqrcode", "~> 2.2"
 
 # Pagination!
--- a/app/views/layouts/nude.html.erb
+++ b/app/views/layouts/nude.html.erb
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title><%= page_title %></title>
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <%# Env variables being made accessible to JavaScript %>
+    <%= tag :meta, name: :sentry_dsn, content: ENV["SENTRY_DSN"] %>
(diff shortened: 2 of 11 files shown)


Gemfile
```diff
@@ -43,6 +43,8 @@ gem "gretel", "~> 4.6"
 # Better UI components
 gem "lookbook", ">= 2.0.0.beta.4"
 gem "view_component", "~> 3.7"
+# QR Code GeneratioN!
```
anaulin, line 46:
```suggestion
# QR Code Generation!
```

app/views/layouts/nude.html.erb
anaulin, line 1: Let's maybe call this something like "simple" or "minimal"?

app/views/layouts/nude.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    <%= tag :meta, name: :release_tag,
+        content: "#{ENV.fetch("APP-BASE", "convene")}@#{ENV.fetch("HEROKU_RELEASE_VERSION", "??")}" %>
+
+    <%= csrf_meta_tags %>
+    <%= csp_meta_tag %>
+    <%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>
+    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
+    <%= render "layouts/hotjar" %>
+    <meta name="turbo-visit-control" content="reload">
+  </head>
```
anaulin, line 17: Might be worth extracting `<head>...</head>` into a partial, and reusing it across layouts.

anaulin (approved):
Great start, thank you!

We could also consider generating this as a PDF (e.g. using the `prawn` gem or similar), which would free us from having to use an alternate HTML layout, and would give us more deterministic layout and formatting for printing purposes.

zspencer:
Yea, I was thinking it would be great to add a `prawn` pdf but did not want to go down that path just yet. That said, I think it's worth doing next Wednesday as an incremental improvement to this patch!

app/views/layouts/nude.html.erb
zspencer: Ooo good point.  We'll need to maybe add in a way to pass in extra nonsense but otherwise it should be good.

app/views/layouts/nude.html.erb
zspencer: minimal is good, I think

PR #1963, opened 2023-11-24
🧹 `Components`: Rename `nude` to `minimal`

zspencer:

Sometimes there are pages where we don't want to include anything but the body; such as in
https://github.com/zinc-collective/convene/issues/1956

This renames that layout from `nude` to `minimal`


anaulin (approved):
👍🏼

PR #1964, opened 2023-11-24
✨ `Gizmo`: `SectionNavigation`

zspencer:

Sprouts a tiny Gizmo for linking to Sections. At some point, we'll want to add a UI for picking which `Sections` to include, preferably an `AllowList` so that adding a `Section` doesn't immediately publish it.


anaulin (approved):
I've been wanting to do this for a while, thank you!

Eventually we should replace the automatic section navigation with this.

zspencer:
Agreed, I just figured it was time

PR #1979, opened 2023-11-30
🐞🔨 `Marketplace`: `Buying Products` Fix `Delivery` Form

zspencer:

Turns out, our `Delivery` form was going to a weird spot which may or may not be why we couldn't place orders on Crumble & Whisk

Anyway, now instead of posting to
`/marketplace/carts/:cart_id/delivery.:delivery_id` it will post to `/marketplace/carts/:cart_id/delivery`


anaulin (approved):
Thanks for fixing!

(I feel like we keep tripping over our `location` business, but I don't have a suggestion for improvement at the moment.)

zspencer:
Sadly, this is because of how Rails presumes plural `resource` routing by default rather than singular. We would have had the same problem if we had used `form_with(model: delivery)`.

We could err on the side of over-specification and always include a `url:` argument in `form_with` calls.

PR #1981, opened 2023-11-30
`Marketplace`: The Back button works!

zspencer:

So, TIL about `data-turbo-action=advance` which tells turbo to update the URL when moving from spot to spot.

Now, when navigating through the Marketplace Management and Checkout flow the URL bar actually update! Magic! Ponies! Sparkles! Hooray!


rosschapman:
👏

PR #1982, opened 2023-11-30
🧹 `Api`: Use the new, non-deprecated, openapi* methods when initializing RSwag

anaulin:


zspencer (approved):
oooo fancy we're so conformist.

rosschapman (approved):

PR #1989, opened 2023-12-04
🧹 `Section`: Remove Navigation in favor of `SectionNavigation` Gizmo

zspencer:

Now that we have a Gizmo for navigating between Sections, we don't want to keep it everywhere.


zspencer:
I've gotten rid of the feature tests that cover the presumption that Sections will always have a navigation element. I took the time to use `yarn exec -- cucumber-js  --dry-run --format=usage` to find all the step definitions that are unused and purge them as well.

app/furniture/section_navigation/section_navigations/_section_navigation.html.erb
```diff
@@ -1,10 +1,7 @@
 <div class="mt-3 grid grid-cols-1 gap-5 sm:gap-6 sm:grid-cols-2 lg:grid-cols-4 items-center">
   <% policy_scope(section_navigation.rooms).each do |room| %>
-    <%= link_to [room.space, room], class: "no-underline" do %>
-      <%= render CardComponent.new(
-        data: { access_level: room.access_level, slug: room.slug, model: "room", id: room.id },
-        classes: "group self-stretch hover:bg-orange-50"
-        ) do %>
+    <%= link_to polymorphic_path(room.location), class: "no-underline" do %>
```
zspencer, line 3: I'm not sure why, but `link_to` is including the full url; which I think is related to our over-loading of `url_for`. 

I didn't want to get into it; but it's annoying.

spec/furniture/section_navigation/section_navigation_system_spec.rb
rosschapman, line 1: I'm curious why use a system spec vs a view spec to test rendered content?

rosschapman (approved):
I have one question about the test but on scan the rest looks good. Nice cleanup 🧹

rosschapman (approved):
I have one question about the test but on scan the rest looks good. Nice cleanup 🧹

spec/furniture/section_navigation/section_navigation_system_spec.rb
zspencer: This doesn't have a `ViewComponent` yet and I was doing a lot of other cleanup so I didn't want to add that change too.

Also since I was deleting System specs it felt better to create new system specs. A `ViewComponent` spec would definitely be a better call when we get to more permutations.

rosschapman: Is there a RoR convention (informal | formal) that a "view spec" must be backed by a `ViewComponent`? Or this is a Convene worldview?

anaulin: I'm not aware of there being such a convention in neither RoR at large nor Convene specifically.

That said, I personally don't see the point of view specs, and would much rather do that kind of testing either in smaller component specs, or as part of larger integration/system tests (depending on the level of granularity desired). A view specs to me feels like "testing HTML", which smells funky.

YMMV YDY YOLO

zspencer: Yea, I tend to not write Rails view specs because once they get to the point of complexity that I would want one, I  wrap it into a `ViewComponent`.

Issue #1983, opened 2023-11-30
🐞 `Marketplace`: `Buying Products`: Saving `Delivery Details` on `Checkout` when no `Cart Delivery Area` is set


anaulin:
- https://github.com/zinc-collective/convene/issues/1326

When a `Cart` does not have a `DeliveryArea`, you can still `Checkout`.

However, the `Checkout` screen does not have a way to set the `DeliveryArea`, and if you try to set your `DeliveryAddress` it fails to save.

anaulin:
@zspencer now I can't get the "Checkout" button to work at all in production:

https://github.com/zinc-collective/convene/assets/6729309/da283cfd-630d-4920-9ead-8896eece9976

zspencer:
You have no products; so you can't check out. But we should probably make that clearer...

anaulin:
@zspencer that's not the issue, if you watch the video you'll see that i have products in my Cafe Gabriela and my Crumble carts

anaulin:
After some more testing, I see that if I go to a vendor where I didn't already have a cart, and I select a delivery area, and also add something to my cart, then when I click the "Checkout" button it takes me to the next step.

I think there are at least two issues:
* it is not possible to visually tell the difference between an enabled and disabled "checkout" button, which makes it feel broken when you click on it and nothing happens
* there is something wonky with pre-existing carts, which might be related to the changes in https://github.com/zinc-collective/convene/pull/1992

zspencer:
Oh dang, I saw the first one and the last one but didn't realize Cafe Gabriela had the issue too. It is odd for the Cafe Gabriela; I think it's likely that there are some `Carts` that we should delete now that we've stopped creating multiple `Carts` per `Shopper`. I'll go get rid of the carts in `pre_checkout`.

I've got a PR that removes the Checkout button completely as a kind of "stop-gap" because I didn't want to think about how to present a disabled Checkout button: https://github.com/zinc-collective/convene/pull/1999

zspencer:
I've purged the other `Carts` that were in the `pre_checkout` state; and re-tested and it appears to work now?

anaulin:
Does seem to work now! (There is a separate bug about the delivery area saving not updating the cart, not sure if we have a separate issue to track that.)

zspencer:
I've created an issue https://github.com/zinc-collective/convene/issues/2002

PR #1991, opened 2023-12-04
🐞🔨 `Marketplace`: `Checkout` is prevented without a `Cart::DeliveryArea`

zspencer:
- https://github.com/zinc-collective/convene/issues/1983

Prevents `Shoppers` from visiting the `Checkout` screen until a `DeliveryArea` has been selected; which should prevent the sad bug.

### DIFF
--- a/app/furniture/marketplace/cart.rb
+++ b/app/furniture/marketplace/cart.rb
@@ -54,7 +54,7 @@ class Marketplace
     end
 
     def ready_for_checkout?
-      delivery.details_filled_in? && cart_products.present? && cart_products.all?(&:valid?)
+      delivery_area.present? && cart_products.present? && cart_products.all?(&:valid?)
     end
   end
 end
--- a/app/furniture/marketplace/carts/_footer.html.erb
+++ b/app/furniture/marketplace/carts/_footer.html.erb
@@ -12,7 +12,7 @@
   </tr>
   <tr>
     <td class="text-right py-3.5" colspan="8">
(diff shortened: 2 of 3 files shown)


app/furniture/marketplace/cart.rb
```diff
@@ -54,7 +54,7 @@ def price_total
     end
 
     def ready_for_checkout?
-      delivery.details_filled_in? && cart_products.present? && cart_products.all?(&:valid?)
+      delivery_area.present? && cart_products.present? && cart_products.all?(&:valid?)
```
rosschapman, line 57: I was curious about where this method used but when I scanned the codebase I found no invocations/references?

zspencer: It was not being used, which was why this bug exists. Removing the constraint that it requires address information and adding the call to `ready_for_checkout?` back to the `Checkout` button in [`carts/_footer.html.erb`](https://github.com/zinc-collective/convene/pull/1991/files#diff-7669466e6284058bbba9590f22f1a03283f91386f361314995a1824369349d38R15) is what fixes the bug.

rosschapman: Yeah I see the fix, I was confused about why it wasn't already invoked somewhere and thought maybe something funny was going on. All looks good though.

zspencer: I think it was when we were moving entering the `Delivery Address` to the `Checkout` page the `ready_for_checkout?` would never be true until too late. A smarter fix would have been to change it to checking the `delivery_area` but  ah well.

Issue #1962, opened 2023-11-24
🐞 `Marketplace`: `Buying Products` Switching `Marketplaces` should not destroy your `Cart`


zspencer:
- https://github.com/zinc-collective/convene/issues/1326

During Sutty's Heuristic Analysis, we discovered that if we open a new `Marketplace` the `Cart` from the previous `Marketplace` is gone!

Oops.

zspencer:
I have fixed this! Hooray!

PR #1992, opened 2023-12-04
🐞🔨 `Marketplace`: Don't clear `Cart` when switching `Marketplaces`

zspencer:
- https://github.com/zinc-collective/convene/issues/1962

There was a bug we noticed during the Sutty demo where carts were being cleared.

Turns out, if a `Marketplace` had multiple `DeliveryAreas` we would create a new `Cart` for the shopper every time the page loaded. Womp. Womp.

### DIFF
--- a/app/furniture/marketplace/marketplace_component.rb
+++ b/app/furniture/marketplace/marketplace_component.rb
@@ -14,7 +14,7 @@ class Marketplace
 
     def cart
       delivery_area = marketplace.delivery_areas.first if marketplace.delivery_areas.size == 1
-      @cart ||= carts.create_with(contact_email: shopper.person&.email).find_or_create_by(delivery_area: delivery_area, shopper: shopper, status: :pre_checkout)
+      @cart ||= carts.create_with(contact_email: shopper.person&.email, delivery_area: delivery_area).find_or_create_by(shopper: shopper, status: :pre_checkout)
     end
 
     def delivery_area_component
--- a/spec/furniture/marketplace/marketplace_component_spec.rb
+++ b/spec/furniture/marketplace/marketplace_component_spec.rb
@@ -35,4 +35,42 @@ RSpec.describe Marketplace::MarketplaceComponent, type: :component do
       end
     end
   end
+
+  describe "#cart" do
+    let(:current_person) { create(:person) }
+
+    context "when there is not a Cart for the Shopper in the Marketplace" do
+      it "creates a Cart for the Shopper" do
+        expect { component.cart }.to change { marketplace.carts.count }.by(1)
+        expect(component.cart).to be_persisted
+        expect(component.cart.delivery_area).to eql(marketplace.delivery_areas.first)
+      end
+
+      context "when the Marketplace has multiple Delivery Areas" do
+        it "creates a Cart for the Shopper without a Delivery Area" do
+          create(:marketplace_delivery_area, marketplace:)
+          expect { component.cart }.to change { marketplace.carts.count }.by(1)
+          expect(component.cart).to be_persisted
+          expect(component.cart.delivery_area).to be_nil
+        end
+      end
+    end
+
+    context "when there is a Cart for the Shopper in the Marketplace" do
+      before do
+        marketplace.carts.create(shopper: Marketplace::Shopper.create(person: current_person))
+      end
+
+      it "doesn't create another cart" do
+        expect { component.cart }.not_to change { marketplace.carts.count }
+      end
+
+      context "when the Marketplace has multiple DeliveryAreas" do
+        it "doesn't create another cart" do
+          create(:marketplace_delivery_area, marketplace:)
+          expect { component.cart }.not_to change { marketplace.carts.count }
+        end
+      end
+    end
+  end
 end


app/furniture/marketplace/marketplace_component.rb
```diff
@@ -14,7 +14,7 @@ def render?
 
     def cart
       delivery_area = marketplace.delivery_areas.first if marketplace.delivery_areas.size == 1
-      @cart ||= carts.create_with(contact_email: shopper.person&.email).find_or_create_by(delivery_area: delivery_area, shopper: shopper, status: :pre_checkout)
+      @cart ||= carts.create_with(contact_email: shopper.person&.email, delivery_area: delivery_area).find_or_create_by(shopper: shopper, status: :pre_checkout)
```
rosschapman, line 17: Ah yeah, that's subtle.

rosschapman (approved):
Great tests 🧑‍🍳💋

zspencer:
Yea, the subtlety is part of why I was tempted to pull this logic out into `Marketplace::ShopperCartFactory` but... decided not to just yet.

rosschapman:
> Yea, the subtlety is part of why I was tempted to pull this logic out into `Marketplace::ShopperCartFactory` but... decided not to just yet.

Why hold back? Not sufficiently complex enough yet?

app/furniture/marketplace/marketplace_component.rb
anaulin, line 17: I think we should only have 1 cart per (marketplace, shopper) pair, no matter what is the chosen delivery area, so the delivery area shouldn't be mentioned here at all.

From a user perspective, I don't expect an online store to keep for me two separate carts depending on where my order is getting delivered to. I expect to have 1 cart, and then I can change the delivery area (and delivery address) before checking out as needed.

zspencer: Agreed; that's why there was the bug, it was trying to create a cart each time; rather than including the default `delivery_area` in cases where there is only 1 for the marketplace.

Now it only has 1 cart per shopper per marketplace; since the `find_or_create_by` no longer references the `delivery_area`.

zspencer:
> > Yea, the subtlety is part of why I was tempted to pull this logic out into `Marketplace::ShopperCartFactory` but... decided not to just yet.
> 
> Why hold back? Not sufficiently complex enough yet?

I'm not sure it's the right design. I would *rather* we didn't have to know about `session` and the `marketplace` was able to work off of just the `current_person` regardless of whether it's a `Guest` or a `Neighbor`

app/furniture/marketplace/marketplace_component.rb
anaulin: Ah, yes, d'oh, I was reading that the other way around.

zspencer: I can imagine splitting the `create_with` and `find_or_create_by` onto different lines would help 🙈

rosschapman: I think I hate `find_or_create_by`. This is a huge conflation of operations. 😅

anaulin: It is extra-fun if you chain the find-or-create with a create-with! 😹

zspencer: If we wanted to stop using both of those, we would need to write something like this:

```rb
cart = carts.find_by(shopper:, status: :pre_checkout) || carts.create(shopper: shopper, status: :pre_checkout)
cart.update(contact_email: shopper.person&.email,  delivery_area:) if cart.previously_new_record?
```

Which I don't find much more legible than:

```rb
cart = carts.create_with(contact_email: shopper.person&.email,  delivery_area:)
            .find_or_create_by(shopper:, status: :pre_checkout)
```

I agree mashing them both on one line is a bit of a disaster; but both `find_or_create_by` and `create_with` are useful tools.

rosschapman: 🏌️ Is it un-chill to do something like this in ruby/Rails? Is there supposed to be a perf gain or something with those convenience query methods? (Is this statement below materially different?)

```ruby
cart = carts.find_by(shopper:, status: :pre_checkout)
if cart.blank?
  cart = carts.create(
    shopper: shopper, 
    status: :pre_checkout, 
    contact_email: shopper.person&.email, 
    delivery_area:
)
```

zspencer: I think that also works tho the re-assignment in an if block feels odd to me but not in a horrible way.

anaulin: I personally find Ross's snippet easier to read and understand. Here is a riff on that, but without the re-assignment:
```ruby
cart = carts.find_by(shopper:, status: :pre_checkout) || carts.create(
  shopper: shopper, 
  status: :pre_checkout, 
  contact_email: shopper.person&.email, 
  delivery_area:
)
```

anaulin: And to wrap it all up, this could be moved into a factory/finder method on `Marketplace`, something like `Marketplace.find_or_create_cart(shopper:, delivery_area: nil)`.

zspencer: 🚢 :it:

anaulin: https://github.com/zinc-collective/convene/pull/2009/

Issue #1980, opened 2023-11-30
Notes from heuristic analysis of Convene Marketplace (11/24/2023)


anaulin:
Notes from a heuristic analysis session conducted by Edu and Elio of https://sutty.nl/ for the Convene Marketplace.

### Notes from session
* Landing page (https://marketplace.piikup.com/)
  * there isn't much context here -- what is this Marketplace?
  * "Home" button is mixed in with the list of vendors which is confusing -- is it a vendor that is named "Home"?
    - [x] https://github.com/zinc-collective/convene/pull/1964
    - [x] https://github.com/zinc-collective/convene/pull/1989
  * looks for a search functionality, but it doesn't exist
  * too much reliance on text in buttons -- humans are better as scanning for visuals
* Vendor page (https://marketplace.piikup.com/rooms/cafe-gabriela) 
  * there is no context for the vendor
    - [x] https://github.com/zinc-collective/convene/issues/2228
    - [x] https://github.com/zinc-collective/convene/issues/1961
  * "what does GF, NF mean???" 
    - [ ] https://github.com/zinc-collective/convene/issues/2189
  * "where are you ordering from?" section
    * small gray text is not accessible -- too small, too low-contrast 
    * it could try using the IP location for the user
  * product amounts aren't clear "feeds... how many?" 
    - [x] https://github.com/zinc-collective/convene/issues/2170
  * delivery area "change" link is too far from what it is changing, relationship isn't clear 
  * i want to make the product images bigger, and i want more images
    - [x] https://github.com/zinc-collective/convene/pull/2072
  * when navigating to another vendor, thought that the cart would be shared across vendors, but it wasn't
  * bug: somehow Cafe Gabriela cart wasn't saved and got cleared out 
    - [x] (https://github.com/zinc-collective/convene/issues/1962)
  * going to checkout: was expecting that it would show me my cart again, before asking me to enter payment details, so that i can review and confirm
    - [x] https://github.com/zinc-collective/convene/pull/1997
  * bug: checkout details form is broken 
    - [x] https://github.com/zinc-collective/convene/pull/1979
    - [x] https://github.com/zinc-collective/convene/issues/1983
  * could offer to save checkout information after checkout
* As the marketplace operator
  * there should be a clear indication that you are "in marketplace management mode" 
  * options under marketplace management should be grouped into "configuration" and "operations"
  * missing title/header when editing Delivery Areas, and some other marketplace management sections
  * there is no way to delete a Tax Rate
[...]

PR #1997, opened 2023-12-05
`Marketplace`: `Checkout` includes `Products`, `Taxes`, and `DeliveryFee`

zspencer:
- https://github.com/zinc-collective/convene/issues/1980

It's not the prettiest, but it includes each product being purchased,
the quantity, the unit pricer, and the total price as well as the taxes
and delivery fees.

Screenshot 2023-12-04 at 6 25 43 PM
Screenshot 2023-12-04 at 6 25 39 PM

### DIFF
--- a/app/furniture/marketplace/cart/deliveries/_delivery.html.erb
+++ b/app/furniture/marketplace/cart/deliveries/_delivery.html.erb
@@ -1,12 +1,12 @@
-<%- if delivery.delivery_area.present? %>
-  <p class="text-lg font-bold"><%= render(delivery.delivery_area) %></p>
-<%- end %>
-<p class="text-left py-1 text-sm">
-  <%= render(Marketplace::Cart::DeliveryExpectationsComponent.new(cart: delivery.cart)) %>
-</p>
+
 <%= turbo_frame_tag(delivery) do %>
   <% if delivery.details_filled_in? %>
-    <p class="font-bold">Delivering to:</p>
+    <h3 class="font-bold">Delivering to:</h3>
+    <p>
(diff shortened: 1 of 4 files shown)


anaulin (approved):
Yay, this is better!

app/furniture/marketplace/cart/deliveries/_delivery.html.erb
```diff
@@ -1,13 +1,13 @@
...
-</p>
+
 <%= turbo_frame_tag(delivery) do %>
   <% if delivery.details_filled_in? %>
-    <p class="font-bold">Delivering to:</p>
-    <div class="rounded p-2 bg-orange-50">
+    <h3 class="font-bold">Delivering to:</h3>
+    <p>
+      <%= render(delivery.delivery_area) %><br />
+      <span class="text-sm font-light italic"><%= render(Marketplace::Cart::DeliveryExpectationsComponent.new(cart: delivery.cart)) %></span>
```
rosschapman, line 7: Could this text styling become part of the component? Layout styling makes sense for wrapper tags but if these seem like inherent styling attributes of the component itself not of this "area" of the page.

zspencer: Sure, I think pushing this responsibility down is a good idea

zspencer: Ehh, actually let's not do that just yet. `DeliveryExpectationsComponent` is rendered in a number of places, with a substantially different information hierarchy. 

Presuming it should always be small, light, and italic does not seem to be valid.

PR #1998, opened 2023-12-05
🧹 Remove Hotjar script, extract `<head>` section of layout into shared partial

anaulin:
This removes the Hotjar JS, since we don't seem to be getting much value from it, we don't generally enjoy trackers, and anyway this was meant as an experiment.

Also took the chance to extract the repeated `<head>` tag content from our two layouts into a shared partial.

This change might or might not be a fix for the issue reported in https://github.com/zinc-collective/convene/issues/1986#issuecomment-1839912573


zspencer (approved):
purge the hotjaritic!

PR #1999, opened 2023-12-05
🌸 `Marketplace`: `Checkout` button is not shown until you can Checkout

zspencer:
- https://github.com/zinc-collective/convene/issues/1983#issuecomment-1839919010

It feels odd that we have a `Checkout` button when you can't click it.

https://github.com/zinc-collective/convene/assets/50284/700694ad-6b78-4be7-98fe-81c66ea577f0

### DIFF
--- a/app/furniture/marketplace/carts/_footer.html.erb
+++ b/app/furniture/marketplace/carts/_footer.html.erb
@@ -10,9 +10,11 @@
       <%= render "marketplace/carts/total", cart: cart %>
     </td>
   </tr>
-  <tr>
-    <td class="text-right py-3.5" colspan="8">
-      <%= render ButtonComponent.new(label: "Checkout", href: cart.location(child: :checkout), scheme: :primary, method: :get, disabled: !cart.ready_for_checkout?) %>
-    </td>
-  </tr>
+  <%- if cart.ready_for_checkout? %>
+    <tr>
+      <td class="text-right py-3.5" colspan="8">
+        <%= render ButtonComponent.new(label: "Checkout", href: cart.location(child: :checkout), scheme: :primary, method: :get) %>
+      </td>
+    </tr>
+  <%- end %>
 </tfoot>


anaulin:
What is the experience when you have products but did not select a delivery area?

I'm concerned that a button that appears/disappears without explanation is no less confusing than having the button look enabled but then not work. We should at least have some way to communicate to the shopper that in order to checkout they need to (1) select a delivery area, and (2) add products in order to be able to initiate checkout.

Ideally the button would become grayed out or some other visual indication that it is "not available yet", with an explanation of why ("add products to initiate checkout", or "select delivery area to initiate checkout"). (Unrelated to this PR, maybe the selection of delivery area should happen after you initiate checkout anyway, together with entering the delivery address.)

anaulin (approved):
I appreciate the quick iteration on this, thank you!

I am all for merging this PR and then iterating to make the button a little less Dave Copperfield-y.

zspencer:
> What is the experience when you have products but did not select a delivery area?

Right now there's *another* bug where updating the `DeliveryArea` doesn't update the `Cart`; which basically means that the `Checkout` button doesn't show up until they change their `Cart` in other ways. It's annoying; but is going to exist if we disable / enable or hide / show.

My hope is to get a bit further on splitting the `Cart` interface from the `Menu` interface; which should both eliminate the error and actually make things better.

> I'm concerned that a button that appears/disappears without explanation is no less confusing than having the button look enabled but then not work. We should at least have some way to communicate to the shopper that in order to checkout they need to (1) select a delivery area, and (2) add products in order to be able to initiate checkout.
>
> Ideally the button would become grayed out or some other visual indication that it is "not available yet", with an explanation of why ("add products to initiate checkout", or "select delivery area to initiate checkout"). 

Agreed. I'm tilting at that next.

>  (Unrelated to this PR, maybe the selection of delivery area should happen after you initiate checkout anyway, together with entering the delivery address.)

I think it is probably better to have them pick their Delivery Area prior to showing the `Menu` since the delivery schedule is tied to the area; and picking an entire catering order only to find out you can only get stuff delivered on Friday probably won't feel great.

PR #2008, opened 2023-12-07
Add .env.test.example with env variables that are needed for system tests

anaulin:
I _think_ these are all the env variables that are needed for the new additions to our system tests to work.

Sadly, I can't be 100% certain because spending over 30 mins on setting this up, I still haven't managed to make `./spec/furniture/marketplace/buying_products_system_spec.rb` pass on my machine (I get some weird exceptions about the stripe_cli terminating with exceptions, no indication of what the exception is though).


zspencer (approved):
Oh good call! I have just been  shoveling them in my .env 🙈

PR #2009, opened 2023-12-07
🧹 `Marketplace`: Move logic for finding or creating a cart for a shopper into a factory method

anaulin:
Following up to the convo in https://github.com/zinc-collective/convene/pull/1992#discussion_r1418188891, this PR tightens up the logic for finding or creating a cart for a shopper.

### DIFF
--- a/app/furniture/marketplace/marketplace.rb
+++ b/app/furniture/marketplace/marketplace.rb
@@ -26,6 +26,15 @@ class Marketplace
     # Square order notifications integration
     setting :square_location_id
 
+    def find_or_create_cart(shopper:, delivery_area: nil, cart_status: :pre_checkout)
+      carts.find_by(shopper:, status: cart_status) || carts.create(
+        shopper: shopper,
+        contact_email: shopper.person&.email,
+        delivery_area: delivery_area,
+        status: cart_status
+      )
+    end
+
     def square_access_token=square_access_token
       secrets["square_access_token"] = square_access_token
     end
--- a/app/furniture/marketplace/marketplace.rb
+++ b/app/furniture/marketplace/marketplace.rb
@@ -26,7 +26,7 @@ class Marketplace
     # Square order notifications integration
     setting :square_location_id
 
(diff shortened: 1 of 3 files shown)


app/furniture/marketplace/marketplace.rb
```diff
@@ -26,6 +26,15 @@ class Marketplace < Furniture
     # Square order notifications integration
     setting :square_location_id
 
+    def find_or_create_cart(shopper:, delivery_area: nil, cart_status: :pre_checkout)
```
anaulin, line 29: Now that I've implemented this, I'm wondering if this version still has the confusion issues that @rosschapman was originally pointing about the conflation of "finding" and "creating" in one call.

Thoughts?

app/furniture/marketplace/marketplace_component.rb
```diff
@@ -14,7 +14,7 @@ def render?
 
     def cart
       delivery_area = marketplace.delivery_areas.first if marketplace.delivery_areas.size == 1
-      @cart ||= carts.create_with(contact_email: shopper.person&.email, delivery_area: delivery_area).find_or_create_by(shopper: shopper, status: :pre_checkout)
+      @cart ||= marketplace.find_or_create_cart(shopper: shopper, delivery_area: delivery_area)
```
zspencer, line 17: This kind of reads as if the cart is being created for the `delivery_area`; rather than the default delivery area being set in the event a cart is created.

Perhaps we put line 16 in the `marketplace.find_or_create_cart(shopper: shopper)` implementation? Otherwise I think this is a bit better?

zspencer (approved):

rosschapman:
I think the confusion remains because the naming is too much _how_ vs _what_ for an abstraction. What if the method was called `new_shopper_cart` or something similar that better describes the value within context of it's usage.

rosschapman (approved):

anaulin:
@rosschapman would something like `cart_for_shopper` be a better name for the method? (`new_cart` wouldn't be right, since it is not necessarily a new cart).

PR #2010, opened 2023-12-07
🧹  Fix Rails 7.1 deprecation warning: `fixture_paths` are now plural, and an array

anaulin:
Resolves this deprecation warning when running tests:
image


zspencer (approved):
YESSSSSS

PR #2012, opened 2023-12-08
Adds test assertions for square order events, and docstring for order events

rosschapman:


app/furniture/marketplace/square_order.rb
```diff
@@ -31,17 +31,13 @@ def initialize(square_client)
     def send_order(order, stripe_charge_id)
       @order = order
       @stripe_charge_id = stripe_charge_id
-      Rails.logger.info("Start creating Square order")
```
rosschapman, line 34: I don't remember why I wanted these logs in here. 🤷

anaulin (approved):
Good catches and cleanups!

PR #2015, opened 2023-12-10
🐞🔨 `Marketplace`: `DeliveryAreas` cannot be Removed when in use

zspencer:
- https://github.com/zinc-collective/convene/issues/2013

This:

- Removes the affordance to `destroy` `DeliveryAreas` with an `Order`
- Un-sets `Cart#delivery_area` on `DeliveryAreas#destroy`

### DIFF
--- a/app/furniture/marketplace/delivery_area.rb
+++ b/app/furniture/marketplace/delivery_area.rb
@@ -4,8 +4,8 @@ class Marketplace
     location(parent: :marketplace)
 
     belongs_to :marketplace, inverse_of: :delivery_areas
-    has_many :orders, inverse_of: :delivery_area
-    has_many :carts, inverse_of: :delivery_area
+    has_many :orders, -> { checked_out }, inverse_of: :delivery_area
+    has_many :carts, inverse_of: :delivery_area, dependent: :nullify
     has_many :deliveries, inverse_of: :delivery_area
 
     attribute :delivery_window
--- a/app/furniture/marketplace/delivery_areas_controller.rb
+++ b/app/furniture/marketplace/delivery_areas_controller.rb
@@ -32,11 +32,7 @@ class Marketplace
     end
 
     def destroy
-      authorize(delivery_area)
-      # rubocop:disable Rails/SkipsModelValidations
-      delivery_area.carts.update_all(delivery_area_id: nil)
-      # rubocop:enable Rails/SkipsModelValidations
-      delivery_area.destroy
+      authorize(delivery_area).destroy
 
       respond_to do |format|
         format.turbo_stream do
--- a/app/furniture/marketplace/delivery_areas_controller.rb
+++ b/app/furniture/marketplace/delivery_areas_controller.rb
@@ -32,7 +32,11 @@ class Marketplace
     end
 
     def destroy
-      authorize(delivery_area).destroy
+      authorize(delivery_area)
+      # rubocop:disable Rails/SkipsModelValidations
+      delivery_area.carts.update_all(delivery_area_id: nil)
+      # rubocop:enable Rails/SkipsModelValidations
+      delivery_area.destroy
 
       respond_to do |format|
         format.turbo_stream do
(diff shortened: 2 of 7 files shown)


app/furniture/marketplace/delivery_areas_controller.rb
```diff
@@ -32,7 +32,11 @@ def update
     end
 
     def destroy
-      authorize(delivery_area).destroy
+      authorize(delivery_area)
+      # rubocop:disable Rails/SkipsModelValidations
+      delivery_area.carts.update_all(delivery_area_id: nil)
+      # rubocop:enable Rails/SkipsModelValidations
+      delivery_area.destroy
```
anaulin, line 39: Might be worth wrapping these two lines in a transaction.

app/furniture/marketplace/stripe_events_controller.rb
```diff
@@ -13,7 +13,7 @@ def create
 
       when "checkout.session.completed"
         payment_intent = Stripe::PaymentIntent.retrieve(event.data.object.payment_intent, {api_key: marketplace.stripe_api_key})
-        order = marketplace.orders.find_by(id: payment_intent.transfer_group)
+        order = Order.unscoped.find_by(id: payment_intent.transfer_group, marketplace:)
```
anaulin, line 16: I don't love how this reads. It immediately makes me want to go and check why does this need to be unscoped and what does that mean. Does that mean that really this is a "cart_or_order"? Or should this be a "cart"? Or something else?

app/furniture/marketplace/order.rb
```diff
@@ -1,6 +1,7 @@
 class Marketplace
   class Order < Record
     self.table_name = "marketplace_orders"
+    default_scope { where.not(status: :pre_checkout) }
```
anaulin, line 4: Why do we need a default scope here? How certain are we that this won't introduce yet another bug somewhere where an edge case isn't tested? (I'm not certain at all, personally.)

I'd rather we explicitly applied named scopes as needed, instead of having a default scope.

zspencer: Yea, this is a side effect of not having split the `marketplace_carts` and `marketplace_orders` tables out. I like putting it in a named scope!

app/furniture/marketplace/stripe_events_controller.rb
zspencer: This is a consequence of that default scope that I added; I'm going to revert that.

app/furniture/marketplace/delivery_areas_controller.rb
zspencer: ooo good point.

anaulin: Random thought: could we accomplish this with something like this on the Delivery Area model?

```ruby
  has_many :pre_checkout_carts, ...appropriate scopes go here..., dependent: :nullify
```

app/furniture/marketplace/delivery_area.rb
```diff
@@ -4,8 +4,8 @@ class DeliveryArea < Record
     location(parent: :marketplace)
 
     belongs_to :marketplace, inverse_of: :delivery_areas
-    has_many :orders, inverse_of: :delivery_area
-    has_many :carts, inverse_of: :delivery_area
+    has_many :orders, -> { checked_out }, inverse_of: :delivery_area
+    has_many :carts, inverse_of: :delivery_area, dependent: :nullify
```
anaulin, line 8: Ah, and this works because `Marketplace::Carts` has a default scope of `:pre_checkout`, right? Nice

anaulin (approved):
Nice, ship it!

app/furniture/marketplace/delivery_area.rb
zspencer: Yes, which is part of why I kinda-like the default scopes; but also I now have a ticket for the refactor: https://github.com/zinc-collective/convene/issues/2020

PR #2021, opened 2023-12-11
🐞🔨`Section`: `Sections` with `Gizmos` may not be destroyed

zspencer:
- https://github.com/zinc-collective/convene/issues/2019

There's probably a smarter thing to do here; but it will have to wait til we turn on Rails polymorphism for Gizmos.

As it stands, trying to destroy Gizmos when they aren't instantiated as their actual Type results in a lot of foreign key violations.


anaulin (approved):
Thanks for the quick fix!

Issue #2014, opened 2023-12-10
`Distributor` archives `DeliveryArea`


zspencer:

anaulin:
FWIW, I've not hated using this gem for "archiving" or other soft-deleting: https://github.com/jhawthorn/discard

PR #2024, opened 2023-12-12
🥔✨ `Marketplace`: `DeliveryAreas` may be `Archived`

zspencer:
- https://github.com/zinc-collective/convene/issues/2014

This adds the `discard` gem so that we can archive objects. It also updates the workflow for removing DeliveryArea to have an `Archive` step first.

Once a `DeliveryArea` is archived, it can not yet be restored; but it can be removed if there are no `Orders` associated with the `DeliveryArea`

### DIFF
--- a/app/furniture/marketplace/delivery_area.rb
+++ b/app/furniture/marketplace/delivery_area.rb
@@ -16,5 +16,13 @@ class Marketplace
 
     attribute :delivery_window
     monetize :price_cents
+
+    def discardable?
+      persisted? && kept?
+    end
+
+    def destroyable?
+      persisted? && discarded? && orders.empty?
+    end
   end
 end
--- a/app/furniture/marketplace/delivery_area.rb
+++ b/app/furniture/marketplace/delivery_area.rb
@@ -1,11 +1,19 @@
 class Marketplace
   class DeliveryArea < Record
+    include ::Discard::Model
+
     self.table_name = "marketplace_delivery_areas"
     location(parent: :marketplace)
 
     belongs_to :marketplace, inverse_of: :delivery_areas
     has_many :orders, -> { checked_out }, inverse_of: :delivery_area
     has_many :carts, inverse_of: :delivery_area, dependent: :nullify
+    after_discard do
+      # rubocop:disable Rails/SkipsModelValidations
+      carts.update_all(delivery_area_id: nil)
+      # rubocop:enable Rails/SkipsModelValidations
+    end
+
     has_many :deliveries, inverse_of: :delivery_area
 
     attribute :delivery_window
--- a/app/furniture/marketplace/delivery_area.rb
+++ b/app/furniture/marketplace/delivery_area.rb
@@ -9,9 +9,7 @@ class Marketplace
     has_many :orders, -> { checked_out }, inverse_of: :delivery_area
     has_many :carts, inverse_of: :delivery_area, dependent: :nullify
     after_discard do
-      # rubocop:disable Rails/SkipsModelValidations
-      carts.update_all(delivery_area_id: nil)
-      # rubocop:enable Rails/SkipsModelValidations
+      carts.update_all(delivery_area_id: nil) # rubocop:disable Rails/SkipsModelValidations
     end
 
     has_many :deliveries, inverse_of: :delivery_area
--- a/app/furniture/marketplace/delivery_area_component.html.erb
+++ b/app/furniture/marketplace/delivery_area_component.html.erb
@@ -2,6 +2,9 @@
 <%= render CardComponent.new(dom_id: dom_id(delivery_area), classes: "flex flex-col justify-between gap-y-2 w-full") do %>
   <header class="flex font-bold">
     <%= label %>
+    <%- if delivery_area.discarded?%>
+      <span class="italic">(archived)</span>
+    <%- end %>
   </header>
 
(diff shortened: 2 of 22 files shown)


zspencer:
I decided to merge my PRs that were building on this one.

app/furniture/marketplace/delivery_area.rb
```diff
@@ -1,11 +1,19 @@
...
     self.table_name = "marketplace_delivery_areas"
     location(parent: :marketplace)
 
     belongs_to :marketplace, inverse_of: :delivery_areas
     has_many :orders, -> { checked_out }, inverse_of: :delivery_area
     has_many :carts, inverse_of: :delivery_area, dependent: :nullify
+    after_discard do
+      # rubocop:disable Rails/SkipsModelValidations
+      carts.update_all(delivery_area_id: nil)
+      # rubocop:enable Rails/SkipsModelValidations
```
anaulin, line 14:
```suggestion
      carts.update_all(delivery_area_id: nil) # rubocop:disable Rails/SkipsModelValidations
```

app/furniture/marketplace/delivery_area_component.html.erb
```diff
@@ -2,6 +2,9 @@
 <%= render CardComponent.new(dom_id: dom_id(delivery_area), classes: "flex flex-col justify-between gap-y-2 w-full") do %>
   <header class="flex font-bold">
     <%= label %>
+    <%- if delivery_area.discarded?%>
```
anaulin, line 5:
```suggestion
    <%- if delivery_area.discarded? %>
```

app/furniture/marketplace/delivery_area_component.rb
```diff
@@ -28,20 +28,30 @@ def edit_button?
...
+    def discard_button
+      return unless discard_button?
+
+      ButtonComponent.new(label: "#{t("icons.discard")} #{t("discard.link_to")}",
+        title: t("marketplace.delivery_areas.discard.link_to", name: delivery_area.label),
+        href: delivery_area.location, method: :delete, scheme: :secondary)
+    end
+
+    def discard_button?
+      delivery_area.persisted? && !delivery_area.discarded? && policy(delivery_area).destroy?
```
anaulin, line 40: `!discarded?` is the same as `kept?` and also the same as `undiscarded?`, I would slightly prefer:
```suggestion
      delivery_area.persisted? && delivery_area.kept? && policy(delivery_area).destroy?
```

app/furniture/marketplace/delivery_area_component.rb
```diff
@@ -28,20 +28,30 @@ def edit_button?
...
 
+    def discard_button
+      return unless discard_button?
+
+      ButtonComponent.new(label: "#{t("icons.discard")} #{t("discard.link_to")}",
+        title: t("marketplace.delivery_areas.discard.link_to", name: delivery_area.label),
+        href: delivery_area.location, method: :delete, scheme: :secondary)
+    end
+
+    def discard_button?
```
anaulin, line 39: Minor: it feels like this method and the `destroy_button?` methods want to instead to be methods on Delivery Area, `DeliveryArea#discardable?` and `DeliveryArea#destroyable?`. Then those can be unit-tested in the model spec, and this component calls those methods to determine if the buttons can/should be displayed.

It also helps the model self-document its behavior in terms of when it can be discarded or destroyed.

app/furniture/marketplace/delivery_areas/index.html.erb
```diff
@@ -19,5 +23,12 @@
           scheme: :secondary) %>
       <%- end %>
     </div>
+    <div class="text-center mt-3">
+      <%- if params[:archive] %>
+        <%= link_to "Active", marketplace.location(child: :delivery_areas) %>
+      <%- else %>
+        <%= link_to "Archive", marketplace.location(child: :delivery_areas, query_params: { archive: true }) %>
```
anaulin, line 30: Maybe the links want to be say "Active areas" and "Archived areas" (or "Active delivery areas", if there is enough space), otherwise I'm not sure it's clear what they are for.

anaulin (approved):
Nice, thank you!

Some suggestions below, but nothing blocking.

app/furniture/marketplace/delivery_area_component.rb
zspencer: I dig it! I've left the policy check but pulled out the state checks.

app/furniture/marketplace/delivery_areas/index.html.erb
zspencer: Nice, I've updated these too!

app/furniture/marketplace/delivery_area_component.rb
rosschapman: I was thinking the same thing @anaulin! 

Was thinking about how in React (or similar) -- which supposedly influences ViewComponent -- it wasn't encouraged to pass entities into a component but instead a _view_ over the model<sup>1</sup> with strong indirections already prepped so you wouldn't end up coupling your entity layer to your view layer so tightly. In that modality we'd have a view function like: 

```ts
class MyComponent {
	constructor(data) {
		this.model = data;
	}

	showDiscardButton() {
		return this.model.canRemoveDeliveryArea;
	}
}
```ts

<sup>1</sup> Presenter, view model, etc... but a simple data bag

PR #2030, opened 2023-12-13
🥗🧹🥔✨ `Marketplace`: Archiving `Products`

zspencer:

`Products` may be `Archived`, and (if `#unordered?`) `Removed`. Archived `Products` are not available for sale.

This also wraps the `Products#new` and `Products#edit` pages with the `Marketplace::ManagementComponent`
which is a much better experience.


app/furniture/marketplace/delivery_area.rb
```diff
@@ -22,7 +22,15 @@ def discardable?
     end
 
     def destroyable?
-      persisted? && discarded? && orders.empty?
+      persisted? && discarded? && unordered?
```
rosschapman, line 25: What about adding a new `.safe_destroy` method to the model and putting the check therein, so you wouldn't have to write `delivery_area.destroy if delivery_area.destroyable?` ? I'm not worried about the extra keystrokes so much as making it easier to avoid a mistake. Does creating a method like that contradict conventional ActiveRecord model configuration?

app/furniture/marketplace/delivery_area.rb
```diff
@@ -22,7 +22,15 @@ def discardable?
...
+    end
+
+    def ordered?
+      orders.present?
+    end
+
+    def unordered?
+      !ordered?
```
rosschapman, line 33: I feel like it will be so easy to interpret these methods for sorting, although I understand this type of grammar is popular in ruby/Rails. Because of the potential confusion could an alternative like `has_orders` work here?

app/furniture/marketplace/product.rb
```diff
@@ -14,9 +16,14 @@ class Product < Record
...
     has_many :carts, through: :cart_products, inverse_of: :products
+    after_discard do
+      # This is redonkulus, and due to the inappropriate polymorphism where
+      # carts/orders are the same table.
+      cart_products.where(cart: carts).destroy_all
+    end
 
     has_many :ordered_products, inverse_of: :product, dependent: :destroy
-    has_many :orders, through: :ordered_products, inverse_of: :products
+    has_many :orders, -> { checked_out }, through: :ordered_products, inverse_of: :products
```
rosschapman, line 26: For my edification, how does "checked out" differ from "paid"?

app/furniture/marketplace/delivery_area.rb
zspencer: I could see something like `destroy_safely` which:
- Discards if it's `discardable?`
- Destroys if it's `destroyable?`
- Does nothing if either.

app/furniture/marketplace/delivery_area.rb
zspencer: I was trippin' over that myself, what about something like `with_orders?` and `without_orders?`Or I could just keep doing `ordes.present?` since I don't think it's the end of the world and adding in the predicate method could be confusing.

app/furniture/marketplace/carts/_cart.html.erb
```diff
@@ -16,7 +16,7 @@
         </tr>
       </thead>
       <tbody class="divide-y divide-gray-200 bg-white">
-        <%- cart.marketplace.products.each do |product| %>
+        <%- cart.marketplace.products.kept.each do |product| %>
```
anaulin, line 19: Since we already ensure that when a Product is discarded it gets removed of all non-order carts, we shouldn't need to have this extra scope here, right?

When I read this with the scope it makes me think that maybe a Cart (not-order) can contain archived Products, which seems wrong.

rosschapman, line 19: Is `cart.marketplace.products` different from `cart.marketplace_products`? Is there a reason to use one or the other?

spec/furniture/marketplace/product_spec.rb
```diff
@@ -18,4 +18,50 @@
...
+
+          let(:order) { create(:marketplace_order) }
+
+          before { order.products << product }
+
+          it { is_expected.not_to be_destroyable }
+        end
+      end
+    end
+  end
```
anaulin, line 66: Might be worth having a test case that verifies that when a product that is in a pre-checkout cart is discarded, it gets removed from those carts.

app/furniture/marketplace/product.rb
zspencer: `paid` is probably fine to use instead at this time; but as we think about what states an Order can flow through (`placed`, `paid`, `preparing`, `ready_for_pickup`, `in_transit`, `delivered`, `reviewed`) 

Tho to be honest, all this goes away once I get the `Cart` and `Order` into different tables. That's just a project for Spoons Having Zee

spec/furniture/marketplace/product_spec.rb
anaulin: You have a test for that in the system spec already. 👍🏼

anaulin (approved):

app/furniture/marketplace/delivery_area.rb
anaulin: Maybe. I worry a bit about losing the explicit intent to discard vs destroy.

app/furniture/marketplace/carts/_cart.html.erb
zspencer: This is the `marketplace.products`, not the `cart` products; which is a manifestation of the gross-ness of not having a `menu` which is distinct from the `cart`

app/furniture/marketplace/delivery_area.rb
anaulin: I like `with_orders?`

app/furniture/marketplace/carts/_cart.html.erb
zspencer: There is no `cart.marketplace_products`. There is `cart.products`, which is the products associated to that particular `cart` and there is `cart.marketplace.products` which is the `products` that belong the the `marketplace` that the `cart` is part of.

spec/furniture/marketplace/product_spec.rb
zspencer: Yea, I think I want to push the discard-removal checks down into the model specs; I just wasn't sure where to put it when I started and I started with the system tests.

app/furniture/marketplace/delivery_area.rb
zspencer: You're right, it's better for `destroy_safely`  to be `destroy if destroyable?` and otherwise do nothing.

app/furniture/marketplace/carts/_cart.html.erb
rosschapman: Totally subjective commentary: for some reason `kept` feels like a weird word to me to see in code, idk why.

rosschapman: > There is no `cart.marketplace_products`. There is `cart.products`, which is the products associated to that particular `cart` and there is `cart.marketplace.products` which is the `products` that belong the the `marketplace` that the `cart` is part of.

Oh I saw `cart.cart_products` in the code elsewhere and misread it.

anaulin: Ah, yes, sorry, I misread!

app/furniture/marketplace/product.rb
rosschapman: It sounds like what you're saying is that they are the same, which makes sense since it only looks like nominal mapping to me that happens at https://github.com/zinc-collective/convene/blob/main/app/furniture/marketplace/order.rb#L4.

app/furniture/marketplace/carts/_cart.html.erb
anaulin: Agreed that `kept` is somehow... funny? I believe the discard gem also offers `undiscarded` as an alias to the `kept` predicates and scopes, so maybe we should use that instead, since it is anyway clearer.

FWIW, on a client project where we use this gem we alias the `discard`/`undiscard` methods and scopes to a name that matches semantically whatever we are doing on that particular model. So e.g. on an accounting model that represents a charge to a client, I aliased `discard` to be `waive`. We could do the same here and make scopes `archived`, `unarchived`, etc.

zspencer: I like that, I kinda maybe went overboard by renaming everything.

app/furniture/marketplace/product.rb
zspencer: Ehh, they're not exactly the same. `where.not(status: :pre_checkout)` will return records where status is `preparing`, `ready_for_pickup`, etc, whereas the `paid` scope will only return records `where(status: :paid)`.

app/furniture/marketplace/archivable.rb
```diff
@@ -0,0 +1,30 @@
...
+  module Archivable
+    def self.included(model)
+      model.include(Discard::Model)
+      model.alias_method :archive, :discard
+      model.alias_method :archived?, :discarded?
+      model.alias_method :available?, :kept?
+      model.alias_method :retrieve, :undiscard
+
+      model.scope(:archived, -> { discarded })
+      model.scope(:available, -> { kept })
```
anaulin, line 11: I like having all of this in a module a lot! Expecting models within the Marketplace to be able to answer `with_orders?` also makes a lot of sense.

My one nit is that I would prefer `not_archived` to `available`, because it seems clearer. `Available` makes me think of "product not available in this delivery area" or "product not inventory". But I won't fight you on the name at this point. 😺

zspencer: I was having similar feels re: Available; I'm down for `unarchived`!

app/furniture/marketplace/carts/_cart.html.erb
rosschapman: Aliasing sounds like a good idea since it's more domain-driven.

app/furniture/marketplace/product.rb
rosschapman: Right, what I was trying to understand is if they could be effectively considered equivalent _states_ in the current code since the `status` enum only declares `pre_checkout` and `paid`. But I get what you mean here.

app/furniture/marketplace/archivable.rb
```diff
@@ -0,0 +1,30 @@
+class Marketplace
+  module Archivable
+    def self.included(model)
+      model.include(Discard::Model)
+      model.alias_method :archive, :discard
+      model.alias_method :archived?, :discarded?
+      model.alias_method :unarchived?, :kept?
+      model.alias_method :retrieve, :undiscard
```
rosschapman, line 8: Should `retrieve` be `unarchive` now to be consistent?

zspencer: Probably, also if you have suggestions you can always make a PR implementing your suggestions!

app/furniture/marketplace/product.rb
zspencer: I would say they're equivalent _for now_ rather than that they're the same; but objective equality is hard enough in deterministic computational systems; much less ones subjected to the chaos of temporality 🤣

PR #2034, opened 2023-12-15
🥗 `Marketplace`: End-to-end test `Seller` receives `Order in `Square`  Point of Sale

rosschapman:

Simply adds `with_square` trait to the Marketplace factory which effectively enables the Square notifications workflow and will send orders to a Square sandbox account during the test.


rosschapman:
@zspencer could enabling the Square workflow in these tests be enough? Also, I think there may be something [wrong](https://github.com/zinc-collective/convene/actions/runs/7228187339/job/19697184425?pr=2034) with the Square sandbox credentials. Can you double check please?

zspencer:
I'm going to be out of commission until after Christmas; and don't have bandwidth to debug. 

If you have interest + capacity to pair with @anaulin, she should have the permissions to access both the Square account and update the settings. Otherwise, let's dig in next Wednesday!

spec/furniture/marketplace/buying_products_system_spec.rb
```diff
@@ -57,6 +57,13 @@ def url_options
   end
 
   it "Works for Guests" do # rubocop:disable RSpec/ExampleLength
+    puts "START####################################################################################################################################################################################"
+
+    puts marketplace.square_location_id
+    puts marketplace.square_environment
+    puts marketplace.square_access_token
+
+    puts "END####################################################################################################################################################################################"
```
zspencer, line 66: Let's purge these debug logs?

zspencer (approved):
Let's merge this as is and follow on a `Buying Products` without `Square` test someday later once it becomes important.

PR #2038, opened 2023-12-17
🥗🌸✨ `Marketplace`: Hide `archived` `Products` from `Products#index`

zspencer:
- https://github.com/zinc-collective/convene/issues/2023

This makes `Products` more consistent with `DeliveryAreas` by:
- Removing the `Remove` link from the `ProductComponent` unless the `Product` is already `archived?` 
- Adding a visual indicator that `Products` have been `archived` to the `ProductComponent`
- Moving the `archived` `Products` to another page.

## Nerd Notes:
 
 I'm not convinced this Component is the right component to extract. Maybe a `DeliveryArea::ManagementComponent` and `Product::ManagementComponent` are the right step instead?

### DIFF
--- a/app/furniture/marketplace/delivery_areas/index.html.erb
+++ b/app/furniture/marketplace/delivery_areas/index.html.erb
@@ -24,11 +24,7 @@
       <%- end %>
     </div>
     <div class="text-center mt-3">
-      <%- if params[:archive] %>
-        <%= link_to "Active Delivery Areas", marketplace.location(child: :delivery_areas) %>
-      <%- else %>
-        <%= link_to "Archived Delivery Areas", marketplace.location(child: :delivery_areas, query_params: { archive: true }) %>
-      <%- end %>
+      <%= render Marketplace::Archivable::IndexLinkComponent.new(marketplace: marketplace, resource: :products) %>
     </div>
   </section>
 <% end %>
--- a/app/furniture/marketplace/delivery_areas/index.html.erb
+++ b/app/furniture/marketplace/delivery_areas/index.html.erb
@@ -24,7 +24,7 @@
       <%- end %>
     </div>
     <div class="text-center mt-3">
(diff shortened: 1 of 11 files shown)


app/furniture/marketplace/delivery_areas/index.html.erb
```diff
@@ -24,11 +24,7 @@
       <%- end %>
     </div>
     <div class="text-center mt-3">
-      <%- if params[:archive] %>
-        <%= link_to "Active Delivery Areas", marketplace.location(child: :delivery_areas) %>
-      <%- else %>
-        <%= link_to "Archived Delivery Areas", marketplace.location(child: :delivery_areas, query_params: { archive: true }) %>
-      <%- end %>
+      <%= render Marketplace::Archivable::IndexLinkComponent.new(marketplace: marketplace, resource: :delivery_areas, to_archive: !params[:archive]) %>
```
rosschapman, line 27: Will more logic go into this component? It seems far too granular to me because this component is essentially a one-liner if you cast the param to a bool and just pass it in, along with the new translations:

```ruby
<%= link_to 
	t("delivery_areas.#{!params[:archive] ? :archive : :active}"), 
	marketplace.location(
		child: :delivery_areas, 
		query_params: { 
			archive: !params[:archive] 
		}
	) 
%>
```
Unless rails will do something weird with the empty value in the `query_parms` dictionary? I think your gut is right that the abstraction is at a higher level.

rosschapman (approved):

app/furniture/marketplace/delivery_areas/index.html.erb
zspencer: Yea, that's part of why I think a Smerter :tm:  Abstraction would be a `Marketplace::Arhcivable::ManagementComponent` or something.

Initially I componentized it to get some tests around it; since I wanted to make sure I was looking up the right translation in the right context.

PR #2053, opened 2023-12-21
🌸 `Marketplace`: `NotificationMethods` is framed by ManagementComponent

zspencer:
- https://github.com/zinc-collective/convene/issues/1331

I noticed this while setting up [Zee's
Pizza](https://marketplace-sandbox.zinc.coop/rooms/zee-s-pizza)

Now it's prurty

PR #2054, opened 2023-12-21
🌸 `Marketplace`: All `Member` pages are in the `ManagementComponent`

zspencer:

- While working on https://github.com/zinc-collective/convene/pull/2053, I thought to myself "Wait, what if... What if we didn't have to remember to wrap things in the `ManagementComponent` at all?! What if all that could be eliminated from the `new`/`edit`/`index`/`show` pages entirely?

  So I did! Hooray!


anaulin (approved):
Oh, you beautiful genius! 😻

PR #2055, opened 2023-12-21
✨ `Section`: Allow `description` to be set for SEO

zspencer:

- Meta description tags are useful for SEO; and maybe even useful for https://github.com/zinc-collective/convene/issues/1988

- I would love if we allowed `Section#title` to be set as well

### DIFF
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -28,7 +28,7 @@ class Room < ApplicationRecord
   has_many :gizmos, dependent: :destroy, inverse_of: :room, class_name: :Furniture
   accepts_nested_attributes_for :gizmos
 
-  validates :description, length: {maximum: 300}
+  validates :description, length: {maximum: 300, allow_blank: true}
 
   def full_slug
     "#{space.slug}--#{slug}"
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -28,6 +28,8 @@ class Room < ApplicationRecord
   has_many :gizmos, dependent: :destroy, inverse_of: :room, class_name: :Furniture
   accepts_nested_attributes_for :gizmos
 
+  validates :description, length: {maximum: 160}
+
   def full_slug
     "#{space.slug}--#{slug}"
   end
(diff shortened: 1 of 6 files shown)


anaulin:
CI is sad about this PR, for what appears to be legitimate reasons. For example:
image

db/schema.rb
```diff
@@ -30,6 +30,12 @@
     "revoked",
   ], force: :cascade
 
+  create_enum :slipvector_survey_status, [
+    "preparing",
+    "in_progress",
+    "complete",
+  ], force: :cascade
+
```
anaulin, line 38: This seems like a case of mistaken schema change. `slipvector_survey_status` isn't related to this PR, and I don't see the schema change to add the `description` column to `rooms`.

app/models/room.rb
```diff
@@ -28,6 +28,8 @@ class Room < ApplicationRecord
   has_many :gizmos, dependent: :destroy, inverse_of: :room, class_name: :Furniture
   accepts_nested_attributes_for :gizmos
 
+  validates :description, length: {maximum: 160}
```
anaulin, line 31: Some quick searches suggest that a reasonable estimate of the average character length for an English sentence is about 100, so I think we should allow at least 200 characters here, or even 300, so that the description can be a couple of sentences.

anaulin (approved):
This looks good, thank you!

You'll need to fix up the schema changes in this PR before merging.

db/schema.rb
zspencer: Womp womp. Sorry, I think I added the wrong bits. Will fix.

app/models/room.rb
zspencer: Sadly, [Google limits to 160 characters](https://moz.com/learn/seo/meta-description), but I agree a longer limit is probably better.

db/schema.rb
zspencer: My bad! I have a long-running-secret-branch that I never rolled back the migrations on.

PR #2056, opened 2023-12-21
Add People with well-known emails, and a space with sections and gizmos to our `db/seeds.rb`

anaulin:
Now running `rails db:seed` will populate the local database with an operator, and a space with members, sections, and a few gizmos.

To sign in locally as the operator, use `operator@example.com`.
To sign in locally as a Space member, use `member@example.com`.
Or check the constants at the top of `seeds.rb` if you forget. 🧠

### DIFF
--- a/bin/setup-rails
+++ b/bin/setup-rails
@@ -23,12 +23,9 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Copying sample files =="
   system!('bin/copy-example-envs')
 
-  puts "\n== Preparing development database =="
-  system! 'bin/rails db:prepare'
-
-  puts "\n== Seeding development database according to feature settings in .env =="
-  system! 'bin/rake db:seed'
-  system! 'bin/rake release:after_build'
+  puts "\n== Creating and seeding development database =="
+  system! 'bin/rails db:setup'
+  system! 'bin/rails release:after_build'
 
   puts "\n== Preparing test database =="
   system! 'bin/rails db:test:prepare'
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,9 +1,44 @@
 # frozen_string_literal: true
 
-# This file should contain all the record creation needed to seed the database with its default values.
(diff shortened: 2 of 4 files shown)


zspencer (approved):
Thisis nice, thank you

bin/setup-rails
```diff
@@ -23,12 +23,9 @@ FileUtils.chdir APP_ROOT do
...
   system!('bin/copy-example-envs')
 
-  puts "\n== Preparing development database =="
-  system! 'bin/rails db:prepare'
-
-  puts "\n== Seeding development database according to feature settings in .env =="
-  system! 'bin/rake db:seed'
-  system! 'bin/rake release:after_build'
+  puts "\n== Creating and seeding development database =="
+  system! 'bin/rails db:setup'
```
zspencer, line 27: I think `prepare` is the new preference because it will setup *or* migrate; whereas setup presumes a new database.

db/seeds.rb
```diff
@@ -1,9 +1,44 @@
...
+  :with_notification_methods,
+  :with_orders,
+  :with_tax_rates,
+  room: marketplace_section
+)
+
+journal_section = FactoryBot.create(:room, space:, name: "Jazzy Journal")
+journal = FactoryBot.create(:journal, room: journal_section)
+FactoryBot.create_list(:journal_entry, 3, journal:)
+FactoryBot.create_list(:journal_keyword, 2, journal:)
```
zspencer, line 44: I wonder if this would be nicer as `create_list(:journal_entry, 3, :with_keywords, journal:)` so the entries themselves will have keywords in them rather than orphaned keywords sans entries.

bin/setup-rails
anaulin: Ah, I didn't realize was nicely idempotent like that! Done.

db/seeds.rb
anaulin: Added a `:with_keywords` trait and I'm using it here!

PR #2057, opened 2023-12-21
🥔✨ `SectionNavigation`: Include `Section#description` in SectionLink

zspencer:
 
 ## TODO
- [x] Make Pretty
- [x] Fix tests?
- [x] Split the `CardComponent` changes out?

## After
Screenshot 2023-12-26 at 8 39 41 PM
Screenshot 2023-12-26 at 8 39 34 PM

### DIFF
--- a/app/components/card_component.rb
+++ b/app/components/card_component.rb
@@ -14,6 +14,7 @@ class CardComponent < ApplicationComponent
     [
       "shadow",
       "rounded-lg",
+      "h-full",
       "bg-white"
     ].compact.join(" ")
   end
--- a/app/components/card_component.rb
+++ b/app/components/card_component.rb
@@ -14,7 +14,6 @@ class CardComponent < ApplicationComponent
     [
       "shadow",
       "rounded-lg",
-      "h-full",
       "bg-white"
     ].compact.join(" ")
   end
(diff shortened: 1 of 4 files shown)


app/components/card_component.rb
```diff
@@ -14,6 +14,7 @@ def card_classes_wrapper
     [
       "shadow",
       "rounded-lg",
+      "h-full",
```
zspencer, line 17: I think this should probably not be in this pr; but I am not going to fiddle with it more tonight.

app/views/layouts/application.html.erb
```diff
@@ -5,7 +5,7 @@
   <body>
     <%= render 'layouts/header' %>
 
-    <main id="<%= current_space.present? ? dom_id(current_space) : nil %>">
+    <main id="<%= current_space.present? ? dom_id(current_space) : nil %>" class="gap-3">
```
zspencer, line 8: Same with this.

anaulin (approved):
Looks good, thank you!

I would like to keep some kind of hover action, or at least the arrow icon, just to make sure that the cards still look/feel like a navigation component.

zspencer:
Screenshot 2023-12-27 at 6 23 41 PM

OK this is where I landed; it's not perfect (I can't get the arrow to align well with the words ffs) but it now has a hover state to make it more obvious that it's button; and it has an arrow which should also make it more visually buttony.

PR #2058, opened 2023-12-21
🐞🔨 `Section`: Permit setting the `#description`

zspencer:

- I forgot this in https://github.com/zinc-collective/convene/pull/2055. Which probably means I should write a system spec.


anaulin (approved):
Oops.

PR #2060, opened 2023-12-21
Allows configuration of Square from Order Notifications

rosschapman:
See comments for discussion.


rosschapman:
Here's an initial idea to put Square configuration inside Order Notifications. 

Simple UX logic:
```
Any fields missing -> show form on load
All fields filled -> hide form on load

AND

User can toggle the form at any time
```

screencast 2023-12-21 14-45-43_optim

app/furniture/marketplace/marketplace.rb
```diff
@@ -122,8 +122,6 @@ def square_connection
       end
     end
 
-    private
```
rosschapman, line 125: Now used outside class: https://github.com/zinc-collective/convene/pull/2060/files#diff-374c7d8560245acc3e9add7b4616d66e26a23b9da37061ce623cf076f4c1f3dfR17

app/furniture/marketplace/marketplaces_controller.rb
```diff
@@ -9,5 +9,17 @@ def show
     def edit
       authorize(marketplace)
     end
+
+    def update
+      if authorize(marketplace).update(marketplace_params)
+        redirect_to marketplace.location(child: :notification_methods), notice: t(".success", name: marketplace.room.name)
```
rosschapman, line 15: 🤔 Actually not seeing any `notice` on screen....

app/furniture/marketplace/routes.rb
```diff
@@ -1,7 +1,7 @@
 class Marketplace
   class Routes
     def self.append_routes(router)
-      router.resources :marketplaces, only: [:show, :edit], module: "marketplace" do
```
rosschapman, line 4: Rails was complaining about no PATCH route so I opened this up.

app/furniture/marketplace/marketplaces_controller.rb
zspencer: In rails, the `flash` method sets a cookie; which is included in the response payload of the request response cycle. 

Because every request in the `Marketplace` is running through `Turbo`,  the only bits that are re-rendered are the elements within the turbo frame.

If you want to see data in `flash[:notice]`, you'll need to add them to something within that frame you could [the `marketplace` layout](https://github.com/zinc-collective/convene/blob/30d64fe782444d2b97c83cfe78e54cc644619e61/app/furniture/layouts/marketplace.html.erb#L3-L10) to include the flash messages.

app/furniture/marketplace/notification_methods/index.html.erb
```diff
@@ -1,21 +1,29 @@
...
+        <%= render ButtonComponent.new(
+          label: "#{t('marketplace.notification_methods.new.link_to')} #{t('icons.new')}",
+          title: t('marketplace.notification_methods.new.link_to'),
+          href: marketplace.location(:new, child: :notification_method),
+          method: :get,
+          scheme: :secondary) %>
+      <%- end %>
+    </div>
+    <%= render CardComponent.new(dom_id: dom_id(notification_method), classes: "flex flex-col justify-between gap-y-2 w-full") do %>
+      <details <%= "open" if !marketplace.square_order_notifications_enabled? %>>
```
zspencer, line 17: I didn't think `<details>` could hold a `<form>`! But it makes sense that they can!

zspencer (commented):
While we can start by shoveling this onto the `Marketplace` itself, it's likely this will want to become a sub-model similar to the [`Marketplace::StripeAccount`](https://github.com/zinc-collective/convene/blob/main/app/furniture/marketplace/stripe_accounts_controller.rb)

However that can wait until we start wanting to redirect to somewhere other than `NotificationMethods#index` places after a successful `Marketplace#update`.

A `furniture/marketplace/configuring_square_system_spec.rb` would be neat!

app/furniture/marketplace/notification_methods/index.html.erb
rosschapman: Yeah. This was my clever way of avoiding having to render out to a separate page like the Marketplace `:edit` route. It's a nice stop gap before this lives in it's own `Utility`.

app/furniture/marketplace/marketplaces_controller.rb
```diff
@@ -9,5 +9,22 @@ def show
...
+
+        if marketplace_params[:square_access_token] || marketplace_params[:square_location_id]
+          flash.notice = "Square notification settings updated succesfully!"
+          flash.alert = "Square notification settings were not upated. Please try again or contact your site administrator."
+        end
```
rosschapman, line 20: Will add template tag to render flash messages within the turbo'd page scope later.

app/furniture/marketplace/marketplace.rb
```diff
@@ -122,10 +122,8 @@ def square_connection
...
-
     def square_order_notifications_enabled?
-      square_location_id.present?
+      square_location_id.present? and square_access_token.present?
```
zspencer, line 126: Bear in mind that `&&` and `and` have different operator precedence; so if `square_location_id.present?` is `false` and `square_access_token.present?` is `true`; I think this will return true.

app/furniture/marketplace/marketplaces_controller.rb
zspencer: Why are the notice and alert in the same `if`?

app/furniture/marketplace/routes.rb
zspencer: You should be able to add `:update` rather than removing the `only`, which exposes `create` and `destroy`

zspencer (approved):
I would recommend a unit test of the `square_notifications_enabled?` method; since I am 83.2% sure that it will give you surprising results in cases where predicate method calls do not return the same value.

app/furniture/marketplace/marketplaces_controller.rb
rosschapman: My goal here was to set custom message text for _any_ interaction with the Square settings.

rosschapman: Oh oops, I actually got confused. This was a mistake 😅

zspencer: Well, it will definitely set a custom message for any interaction! They'll just be... a bit contradictory. That's all!

app/furniture/marketplace/marketplace.rb
rosschapman: Got my operators mixed, didn't even realize ruby supported `and`.

zspencer: Ah, all the more reason for an isolated test of this method then!

PR #2062, opened 2023-12-22
Adds section description hint and makes styling consistent

rosschapman:
While reviewing https://github.com/zinc-collective/convene/pull/2055 I felt Space managers might be interested to know how `Description` would be used. 

Before: 

image

After: 
image

### DIFF
--- a/app/views/application/_text_area.html.erb
+++ b/app/views/application/_text_area.html.erb
@@ -1,5 +1,8 @@
 <div class="<%= local_assigns[:container_classes]%>">
   <%= form.label attribute %>
+  <% if local_assigns[:label_hint] %>
+    <small><%= local_assigns[:label_hint] %></small>
+  <% end %>
   <%= form.text_area attribute, class: local_assigns[:field_classes] %>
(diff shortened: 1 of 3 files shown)


app/views/application/_text_area.html.erb
```diff
@@ -1,5 +1,8 @@
 <div class="<%= local_assigns[:container_classes]%>">
   <%= form.label attribute %>
+  <% if local_assigns[:label_hint] %>
+    <small><%= local_assigns[:label_hint] %></small>
+  <% end %>
```
zspencer, line 5: Thank you! I was thinking ti would be nice to have a hint! But I didn't add it at the moment.

zspencer (approved):
I dig it! Should the remove section bit be in it's own `Card` too?

Feel free to merge once tests are green!

anaulin (approved):
This is great, thank you!

Eventually (soon?) it would be nice to create either a component or some shared CSS styles for displaying this kind of hint. I feel like there are many more places in the app where we should have them.

PR #2071, opened 2023-12-27
🧹 `CardComponent`: Cards fill as much space as they can

zspencer:

This won't do anything in most cases; but if a `CardComponent` is inside of an element with height, such as a `grid`; it will take up the entire thing.

It's possible that we should be using `items-stretch` or something?

Anyway, here's a few places where cards are:

Screenshot 2023-12-26 at 8 39 59 PM
Screenshot 2023-12-26 at 8 39 51 PM
Screenshot 2023-12-26 at 8 39 41 PM
Screenshot 2023-12-26 at 8 39 34 PM

### DIFF
--- a/app/components/card_component.rb
+++ b/app/components/card_component.rb
@@ -14,6 +14,7 @@ class CardComponent < ApplicationComponent
     [
       "shadow",
       "rounded-lg",
+      "h-full",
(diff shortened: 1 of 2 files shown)


anaulin (approved):
Thanks!

Issue #2043, opened 2023-12-19
🕺 `Marketplace`: `Shopper`


zspencer:
- In the https://github.com/zinc-collective/convene/issues/831, a https://github.com/zinc-collective/convene/issues/2043's is the one https://github.com/zinc-collective/convene/issues/1326. 

## Use Cases
- https://github.com/zinc-collective/convene/issues/1326

## Features
- ✨ `Marketplace`: Saved Payment Methods
- ✨ `Marketplace`: Saved Delivery Locations

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #2072, opened 2023-12-28
🌸 `Marketplace`: `Shopper` adds `Product` to `Cart` through `Menu`

zspencer:
- https://github.com/zinc-collective/convene/issues/2043
- https://github.com/zinc-collective/convene/issues/2153

While we definitely need a `Cart`; treating the `Cart` as the `Menu` is a clunky ducky way of operating.

I'm going to start driving towards a more `Menu` centric approach so the Shopping experience looks and feels a bit less nerd-forward.

## After
Screenshot 2024-01-24 at 18-41-58 Zee's Pizza
Screenshot 2024-01-24 at 18-41-44 Zee's Pizza

### DIFF
--- a/app/furniture/marketplace/marketplace_component.html.erb
+++ b/app/furniture/marketplace/marketplace_component.html.erb
@@ -1,5 +1,9 @@
 <div class="grid grid-cols-1 gap-6">
+
+
   <%= render delivery_area_component %>
+
+  <%= render Marketplace::MenuComponent.new(marketplace:, cart:) %>
 
   <%= render cart %>
 </div>
--- a/app/furniture/marketplace/menu/product_component.html.erb
+++ b/app/furniture/marketplace/menu/product_component.html.erb
@@ -1,19 +1,17 @@
 <%= render CardComponent.new(dom_id: dom_id(product)) do |card| %>
-  <%- card.with_header do %>
+  <%- card.with_header(variant: :no_padding) do %>
     <% if product.photo.present? %>
       <figure>
-        <%= image_tag hero_image, class: "w-full" %>
-        <figcaption class="px-4">
+        <%= image_tag hero_image, class: "rounded-t-lg w-full" %>
+        <figcaption class="px-2 pt-4 sm:px-4">
           <h3><%= name %></h3>
-          <%- if product.archived? %>
-            <span class="italic">(archived)</span>
-          <%- end %>
         </figcaption>
       </figure>
     <%- else %>
       <h3 class="px-4"><%= name %></h3>
     <% end %>
   <%- end %>
+
   <div class="text-sm italic">
     <%= description %>
   </div>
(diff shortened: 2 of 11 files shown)


zspencer:
@rosschapman - To decrease the complexity of this PR, I've pulled out one of the larger refactors I had included in my hackery: https://github.com/zinc-collective/convene/pull/2156

I'm also going to pull out the `QuantityPicker` bits, because that looks like a change that can also be made independently.

zspencer:
And wam! Quantity Picker lives in it's own PR: https://github.com/zinc-collective/convene/pull/2157

This should bring this patch to just the substantive workflow changes.

app/furniture/marketplace/menu/product_component.html.erb
```diff
@@ -0,0 +1,32 @@
...
+  </div>
+
+  <%- card.with_footer do %>
+    <%- cart_product = cart.cart_products.find_by(product:) %>
+
+    <%- if !cart_product %>
+      <%= button_to("Add to Cart", cart.location(child: :cart_products), method: :post,  params: { cart_product: { product_id: product.id, quantity: 1 } }, class: "w-full --secondary") %>
+    <%- else %>
+      <%= render cart_product.quantity_picker %>
+    <%- end %>
```
rosschapman, line 30: Could the first branch be part of the component. Would you ever not want to render these two branches together?

app/furniture/marketplace/marketplace_component.html.erb
```diff
@@ -1,5 +1,9 @@
 <div class="grid grid-cols-1 gap-6">
+
+
   <%= render delivery_area_component %>
 
+  <%= render Marketplace::MenuComponent.new(marketplace:, cart:) %>
```
rosschapman, line 6: Should this be called `ProductList` or something that's not so semantically related with _food_? Also the domain concept of a "menu of products" and the UI concept of "menu of items" is a little confusing.

rosschapman (approved):

app/furniture/marketplace/marketplace_component.html.erb
zspencer: I think I would prefer to lean more towards the `Marketplace` being a `CateringMarketplace` until we have other clients, rather than trying to make the underlying systems more generic.

app/furniture/marketplace/menu/product_component.html.erb
zspencer: We could, but we don't have a `quantity_picker` without a `cart_product` and I'm hesitant to instantiate a `cart_product` for every `product `on the page; because once you do it's in the in-memory cache that the `cart` has that `cart_product`; when it doesn't really.

rosschapman: Ah ah right. I got a little confused that `QuantityPicker` is derived from `ActiveModel` and not `Application Component`. My suggestion works if the quantity picker is a View Component class with the `card_product` object passed in. Is `QuantityPicker` a view component in disguise? 😅

zspencer: It's definitely a `ViewComponent` in disguise. I tend to start from the `Model` layer and let things bubble out from there. For it to work as a `ViewComponent` it will need either the`cart` and `product` passed in or the `cart_product`. The main risk to mitigate is the "oops, we instantiated a new `CartProduct` associated with a `cart` and now it looks like we have more items in the cart  than we actually do!"

Which maybe I will do at some point but not today.

PR #2074, opened 2023-12-28
🥔✨ `Marketplace`: `#restore` `Products` and `DeliveryAreas`

zspencer:
- https://github.com/zinc-collective/convene/issues/2014
- https://github.com/zinc-collective/convene/issues/2023

This allows a `Product` and `DeliveryArea` to be Restored; and also makes their forms look the same.

### DIFF
--- a/app/furniture/marketplace/delivery_area_policy.rb
+++ b/app/furniture/marketplace/delivery_area_policy.rb
@@ -9,7 +9,7 @@ class Marketplace
     alias_method :update?, :create?
 
     def permitted_attributes(_)
-      [:label, :price, :order_by, :delivery_window]
+      [:label, :price, :order_by, :delivery_window, :restore]
     end
 
     class Scope < ApplicationScope
--- a/spec/furniture/marketplace/delivery_areas_system_spec.rb
+++ b/spec/furniture/marketplace/delivery_areas_system_spec.rb
@@ -9,6 +9,26 @@ describe "Marketplace: Delivery Areas", type: :system do
     sign_in(space.members.first, space)
   end
 
+  describe "Restoring Delivery Areas" do
+    let!(:delivery_area) do
+      create(:marketplace_delivery_area, :archived, marketplace:,
+        label: "Oakland", price_cents: 10_00)
+    end
+
+    it "Makes the DeliveryArea selectable" do
+      visit(polymorphic_path(marketplace.location(child: :delivery_areas)))
+      click_link("Archived Delivery Areas")
+      within("##{dom_id(delivery_area)}") do
+        click_link("Edit")
+      end
+
+      click_button(I18n.t("restore.link_to"))
+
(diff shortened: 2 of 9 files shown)


spec/furniture/marketplace/delivery_areas_system_spec.rb
```diff
@@ -9,6 +9,26 @@
...
+    let!(:delivery_area) do
+      create(:marketplace_delivery_area, :archived, marketplace:,
+        label: "Oakland", price_cents: 10_00)
+    end
+
+    it "Makes the DeliveryArea selectable" do
+      visit(polymorphic_path(marketplace.location(child: :delivery_areas)))
+      click_link("Archived Delivery Areas")
+      within("##{dom_id(delivery_area)}") do
+        click_link("Edit")
```
rosschapman, line 22: Is there any chance delivery area data persists/leaks across cases so you'd want to do `click_link("Edit", :match => :first)` etc...?

rosschapman (approved):

spec/furniture/marketplace/delivery_areas_system_spec.rb
zspencer: The `within('##{dom_id(delivery_area)}') do ... end` scopes this click_link to child elements of the DOM element with the delivery area's ID. This should mitigate leakage across test case runs.

That said, leakage across test case runs are very unlikely to the point of "not worth worrying about" in most rails apps using RSpec because of [rspec's transactional fixtures feature](https://github.com/zinc-collective/convene/blob/main/spec/rails_helper.rb#L40).

app/furniture/marketplace/delivery_area_policy.rb
```diff
@@ -9,7 +9,7 @@ def create?
     alias_method :update?, :create?
 
     def permitted_attributes(_)
-      [:label, :price, :order_by, :delivery_window]
+      [:label, :price, :order_by, :delivery_window, :restore]
```
rosschapman, line 12: Is it ruby/rails syntax convention to horizontally stack lists? I guess I've gotten used to left-aligned vertical lists for collections so I find them easier to scan that way. 😅

zspencer: Eh, it's kind of a "six in one half dozen in the other" for me. I can imagine there's legibility benefits to vertical lists rather than horizontal ones; and am happy to receive feedback on a case-by-case basis for where it's useful.

rosschapman: I'll think about it 🤨

spec/furniture/marketplace/delivery_areas_system_spec.rb
rosschapman: OH right right I see that scope now. Coolio.

PR #2081, opened 2024-01-04
🐛🔨 `Section`: Shorten Factory Generated `Room#description`

zspencer:

Apparently, some of the Expanse Quotes are quite long; making them [ill-suited for a 300 character limited
field](https://github.com/zinc-collective/convene/actions/runs/7403679313/job/20145303706?pr=2079).

This shortens those down so they don't explode.

### DIFF
--- a/spec/factories/room.rb
+++ b/spec/factories/room.rb
@@ -16,7 +16,7 @@ FactoryBot.define do
     end
 
     trait :with_description do
-      description { Faker::TvShows::TheExpanse.quote.truncate(300) }
+      description { Faker::TvShows::TheExpanse.quote.truncate(Room::DESCRIPTION_MAX_LENGTH) }
     end
 
     trait :with_furniture do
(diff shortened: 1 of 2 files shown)


spec/factories/room.rb
```diff
@@ -16,7 +16,7 @@
     end
 
     trait :with_description do
-      description { Faker::TvShows::TheExpanse.quote }
+      description { Faker::TvShows::TheExpanse.quote.truncate(300) }
```
rosschapman, line 19: Could this value be stored in a constant, like `Room::DescriptionLength`? It feels magical here.

rosschapman (approved):

spec/factories/room.rb
zspencer: I could go with `Room::MAXIMUM_DESCRIPTION_LENGTH`.

zspencer: I went with `DESCRIPTION_MAX_LENGTH` so the starting letter for the constant matches the attribute it is used to validate.

PR #2083, opened 2024-01-04
Make all buttons look uniform, no matter if they use the `ButtonComponent` or not

anaulin:
This makes it possible for form buttons to look the same as buttons created via the `ButtonComponent`, by using the same CSS classes.

It also:
* updates the look of the buttons in the Agreements section of space management, so that they use SVG icons
* adds the missing `:money` SVG icon
* documents where we've been getting the icons from 

#### Before
image

image

#### After
image

image

### DIFF
--- a/app/assets/stylesheets/components/button.scss
+++ b/app/assets/stylesheets/components/button.scss
@@ -1,21 +1,20 @@
-/* @todo DELETE ME */
 @layer components {
   [type="reset"],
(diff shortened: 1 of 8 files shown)


zspencer (approved):
Thank you! I wonder what it would take to ditch the `ButtonComponent` completely; since I don't think it's as useful as it was.

PR #2084, opened 2024-01-04
🧹 `Marketplace`: Wrap `Shopper` view, Consolidate Turbo Frames,

zspencer:
- While working on https://github.com/zinc-collective/convene/pull/2072, I started running into all kinds of weirdness because of how I had flailed my way to a semi-working turbo-framey experience.

Now that I have a bit better handle on things; I think we can get away with a single TurboFrame for the Marketplace; as shown in the `layouts/marketplace.html.erb`, and
`marketplace/_in_room_template.html.erb`.

There will still be sub-frames for when we're replacing particular pieces of the UI (for example, when setting up `Delivery` information, but by and large this structure appears to make things quite nice.

One side effect of this change is that the `ManagementComponent` is always visible to `Members` of the `Space`. On the one hand; this seems kinda not like the worst thing; since it does away with the awkward `Manage Marketplace` button.

On the other; it means there's not really a way to "Preview" the Marketplace other than opening a Private Browsing Window; which may not be ideal.

Anyway, the good news is it should be a little easier to move to a browsable `Menu` rather than our awkward CartMenuAmalgamation.


app/furniture/marketplace/_in_room_template.html.erb
```diff
@@ -0,0 +1 @@
+<%= turbo_frame_tag(furniture, src: furniture.location, data: { turbo_action: :advance }) %>
```
zspencer, line 1: Unfortunately, this has an annoying little flash every single time. I think what I want to do is extract the major substance of the `layouts/marketplace.html.erb` into a component or something and use it both here and in the layout.

app/furniture/marketplace/marketplace_component.html.erb
```diff
@@ -1,18 +1,5 @@
-<h2 class="ml-2 my-2">
-  <%= t('.header', space: marketplace.space.name, room: marketplace.room.name) %>
-</h2>
+<div class="grid grid-cols-1 gap-6">
+  <%= render delivery_area_component %>
 
-
-<%= render onboarding_component %>
```
rosschapman, line 6: So now this will only show in the `ManagementComponent` right? Nice.

rosschapman:
> the ManagementComponent is always visible to Members of the Space

Why don't we put the management controls on a separate route/page altogether that is reached through a button in the header? Was there a reason for bundling these controls into a component that next to other interfaces?

rosschapman (approved):

app/furniture/marketplace/_in_room_template.html.erb
rosschapman: Flash how?

zspencer: Like, there is nothing there then *pow* it flashes into existence.

zspencer:
> > the ManagementComponent is always visible to Members of the Space
> 
> Why don't we put the management controls on a separate route/page altogether that is reached through a button in the header? Was there a reason for bundling these controls into a component that next to other interfaces?

We could totally make an `marketplace/admin/` namespace and route set! I didn't want to take that step with this; but I think it should be doable.

app/furniture/marketplace/shopper_frame_component.html.erb
rosschapman, line 1: Putting "frame" in the filename seems confusing with turbo frames. What about "layout" instead?

app/furniture/marketplace/marketplace_component.html.erb
zspencer: Yep!

app/furniture/marketplace/shopper_frame_component.html.erb
zspencer: I think I can get away with this just being in the layout period-end-of-story and remove the component.

PR #2090, opened 2024-01-05
Displays flash messages on the Marketplace

rosschapman:
After: 

Ross Dev Coop 2024-01-03 19-55-01
Ross Dev Coop 2024-01-03 19-55-31

### DIFF
--- a/app/furniture/marketplace/marketplaces_controller.rb
+++ b/app/furniture/marketplace/marketplaces_controller.rb
@@ -18,7 +18,6 @@ class Marketplace
           flash.notice = "Square notification settings updated succesfully!"
         end
       else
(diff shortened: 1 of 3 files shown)


app/furniture/marketplace/marketplaces_controller.rb
```diff
@@ -18,7 +18,6 @@ def update
           flash.notice = "Square notification settings updated succesfully!"
         end
       else
-        flash.alert = "Square notification settings were not upated. Please try again or contact your site administrator."
```
rosschapman, line 21: Oops, didn't mean to leave this in here. Not too concerned yet with how we handle failure.

zspencer (approved):
Very cool, nice test!

PR #2101, opened 2024-01-09
✨ `Section`: `Member` may upload a `HeroImage`

rosschapman:
- https://github.com/zinc-collective/convene/issues/1961

UI needs some work, but the backend is basically there. 

Before an image is added:
image

After image is added, with control to remove image. User must hit `Save Changes` button (outside of this gif) to apply removal. 
Kapture 2024-01-09 at 13 27 08


app/controllers/rooms_controller.rb
```diff
@@ -25,6 +25,10 @@ def create
 
   def update
     respond_to do |format|
+      if params[:room][:remove_image] == "1"
```
rosschapman, line 28: 🙋 Is this a Rails-y way of doing this?

app/models/room.rb
```diff
@@ -1,5 +1,9 @@
 # A Room in Convene acts as a gathering place.
 class Room < ApplicationRecord
+  has_one_attached :image do |attachable|
+    attachable.variant :thumb, resize_to_limit: [150, 150]
```
rosschapman, line 4: I don't see the variant being saved locally so jury is still out on this line of code.

rosschapman:
@zspencer here's a first draft at Section images with ActiveStorage.

app/controllers/rooms_controller.rb
zspencer: I would probably have a `Room::ImagesController` that has a `#destroy` action; and then I would send `DELETE /rooms/:room_id/images/:image_id` if a `Room` has_many `Images`. If it's a `has_one`, I would send a `DELETE /rooms/:room_id/image`

zspencer (approved):
I think this is probably good enough where it's at; but a little less RESTy than I would like. 

The way I am thinking about this, is each `Section` has one-or-more `Image`s which are used for things like:

- Feed Forward on the `SectionNavigation` component
- A Hero Image on the `Section#show` page
- Teaser Images for when people pass a link to the Section

The `Image` itself may have additional context, such as a title, alt text, etc. and isn't purely a  webp/png/jpeg/etc.

As a result, I would lean towards a dedicated model, perhaps named a `Section::Visual`; and offer management of that as a full resource.

Again, I think this is a good step tho!

app/controllers/rooms_controller.rb
```diff
@@ -25,11 +27,23 @@ def create
...
+      new_media.upload.attach(room_params[:hero_image_upload])
+      room_params.delete("hero_image_upload")
+      room_params_for_update = {}.merge(
+        room_params,
+        {
+          hero_image: new_media
+        }
+      )
+
+      if room.update(room_params_for_update)
```
rosschapman, line 40: @zspencer controller handling for the file upload and media creation before update. I'm not confident this approach is best practice for handling relationships in a single controller action. Or maybe there's a better way to setup the form and accepted parameters. But this helped me prove I could test upload files through the UI and create the records correctly.

zspencer (approved):
This works for now; and I like that we've got a relationship between the Room and it's `hero_image`/`feed_image`

db/schema.rb
```diff
@@ -265,6 +265,8 @@
     t.datetime "updated_at", null: false
     t.uuid "space_id"
     t.string "description"
+    t.uuid "hero_image_id"
+    t.index ["hero_image_id"], name: "index_rooms_on_hero_image_id"
```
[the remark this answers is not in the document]
rosschapman: Ha, Rails defaults to adding indexes for belongs_to apparently, and I didn't take as close a look at the migration as I thought as I was playing with the generator. Yeah, not necessary. I'll remove it.

app/controllers/rooms_controller.rb
rosschapman: I'm cool with making it RESTier, and can work on this in the next iteration which can include prototyping an interface for managing multiple Room images with more concrete functional specs.

db/migrate/20240124011436_remove_hero_image_idx_on_rooms.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveHeroImageIdxOnRooms < ActiveRecord::Migration[7.1]
+  def change
+    # Removes unneeded index that strong migrations forces us to add when
+    # initially adding the reference/foreign key
+    remove_index :rooms, :hero_image_id
```
rosschapman, line 5: @zspencer Strong Migrations complained about performing the first migration when I tried to remove the part which adds an index. I tried a few different permutations of `add_reference` and `add_foreign_key` to no avail. But as you rightly pointed out, an index is unnecessary, hence this second migration. Maybe I missed something.

zspencer: Could you push that up? I would not write a second migration and instead use `safety_assured` in the first migration if Strong Migrations gets cranky; since choosing not to add an index for a reference is reasonable in this case; even though it's not normally.

rosschapman: Ahh, that is nifty. I should have expected there to be an escape hatch. See latest.

spec/models/room_spec.rb
```diff
@@ -6,6 +6,9 @@
   it { is_expected.to have_many(:gizmos).inverse_of(:room).dependent(:destroy) }
   it { is_expected.to belong_to(:space).inverse_of(:rooms) }
 
+  it { is_expected.to belongs_to(:hero_image) }
+  it { is_expected.to belongs_to(:feed_image) }
```
zspencer, line 10: I think this wants to be singular (`belong_to` because rspec/rubyists *LOVE* grammar!)

zspencer (approved):
Once the tests are green, 🚢 🇮🇹 !

Issue #2110, opened 2024-01-11
🌸 `DesignLibrary`: `CardComponent`


zspencer:
- Every https://github.com/zinc-collective/convene/issues/1187 has a https://github.com/zinc-collective/convene/issues/2110, right?

PR #2114, opened 2024-01-13
✨`CardComponent`: Has a Header Slot

zspencer:
- https://github.com/zinc-collective/convene/issues/2110

This is perhaps unnecessary; since `<header>` tags are easy to write; but maybe it's actually useful?

### DIFF
--- a/app/components/card_component.rb
+++ b/app/components/card_component.rb
@@ -1,4 +1,5 @@
 class CardComponent < ApplicationComponent
+  renders_one :header
   renders_one :footer
 
   private
--- a/app/components/card_component.rb
+++ b/app/components/card_component.rb
@@ -1,15 +1,17 @@
 class CardComponent < ApplicationComponent
-  attr_accessor :wrapper_classes
+  HEADER_VARIANTS = {default: "p-2 sm:p-4", no_padding: ""}
+  renders_one :header, ->(variant: :default, &block) {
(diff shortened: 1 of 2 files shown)


app/components/card_component.rb
```diff
@@ -1,32 +1,17 @@
...
-      "p-4",
-      "sm:p-6",
-      # content? is not always working as described, and is returning a proc in some cases rather than a boolean
-      ("rounded-t-none" if content.blank?)
-    ].compact.join(" ")
-  end
+  DEFAULT_FOOTER = "bg-slate-50 p-2 sm:p-4"
+  FOOTER_VARIANTS = {
+    default: DEFAULT_FOOTER,
+    action_bar: [DEFAULT_FOOTER, "flex flex-row justify-between"].join(" ")
```
rosschapman, line 10: Suggestion: create a separate `ActionCardComponent` -- with duplicate styling/code as necessary. I tend to dislike parameterizing layout or style unless it's related to component state. But that's because I've seen this type of parameterization grow unwieldy when a single component tries to handle too many use cases. We're not there yet, of course.

rosschapman (approved):

app/components/card_component.rb
zspencer: Yea, I was thinking if we got to 3+ variants I would want to start pulling out the more special-purpose ones; and I feel like an `ActionCardComponent` is a pretty good next step when that happens!

PR #2136, opened 2024-01-20
✨ `Media`: Create `Media` Model and Database Tables

rosschapman:
- https://github.com/zinc-collective/convene/issues/2137

While exploring images for [Sections](https://github.com/zinc-collective/convene/pull/2101), it began to make sense for us to create a general purpose _media_ resource to handle file uploads. No time like the present, particularly because this avoids the complicatedness of writing forms or controller handling to manage the nested belongs_to image resource for Sections. Yes, Rails provides mechanisms for that -- `fields_for`, `accepts_nested_attributes`, etc... -- but we know that a _Media_ is a useful domain object on its own for any user experience beyond uploading a single image alongside other Section (currently _Room_)attributes

### DIFF
--- a/app/furniture/marketplace/product.rb
+++ b/app/furniture/marketplace/product.rb
@@ -4,6 +4,7 @@ class Marketplace
   class Product < Record
     include Archivable
 
+    # TODO: Refactor to use Media model
     has_one_attached :photo, dependent: :destroy
 
     self.table_name = "marketplace_products"
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -1,0 +1,4 @@
+# The Media resource manages file uploads to the platform
+class Media < ApplicationRecord
+  has_one_attached :upload
+end
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -1,4 +1,4 @@
 # The Media resource manages file uploads to the platform
 class Media < ApplicationRecord
-  has_one_attached :upload, dependent: :destroy
+  has_one_attached :upload
 end
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -1,4 +1,7 @@
 # The Media resource manages file uploads to the platform
 class Media < ApplicationRecord
+  # NOTE: Dependent destroy is defaulted, but when it becomes important to
+  # separate the destroy request and purge operations, let's add the
+  # `dependent: :purge_later` option.
   has_one_attached :upload
 end
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -1,4 +1,4 @@
 # The Media resource manages file uploads to the platform
 class Media < ApplicationRecord
-  has_one_attached :upload
+  has_one_attached :upload, dependent: :destroy
 end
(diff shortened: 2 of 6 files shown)


app/furniture/marketplace/product.rb
```diff
@@ -4,6 +4,7 @@ class Marketplace
   class Product < Record
     include Archivable
 
+    # TODO: Refactor to use Media model
```
rosschapman, line 7: I can create an issue and tag it here if that makes sense?

spec/models/media_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
+
+RSpec.describe Media do
+  it { is_expected.to have_one_attached(:upload) }
```
rosschapman, line 4: 🤔 Unfortunately shoulda-matchers doesn't support chaining a `.dependent(:destroy)` check here.

app/models/media.rb
```diff
@@ -0,0 +1,4 @@
+# The Media resource manages file uploads to the platform
+class Media < ApplicationRecord
+  has_one_attached :upload
```
rosschapman, line 3: Note: According the rails docs for `has_one_attached`: 
```txt
If the :dependent option isn’t set, the attachment will be purged (i.e. destroyed) whenever the record is destroyed.
```
However we could set the option to perform the attachment purge in a background job, if that becomes preferable:
```ruby
dependent: :purge_later
```

app/furniture/marketplace/product.rb
zspencer: Sure! Link it up as a 🧹to https://github.com/zinc-collective/convene/issues/1428!

app/models/media.rb
zspencer: Nice, I think background job is slightly preferable from a "reduce ops-between-request-and- response" perspective; but I would be surprised if it was meaningful within the next few years.

db/schema.rb
```diff
@@ -10,19 +10,19 @@
...
 
   create_enum :invitation_status, [
     "pending",
-    "sent",
     "accepted",
     "rejected",
     "expired",
     "ignored",
     "revoked",
+    "sent",
```
zspencer, line 25: This looks like an accident, may want to `git checkout -p main db/schema.rb` and get rid of the stuff that's not related to the change you're making

zspencer (approved):
Nice! Would love it if we cleaned up the schema changes before merging; but it's also possible those are the actual schema if more than just you are experiencing schema drift!

app/furniture/marketplace/product.rb
zspencer: https://github.com/zinc-collective/convene/issues/2138

db/schema.rb
rosschapman: Perhaps one of these days I'll memorize the patch options 🤣

`[y,n,q,a,d,j,J,g,/,s,e,?]` 😳😳😳😳😳😳😳😳

PR #2145, opened 2024-01-24
🧹 Our testing strategy has consolidated into rspec

zspencer:
This  is prompted by the failing cucumber builds over the last dozen or so Actions:
Screenshot 2024-01-23 at 7 52 49 PM

Initially, we chose node for our system tests because one of our contributors was trying to learn JavaScript for their job. They got the opportunity to learn more about how to do JavaScript, including work with the debugger. We got system tests! A win-win! We added in Cucumber because it was a tool I like and I think they were using at the time. That wasn't as much of a win, but it was a step in the journey so I accept it for what it was.

As we've moved forward we've used the nodejs bits less and less; and switched to writing end-to-end tests in ruby using capybara; which is far more common for rails projects.

I am sending these lines of code in the great `/dev/null` farm in the sky. Git shall store them for us if we ever wish to bring them back in our lives. Maybe there are other lines of un-maintained, no-longer-valuable code that they get to run around and play with.


zspencer:
Note: Because there are no cucumber tests anymore, this will never be reported; I was planning to remove that from the expected checks upon merge; but could do so before hand since they are failing elsewhere.

rosschapman (approved):
Wow, that's a lot of code. I'm whispering an elegy into `dev/null`. This is a wise act of letting go.

hero

PR #2160, opened 2024-01-29
🛠️ `Gizmos`: Gizmos that don't use S3 shouldn't fail w/o AWS

zspencer:

We initialize `active-storage` and `aws-sdk-s3` when we load the application code, which loads up `config/storage.yml` and sets up an AWS S3 Client.

But the `Journal` doesn't use AWS S3 (yet?), so the continuous integration environment doesn't have AWS credentials. Which causes [build failures].

I *could* add credentials to the Journal Gizmo's CI environment, but that would mean other Gizmos that don't use AWS S3 would need AWS credentials

So I'm going to add some fake test credentials to the .env.test.example file, which should let the AWS S3 client initialize and then explode horribly when we try and shovel some data into it.

[build failures]: https://github.com/zinc-collective/convene-journal/actions/runs/7690936384/job/20955491503


rosschapman (approved):
image

PR #2161, opened 2024-01-29
✨ `Journal`: `Author` provides `Entry#summary`

zspencer:
- When https://github.com/zinc-collective/convene-journal/issues/2, add an https://github.com/zinc-collective/convene-journal/issues/10 so we can eventually display it when https://github.com/zinc-collective/convene-journal/issues/4

### After
image

### Before
image

### DIFF
--- a/.github/workflows/test-convene-journal.yml
+++ b/.github/workflows/test-convene-journal.yml
@@ -1,0 +1,154 @@
+name: Test Journal
+on: push
+
+env:
+  POSTGRES_USER: postgres
+  POSTGRES_PASSWORD: postgres
+  # Connect to locally-running Maildev for tests
+  SMTP_PORT: 1025
+  SMTP_DOMAIN: localhost
+  SMTP_ENABLE_TLS: false
+  REDIS_HOST: redis
+  REDIS_PORT: 6379
+  HEADLESS: true
+
+jobs:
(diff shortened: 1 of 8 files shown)


.github/workflows/test-convene-journal.yml
```diff
@@ -0,0 +1,154 @@
...
+      - name: Setup CI database.yml
+        run: cp config/database.yml.github-actions config/database.yml
+
+      - name: Setup rails
+        run: bin/setup-rails && bin/rails assets:precompile
+
+      - name: Run Tests
+        env:
+          HEADLESS: true
+        run: bundle exec rspec spec/furniture/journal
```
rosschapman, line 122: What about excluding these tests from `.github/workflows/test-convene-web.yml`?

zspencer: I have some Ideas re: how to refactor our CI to split things out; but I don't know enough about the Github Actions syntax to do it.

Ideally we would have:

- An Action that creates a Slug w/all the Dev + Test dependencies
- A set of Actions that execute the tests for each Gizmo
- An Action that executes the tests for the OS itself

That way as we add gizmos the total test suite time doesn't go up.

But I also think that's a thing to deal with on a different day; once we decide it's worth it.

PR #2164, opened 2024-01-30
🧹 `Section`: Manage `Section#hero_image` as an independent Resource

rosschapman:
Allows `Space` members to upload a hero image to a `Room`. 

https://github.com/zinc-collective/convene/assets/5185/fc68a495-0178-461b-a070-92ae7b98d197


app/lib/space_routes.rb
```diff
@@ -8,6 +8,7 @@ def self.append_routes(router)
     router.resources :rooms, only: %i[show edit update new create destroy] do
       Furniture.append_routes(router)
       router.resources :furnitures, only: %i[create edit update destroy]
+      router.resources :images, controller: "room/images"
```
zspencer, line 11: Slight preference for `router.resource :hero_image`, since that is the resource being managed; not an arbitrary collection of images.

app/views/rooms/_form.html.erb
```diff
@@ -3,22 +3,29 @@
...
     <%= render "text_field", attribute: :slug, form: room_form %>
     <%= render "text_area", attribute: :description, form: room_form, label_hint: 'Add brief summary of your section for search engines to display as snippets in results' %>
+    <h3>Privacy and Security</h3>
+    <%= render "radio_group", attribute: :access_level, options: Room::access_levels.values, form: room_form %>
+    <footer>
+      <%= room_form.submit %>
+    </footer>
+  <% end %>
+  <%=  %>
+  <%= form_with url:polymorphic_path(room.location(child: :images))  do |room_images_form| %>
```
zspencer, line 13: The more resource-oriented/railsy way to do this would be `form_with(model: (room.hero_image || room.build_hero_image).location) do |hero_image_form|`.

This would also get rid of the need to pass in a `model` to the `file_field` helper.

config/locales/room/en.yml
```diff
@@ -1,7 +1,7 @@
 en:
   rooms:
-    no_furniture_notice: 'No gizmos yet. Deck out this section by adding some gizmos!'
-    place_furniture_heading: 'Add a new gizmo'
+    no_furniture_notice: "No gizmos yet. Deck out this section by adding some gizmos!"
+    place_furniture_heading: "Add a new gizmo"
```
zspencer, line 4: This looks like it is not related to this change. I don't know that we have a strong policy re: Not including stuff like this in patches or not.

On the one hand, tidying as you go when you notice something that wants tidying is a lovely practice.

On the other hand, `git history` on this line will be a little bit less useful; which may or may not matter.

config/locales/room/images/en.yml
```diff
@@ -0,0 +1,6 @@
+en:
+  room:
+    images:
+      create:
+        success: "'%{room_name}'"
+        failure: "'%{room_name}'"
```
zspencer, line 6: Are these the messages you mean to use?

spec/requests/room/images_controller_request_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+          do_request
+          expect(room.reload.hero_image).to be_present
+        end
+
+        it { is_expected.to redirect_to(room.location(:edit)) }
+      end
+
+      # context "when a hero image does exist" do
+      #   # TODO
+      # end
```
zspencer, line 32: This would be a very useful test case to flesh out!

app/controllers/room/images_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+        redirect_to [:edit, room.space, room]
+      else
+        flash[:notice] = t(".failure", room_name: room.name)
+        render :new, status: :unprocessable_entity
+      end
+    end
+
+    helper_method def room
+      @room ||= if params[:room_id]
+        current_space.rooms.friendly.find(params[:room_id])
```
zspencer, line 17: I think you want to `policy_scope(Room).friendly.find(params[:room_id]`

I also think we don't need the `if` statement at all, since if there is no `:room_id` this will never be routed to, since it's nested within the `/rooms/:room_id/` path.

zspencer (commented):
This looks like a good next step! I left some comments where I noticed things drifting off the "Rails" and am only withholding approval because the opposing test case is todod.

zspencer (approved):
This is great! Let's figure out why that test is failing and then merge away!

app/controllers/room/hero_images_controller.rb
```diff
@@ -0,0 +1,26 @@
+class Room
+  class HeroImagesController < ApplicationController
+    def create
+      old_hero_image = room.hero_image
+      new_media = Media.create
+      new_media.upload.attach(media_params[:upload])
+
+      if room.update(hero_image: new_media)
+        old_hero_image&.destroy
```
zspencer, line 9: I wonder if there's a way to push responsibility for deleting the old hero image into the model...

If we overload `Room#hero_image=`, that might work... but I actually don't know and don't know that it's worth diving into unless it feels like a solid learning experience.

app/views/rooms/hero_image/_form.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <%= render "file_field",
+          attribute: :upload,
+          form: room_images_form,
+          label: "Upload a header image",
+          label_hint: "Images are great for marketing!",
+          required: false
+      %>
+  <footer>
+    <%= room_images_form.submit %>
+  </footer>
```
zspencer, line 16: This presumes that this will always be rendered in a context where a footer is expected. I think that's fine for now; but I wonder if it indicates something is amiss...

zspencer:
🎉🎉🎉 Nice!

Issue #2170, opened 2024-02-01
🥡 `Marketplace`: `Product#servings`


zspencer:
- We capture https://github.com/zinc-collective/convene/issues/2170 so https://github.com/zinc-collective/convene/issues/2043 can make informed choices about how far their $$$ will go on a per  https://github.com/zinc-collective/convene/issues/2169  basis.

PR #2171, opened 2024-02-01
✨🥡 `Marketplace`: Add `Product#servings`

zspencer:
- https://github.com/zinc-collective/convene/issues/2170

### DIFF
--- a/app/furniture/marketplace/menu/product_component.html.erb
+++ b/app/furniture/marketplace/menu/product_component.html.erb
@@ -1,16 +1,26 @@
 <%= render CardComponent.new(dom_id: dom_id(product)) do |card| %>
-  <%- card.with_header(variant: :no_padding) do %>
-    <% if product.photo.present? %>
+
+  <% if product.photo.present? %>
+    <%- card.with_header(variant: :no_padding) do %>
       <figure>
         <%= image_tag hero_image, class: "rounded-t-lg w-full" %>
         <figcaption class="px-2 pt-4 sm:px-4">
           <h3><%= name %></h3>
+          <p class="text-sm">
+            Serves <%= product.servings %>
+          </p>
         </figcaption>
       </figure>
-    <%- else %>
-      <h3 class="px-4"><%= name %></h3>
+    <%- end %>
+  <%- else %>
+    <%- card.with_header do %>
+      <h3><%= name %></h3>
+      <p class="text-sm">
+        Serves <%= product.servings %>
+      </p>
     <% end %>
-  <%- end %>
+  <% end %>
+
 
   <div class="text-sm italic">
     <%= description %>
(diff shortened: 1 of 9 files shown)


app/furniture/marketplace/menu/product_component.html.erb
```diff
@@ -1,16 +1,26 @@
...
       </figure>
-    <%- else %>
-      <h3 class="px-4"><%= name %></h3>
+    <%- end %>
+  <%- else %>
+    <%- card.with_header do %>
+      <h3><%= name %></h3>
+      <p class="text-sm">
+        Serves <%= product.servings %>
+      </p>
```
zspencer, line 20: These 4 lines are repeated 4 times, should pull them into a `Marketplace::Product::CardHeader`? or something?

app/furniture/marketplace/product_component.html.erb
```diff
@@ -1,18 +1,22 @@
...
       </figure>
-    <%- else %>
-      <h3 class="px-4"><%= name %></h3>
     <% end %>
+  <%- else %>
+    <%- card.with_header do %>
+      <h3><%= name %></h3>
+      <p class="text-sm">Serves <%= product.servings %></p>
+    <%- end %>
   <%- end %>
```
zspencer, line 20: Or maybe *all* these lines want to be the `Marketplace::Product::HeaderComponent`?

zspencer:
Thanks for the pair @Toolbear!

db/migrate/20240201014607_marketplace_add_servings_to_products.rb
```diff
@@ -0,0 +1,5 @@
+class MarketplaceAddServingsToProducts < ActiveRecord::Migration[7.1]
+  def change
+    add_column :marketplace_products, :servings, :integer, null: true
```
rosschapman, line 3: Hmmm, serving size is very often a range. What about supporting that in next iteration? Could have some kind of composite property of `servings_min` and `servings_max`?

app/furniture/marketplace/menu/product_component.html.erb
rosschapman: There's no component too small!

rosschapman (approved):

app/furniture/marketplace/menu/product_component.html.erb
zspencer: I long for `<Marketplace::Product::CardHeader data-product="{{ product }}">` syntax 🙃

db/migrate/20240201014607_marketplace_add_servings_to_products.rb
zspencer: I think there's a couple things at play:

- Human-Servings (i.e. "It feeds 2 to 4, depending on appetite")
- FDA-Servings (i.e. "250g per serving, provides this many calories")

I think I'd rather focus my energies on widening the useful information (like the tags for Vegan, GF, etc) unless we get feedback from the client otherwise.

app/furniture/marketplace/menu/product_component.html.erb
rosschapman: Yeah, ViewComponent feels a little behind in this respect.

PR #2174, opened 2024-02-05
🧹 `Journal`: Excludes specs from primary GH workflow

rosschapman:
@zspencer in regards to https://github.com/zinc-collective/convene/pull/2161/files#r1477117473, eventually splitting and parallelizing test sounds cool -- if the project grows. In the meantime I thought this small change to the command in the main github action config would help signal in the code as to what's currently happening with splitting out journal.

### DIFF
--- a/.github/workflows/test-convene-web.yml
+++ b/.github/workflows/test-convene-web.yml
@@ -32,8 +32,7 @@ jobs:
       - name: Update apt
         env:
           DEBIAN_FRONTEND: noninteractive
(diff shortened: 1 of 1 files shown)


zspencer:
Ah, this is great! I'll enable that Workflow in this repo!

zspencer:
I'm also worried about what happens if a change to Convene core accidentally breaks something on the `Journal` or other gizmos; so I am going to close this for now.

Issue #2044, opened 2023-12-19
🧑🏽‍🍳 `Marketplace`: `Vendor`


zspencer:
## Use Cases
- [ ] Adding a `VendorRepresentative`
- [ ] Removing a `VendorRepresentative`
- [x] https://github.com/zinc-collective/convene/issues/1324

## Models
- Manages https://github.com/zinc-collective/convene/issues/2169

rosschapman:
🧹  Relegated to [cold storage](https://docs.google.com/spreadsheets/d/1Z_pC21sAKAMyB7xiRAY1--bm1ZKEuFzCzSXfqzHaF08/edit?gid=1828229424#gid=1828229424)

PR #2181, opened 2024-02-07
✨🥗 `Marketplace`: `VendorRepresentative` manages `Products`

zspencer:
- https://github.com/zinc-collective/convene/issues/2044

OK this should probably be split into a number of smaller steps, each with reasonable test coverage; but I wanted to get something working so we can turn the keys over to the folks at Oaklandia.

This makes it so:

- A `Neighborhood` `Operator` or `Space` `Member` can add a `VendorRepresentative` to a `Marketplace`
- The `VendorRepresentative` can add, remove, and update `Products`

YOLO YOLO YOLO

### DIFF
--- a/app/furniture/marketplace/vendor_representative.rb
+++ b/app/furniture/marketplace/vendor_representative.rb
@@ -1,0 +1,26 @@
+class Marketplace
+  class VendorRepresentative < Record
+    self.table_name = :marketplace_vendor_representatives
+
+    belongs_to :marketplace, inverse_of: :vendor_representatives
+    has_one :room, through: :marketplace
+    belongs_to :person
+
+    attribute :email_address, :string
+    validates :email_address, uniqueness: true, presence: true
+
+    location(parent: :marketplace)
+
+    def claimed?
+      person.present?
+    end
+
+    def claimable?
+      matching_person.present?
+    end
+
+    def matching_person
+      Person.joins(:authentication_methods).find_by(authentication_methods: {contact_location: email_address})
+    end
+  end
+end
--- a/app/furniture/marketplace/vendor_representative.rb
+++ b/app/furniture/marketplace/vendor_representative.rb
@@ -4,7 +4,7 @@ class Marketplace
 
     belongs_to :marketplace, inverse_of: :vendor_representatives
     has_one :room, through: :marketplace
-    belongs_to :person
+    belongs_to :person, optional: true
 
     attribute :email_address, :string
     validates :email_address, uniqueness: true, presence: true
@@ -16,7 +16,7 @@ class Marketplace
     end
 
     def claimable?
-      matching_person.present?
+      !claimed? && matching_person.present?
     end
 
     def matching_person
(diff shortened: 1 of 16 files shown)


app/furniture/marketplace/vendor_representative.rb
```diff
@@ -0,0 +1,26 @@
+class Marketplace
+  class VendorRepresentative < Record
+    self.table_name = :marketplace_vendor_representatives
+
+    belongs_to :marketplace, inverse_of: :vendor_representatives
+    has_one :room, through: :marketplace
+    belongs_to :person
```
rosschapman, line 7: Record creation is failing because this is not optional. Observed this while playing with the branch locally.

rosschapman (approved):

rosschapman:
@zspencer just want to confirm that a vendor rep must be _claimed_ before they actually have any write access to products etc... right? I think that workflow makes sense. Also a Person is generally created through an invitation to membership, right?

app/furniture/marketplace/vendor_representative.rb
zspencer: Weird; I was able to create a `VendorRepresentative` through the UI...

I'll add a `spec/marketplace/system/vendor_representatives_system_spec.rb` and a bit more tests tonight before merging.

anaulin: Certainly as written now this class requires to be associated to a `person` to be valid.

anaulin (approved):
I guess this is fine if it gets the job done.

I'm a bit confused about the intended workflow here. So the marketplace owner adds an owner representative, then talks to them to set up an account, then goes back to the marketplace and "claims" them, is that the intended flow?

Why not do something simpler, where a VendorRepresentative is just a bridge between a marketplace and a person, and the Marketplace owner can add it by selecting from a list of People who are members of the Space? (and if the person they want to add is not a member, they can send an invitation, making that person a member, right?)

app/furniture/marketplace/vendor_representative.rb
zspencer: yea, adding the spec caused it to fail until I put this in... Now I'm wondering why I could do it last night...

zspencer:
> Why not do something simpler, where a VendorRepresentative is just a bridge between a marketplace and a person, and the Marketplace owner can add it by selecting from a list of People who are members of the Space? (and if the person they want to add is not a member, they can send an invitation, making that person a member, right?)

Right now, being a `Member` grants wide permissions across the `Space`; we want to keep the permissions to *only* the `Marketplace`. So we can't add a `Person` from their `Membership`, since that would escalate their permissions too highly.

> I'm a bit confused about the intended workflow here. So the marketplace owner adds an owner representative, then talks to them to set up an account, then goes back to the marketplace and "claims" them, is that the intended flow?

I wanted to figure out a way to link it directly and not have to do this awkward handshake; but I couldn't come up with a way to do it without pushing the `Vendor` into `Marketplace` land. It's not ideal :(.

zspencer:
> @zspencer just want to confirm that a vendor rep must be _claimed_ before they actually have any write access to products etc... right? I think that workflow makes sense. Also a Person is generally created through an invitation to membership, right?

A `Person` is created whenever someone signs in to a `Space` in the `Neighborhood`. We probably want to move so that "`Person` registers in `Neighborhood`" is it's own distinct workflow tho, since I now understand the folly of conflating sign up and sign in.

PR #2188, opened 2024-02-08
🌸✨ `SectionNavigation` and `Section`: Show `Section#hero_image`

rosschapman:
- https://github.com/zinc-collective/convene/issues/1961

image

image

### DIFF
--- a/app/components/card_component.rb
+++ b/app/components/card_component.rb
@@ -2,7 +2,7 @@ class CardComponent < ApplicationComponent
   attr_accessor :media
   FULL_WIDTH_16_BY_9 = [1290, 726]
 
-  def initialize(media:, **kwargs)
+  def initialize(media: nil, **kwargs)
     super(**kwargs)
 
     self.media = media&.variant(resize_to_fill: FULL_WIDTH_16_BY_9)
--- a/app/components/card_component.rb
+++ b/app/components/card_component.rb
@@ -1,5 +1,17 @@
 class CardComponent < ApplicationComponent
-  HEADER_VARIANTS = {default: "p-2 sm:p-4", no_padding: ""}
+  attr_accessor :media
+  FULL_WIDTH_16_BY_9 = [1290, 726]
+
+  def initialize(media:, **kwargs)
+    super(**kwargs)
+
+    self.media = media&.variant(resize_to_fill: FULL_WIDTH_16_BY_9)
+  end
+
+  HEADER_VARIANTS = {
+    default: "p-2 sm:p-4",
+    no_padding: ""
+  }
   renders_one :header, ->(variant: :default, &block) {
     content_tag(:header, class: HEADER_VARIANTS.fetch(variant), &block)
   }
--- a/app/components/card_component.rb
+++ b/app/components/card_component.rb
@@ -1,11 +1,10 @@
 class CardComponent < ApplicationComponent
   attr_accessor :media
-  FULL_WIDTH_16_BY_9 = [1290, 726]
 
   def initialize(media: nil, **kwargs)
     super(**kwargs)
 
-    self.media = media&.variant(resize_to_fill: FULL_WIDTH_16_BY_9)
+    self.media = media
   end
 
   HEADER_VARIANTS = {
@@ -26,4 +25,8 @@ class CardComponent < ApplicationComponent
     classes += " rounded-t-none" unless content? || header?
     content_tag(:footer, class: classes, &block)
   }
+
+  def head_image
+    media&.upload&.variant(resize_to_fill: Media::FULL_WIDTH_16_BY_9)
+  end
 end
--- a/app/furniture/marketplace/marketplace_component.html.erb
+++ b/app/furniture/marketplace/marketplace_component.html.erb
@@ -1,6 +1,6 @@
 <div class="grid grid-cols-1 gap-6">
   <%- if hero_image.present? %>
-    <%= image_tag hero_image %>
+    <%= image_tag hero_image, class: "rounded-lg w-full" %>
   <%- end %>
   <%= render delivery_area_component %>
   <%= render Marketplace::MenuComponent.new(marketplace:, cart:) %>
--- a/app/furniture/marketplace/marketplace_component.html.erb
+++ b/app/furniture/marketplace/marketplace_component.html.erb
@@ -1,9 +1,26 @@
 <div class="grid grid-cols-1 gap-6">
-
-
+  <%- if marketplace.room.hero_image&.upload&.attached? %>
+    <div class="relative overflow-hidden bg-cover bg-no-repeat" style="
+        background-position: 50%;
+        background-image: url(<%= rails_blob_url(marketplace.room.image, disposition: "attachment") %>);
+        height: 250px;
+        color: white;
+        ">
+      <div
+      class="h-full w-full overflow-hidden bg-[hsla(0,0%,0%,0.30)] bg-fixed">
+        <div class="flex h-full items-center justify-center">
+          <div class="">
+            <h1 class="">
+              <%= t('.header', space: marketplace.space.name, room: marketplace.room.name) %>
+            </h1>
+          </div>
+        </div>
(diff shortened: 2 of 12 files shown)


app/components/card_component.rb
```diff
@@ -1,5 +1,17 @@
 class CardComponent < ApplicationComponent
-  HEADER_VARIANTS = {default: "p-2 sm:p-4", no_padding: ""}
+  attr_accessor :media
+  FULL_WIDTH_16_BY_9 = [1290, 726]
+
+  def initialize(media: nil, **kwargs)
+    super(**kwargs)
+
+    self.media = media&.variant(resize_to_fill: FULL_WIDTH_16_BY_9)
```
zspencer, line 8: Why are we processing this here and now, rather than lazily? It presumes that when we initialize the `CardComponent` we know which variant to use?

zspencer (commented):
I like where this is going!

app/javascript/controllers/index.js
```diff
@@ -6,3 +6,9 @@ import { application } from "./application.js";
 
 import MenuController from "./menu_controller.js";
 application.register("menu", MenuController);
+
+import SectionFormController from "./section_form_controller.js";
+application.register("section-form", SectionFormController);
+
+import SectionFormController from "./section_form_controller.js";
+application.register("section-form", SectionFormController);
```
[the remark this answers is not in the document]
rosschapman: Removed, this was when I was playing around with some javascript-y stuff in the form.

app/components/card_component.rb
rosschapman: Yeah that's a good point, I think it would be better to do this in the view.

app/furniture/section_navigation/section_navigations/_section_navigation.html.erb
```diff
@@ -1,13 +1,8 @@
 <div class="grid grid-cols-1 gap-5 sm:gap-6 sm:grid-cols-2 lg:grid-cols-4">
   <% policy_scope(section_navigation.rooms).each do |room| %>
     <%= link_to polymorphic_path(room.location), id: dom_id(room, :link_to), class: "group no-underline" do %>
-      <%= render CardComponent.new(classes:
-       "flex flex-col h-full justify-between") do |card| %>
-
-
+      <%= render CardComponent.new(media: room.hero_image&.upload, classes: "flex flex-col justify-between") do |card| %>
```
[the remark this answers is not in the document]
rosschapman: Yeah, let's pass the domain model.

app/furniture/marketplace/marketplace_component.html.erb
```diff
@@ -1,9 +1,8 @@
 <div class="grid grid-cols-1 gap-6">
-
-
+  <%- if hero_image.present? %>
+    <%= image_tag hero_image, class: "rounded-lg w-full" %>
+  <%- end %>
```
zspencer, line 4: Ah, interesting choice to include it on the `Marketplace`, rather than the `Section`. I was presuming that the `Section#hero_image` would apply to any `Section`, not just ones that use the `Marketplace`.

If you're down for moving this up to the `Section` level I would be stoked; but I think this is fine as is.

app/furniture/marketplace/product_component.rb
```diff
@@ -13,12 +13,8 @@ def edit_button
       super(title: t("marketplace.products.edit.link_to", name: name), href: location(:edit))
     end
 
-    # 16:9 of 1290 is 1290:725.625
-    # We rounded up.
-    # @see https://www.ios-resolution.com/
-    FULL_WIDTH_16_BY_9 = [1290, 726]
     def hero_image
-      product.photo.variant(resize_to_fill: FULL_WIDTH_16_BY_9)
+      product.photo.variant(resize_to_fill: Media::FULL_WIDTH_16_BY_9)
```
zspencer, line 17: Would it make sense for this code to live in `Media#hero_image` (`Product#photo` is not yet a `Media`, so it doesn't make sense right-here, but it may make sense in the other places)

app/components/card_component.rb
```diff
@@ -1,5 +1,16 @@
 class CardComponent < ApplicationComponent
-  HEADER_VARIANTS = {default: "p-2 sm:p-4", no_padding: ""}
+  attr_accessor :media
```
zspencer, line 2: I feel  like `header_media` is more specific, but not necessarily worth changing.

app/models/media.rb
```diff
@@ -1,5 +1,10 @@
 # The Media resource manages file uploads to the platform
 class Media < ApplicationRecord
+  # 16:9 of 1290 is 1290:725.625
+  # We rounded up.
+  # @see https://www.ios-resolution.com/
+  FULL_WIDTH_16_BY_9 = [1290, 726]
```
zspencer, line 6: We have full-width on smol devices; which is fine for the `Product::Card` and `SectionNavigation::SectionCard`, but probably not ideal when rendering the `Section` itself on a larger monitor.

I think it's fine-for-now to just be :this-is-fine:, but I do hope we'll figure out a way to follow up on this so hero-images-on-laptop-sized devices with retina displays are also big and beautiful.

Perhaps this is a place where media-queries will need to come into play; especially since we don't want to load  really-big-photos on a really-smol-devices.

All that said, Mobile First is probably the way to play this; and we can polish up the presentation on medium/large/xlarge screens :later:

zspencer (approved):
This is looking really lovely! And I really like how `CardComponent.new(media: ...)` shook out! Once this is merged I'll refactor the `Product::CardComponent` to use that!

app/furniture/marketplace/marketplace_component.html.erb
rosschapman: Ha, I'm embarrassed because I meant to do that. How about now? I'm just not sure I love the design because the image is so big at the top.

app/views/furnitures/_furniture.html.erb
```diff
@@ -1,8 +1,10 @@
 <% edit_href = nil %>
 <% edit_href = furniture.location(:edit) if furniture.configurable? %>
 <% edit_href = furniture.to_kind_class.location(:edit) if furniture.to_kind_class.has_controller_edit? %>
-
 <section id="<%= dom_id(furniture) %>" class="mt-4 relative">
+  <%- if furniture.room.hero_image&.upload.present? %>
+    <%= image_tag furniture.room.hero_image&.upload&.variant(resize_to_fill: Media::FULL_WIDTH_16_BY_9), class: "w-full" %>
+  <%- end %>
```
zspencer, line 7: Won't this make the hero image show up once per piece `Gizmo`?

app/furniture/marketplace/marketplace_component.html.erb
zspencer: Now it looks like it's in the iterator over ever `Gizmo`. What about putting it in `sections/_section.html.erb`  or even `sections/show.html.erb`? 

And yea, being a Hero image we would want it to be full-width; which may be intense? But maybe it's fine?

zspencer: I think maybe the section Hero Image the the Card Hero Image want to be different aspect ratios; perhaps the Section Hero Image never exceeds say, 420px high; and is as wide as the viewport?

app/views/furnitures/_furniture.html.erb
rosschapman: I was a bit hasty wandering through the layers. Just moved it up to the correct layer (I hope 😅).

zspencer (approved):
This looks good! Thank you!

app/furniture/marketplace/marketplace_component.html.erb
rosschapman: We could definitely use some design review!

app/components/card_component.rb
rosschapman: I thought about using that name, but also didn't feel strongly enough about it to change it.🤷

Issue #2189, opened 2024-02-08
🥡 `Marketplace`: `Product#tags`


zspencer:
- https://github.com/zinc-collective/convene/issues/2169

zspencer:
These shoudl probably be shared across the `Space`, just like `TaxRate`. Ideally they would signal things like `Vegan` or `Gluten Free` etc.

I'd love to store them directly as a column on the `marketplace_products table`; but I wonder if it makes more sense to store them in their own `marketplace_tags` table with a `marketplace_product_tags` join table to link them together...

Visually, I'm thinking they could be little pills underneath the `Product::CardComponent` header image...

PR #2190, opened 2024-02-08
✨ `Marketplace`: Add `Product#tags`

zspencer:
- https://github.com/zinc-collective/convene/issues/2189
    
Products can now be tagged with things like `Gluten Free` or `Vegan`,
and will show up as such under Product listings

## After
Screenshot 2024-02-08 at 8 40 46 AM

### DIFF
--- a/app/furniture/marketplace/management_component.html.erb
+++ b/app/furniture/marketplace/management_component.html.erb
@@ -21,6 +21,7 @@
         <%= link_to_child(:notification_methods, icon: :bell_alert)  if policy(marketplace.notification_methods).index? %>
         <%= link_to_child(:flyer, icon: :qr_code)  if policy(marketplace.flyer).show? %>
         <%= link_to_child(:vendor_representatives, icon: :building_storefront) if policy(marketplace.vendor_representatives).index? %>
+        <%= link_to_child(:tags, icon: :tag) if policy(marketplace.tags).index? %>
       </nav>
     <% end %>
   <% end %>
--- a/app/furniture/marketplace/policy.rb
+++ b/app/furniture/marketplace/policy.rb
@@ -2,7 +2,7 @@ class Marketplace
   class Policy < ApplicationPolicy
     def create?
       return true if current_person.operator?
-      return true if current_person.member_of?(marketplace.space)
+      return true if current_person.member_of?(space)
 
       return true if shopper&.person.blank? && !current_person.authenticated?
 
@@ -23,6 +23,7 @@ class Marketplace
 
       object.marketplace if object.respond_to?(:marketplace)
     end
+    delegate :space, to: :marketplace
 
     module SpecFactories
       def self.included(spec)
--- a/app/furniture/marketplace/tag.rb
+++ b/app/furniture/marketplace/tag.rb
@@ -6,6 +6,8 @@ class Marketplace
     has_many :product_tags, inverse_of: :tag, dependent: :destroy
     has_many :products, through: :product_tags, inverse_of: :tags
 
+    validates :label, uniqueness: {case_sensitive: false, scope: :bazaar_id}
+
     attr_accessor :marketplace
     location(parent: :marketplace)
   end
--- a/app/furniture/marketplace/tag.rb
+++ b/app/furniture/marketplace/tag.rb
@@ -1,0 +1,12 @@
+class Marketplace
+  class Tag < Record
+    self.table_name = "marketplace_tags"
+
+    belongs_to :bazaar, inverse_of: :tags
+    has_many :product_tags, inverse_of: :tag, dependent: :destroy
+    has_many :products, through: :product_tags, inverse_of: :tags
+
+    attr_accessor :marketplace
+    location(parent: :marketplace)
+  end
+end
--- a/spec/furniture/marketplace/product_tags_system_spec.rb
+++ b/spec/furniture/marketplace/product_tags_system_spec.rb
@@ -1,0 +1,46 @@
+require "rails_helper"
+
+describe "Product Tags", type: :system do
+  let(:space) { create(:space, :with_entrance, :with_members) }
+  let(:marketplace) { create(:marketplace, :ready_for_shopping, room: space.entrance) }
+
+  before do
+    sign_in(space.members.first, space)
+  end
+
+  scenario "Adding Tags to a Product" do # rubocop:disable RSpec/Capybara/FeatureMethods,RSpec/ExampleLength
+    muffins = create(:marketplace_product, marketplace:, name: "Mazin' Muffins", description: "Buttery corn muffins")
+
+    visit(marketplace)
+    click_link("Tags")
+
+    click_link("Add Tag")
+
+    fill_in("Label", with: "🚫🌾 Gluten Free")
+
+    click_button("Create")
+
+    click_link("Products")
+    within(muffins) do
+      click_link("⚙️ Edit")
+    end
+
+    visit(marketplace)
+
+    within(muffins) do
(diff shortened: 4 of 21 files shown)


app/furniture/marketplace/management_component.html.erb
```diff
@@ -21,6 +21,7 @@
         <%= render button({child: :notification_methods}, icon: :bell)  if policy(marketplace.notification_methods).index? %>
         <%= render button({child: :flyer}, icon: :receipt_percent)  if policy(marketplace.flyer).show? %>
         <%= render button({child: :vendor_representatives}, icon: :cake) if policy(marketplace.vendor_representatives).index? %>
+        <%= render button({child: :tags}, icon: :tag) if policy(marketplace.tags).index? %>
```
zspencer, line 24: This uses the same icon as `Marketplace#products`, so maybe we should consider getting a different icon for `Product`? Maybe something food-like?

app/furniture/marketplace/policy.rb
```diff
@@ -2,7 +2,7 @@ class Marketplace
   class Policy < ApplicationPolicy
     def create?
       return true if current_person.operator?
-      return true if current_person.member_of?(marketplace.space)
+      return true if current_person.member_of?(space)
```
zspencer, line 5: This allows us to have `Models` within the `Marketplace` that do not reference a `Marketplace` directly, such as `TaxRate` and `Tag`

spec/furniture/marketplace/product_tags_system_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "rails_helper"
+
+describe "Product Tags", type: :system do
+  let(:space) { create(:space, :with_entrance, :with_members) }
+  let(:marketplace) { create(:marketplace, :ready_for_shopping, room: space.entrance) }
```
zspencer, line 5: These two lines are repeated in pretty much every `marketplace/*_system_spec.rb`; maybe we pull them to a mixin?

spec/furniture/marketplace/product_tags_system_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  before do
+    sign_in(space.members.first, space)
+  end
+
+  scenario "Adding Tags to a Product" do # rubocop:disable RSpec/Capybara/FeatureMethods,RSpec/ExampleLength
```
zspencer, line 11: I think we should disable both of these cops in system specs. Thoughts?

spec/furniture/marketplace/product_tags_system_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(page).to have_content("🚫🌾 Gluten Free")
+    end
+  end
+
+  def visit(object_or_path)
+    if object_or_path.respond_to?(:location)
+      super(polymorphic_path(object_or_path.location))
+    else
+      super
+    end
```
zspencer, line 43: Technically we don't need this `else` in this case; since everything that is visited has a `location` but oh well.

.gitignore
```diff
@@ -33,4 +33,4 @@ app/assets/builds/
 
 .devcontainer/output
 
-TAGS
```
zspencer, line 36: This was making us ignore the `tags` views folder 🤦

rosschapman: TIL .gitignore will look at any directories with a declaration made at the top level 😳

rosschapman:
@zspencer can you provide some justification for implementing tags using a _relational_ architecture through a bridge table vs using an array column on a `Product`? I generally understand the tradeoffs between these two methodologies, so was hoping you could help me understand why you might prefer one way over another for Convene. 

Spitballing...I can imagine a `Product` -> `Tags` relationship using an array column, that also retains referential integrity if those tags are only chosen from a `Tags` record table. (Presuming referential integrity being one major benefit to the relational model.)

Does my question make sense?

zspencer:
To be honest, I don't really think there's enough difference to encourage one or the other in this particular use case.

My main goal here was making it so  `Members` can decide what `Tags` are allowed in the `Marketplace` context. Rails doesn't have great built-in support for searching / filtering by array fields in Postgresql; so this is the "most boring" way to go with it.

If we were to implement it as a field-on-a-row; we'd need to get into deeper arcana than I think is necessary.

app/furniture/marketplace/management_component.html.erb
zspencer: https://github.com/zinc-collective/convene/pull/2221

zspencer:
A question I think is of greater importance here is what we call these things. `label` and `tag` are both methods defined in Rails views.

So, what *should* we call them? Do we call them `tags` even though it's a name collision? In the `Journal`, we call them `Keywords`... Maybe it's actually `Product#features`?

zspencer:
@rosschapman - I lied! I forgot that I didn't have clarity on the lesser of N evils, regarding what we call these.

- label feels too generic / collides with HTML
- tag feels too generic + collides with HTML
- Maybe badge? or sticker? or is that too skeumorphic?

zspencer:
Or maybe we just go hella specific and call them `Product#dietary_restrictions` for now?

spec/furniture/marketplace/product_tags_system_spec.rb
anaulin: I don't have a strong opinion either way.

For tests like these, I don't mind duplicating a couple of lines, and I like seeing at a glance how data is being setup for the test. An intermediate solution would be to have a trait like `:with_full_space` on the Marketplace factory that creates a Marketplace with a fully-setup and populated space.

spec/furniture/marketplace/product_tags_system_spec.rb
anaulin: Yeah, that makes sense.

anaulin:
> @zspencer can you provide some justification for implementing tags using a relational architecture through a bridge table vs using an array column on a Product? I generally understand the tradeoffs between these two methodologies, so was hoping you could help me understand why you might prefer one way over another for Convene.

The justification is that Rails people have a bizarre kink for bridge tables. 😉

anaulin:
> Or maybe we just go hella specific and call them Product#dietary_restrictions for now?

Just to record a brief summary of the convo we just had during ensemble: seems like calling them `tags` gives us a bit more flexibility.

app/furniture/marketplace/tag.rb
```diff
@@ -0,0 +1,12 @@
+class Marketplace
+  class Tag < Record
```
anaulin, line 2: Not a blocker, but one way to help users keep their tags tidy would be to add a case-insensitive uniqueness constraint here, e.g:

```ruby
validates :label, uniqueness: true, scope: { :bazaar, case_insensitive: true }
```

This would prevent adding "Gluten-Free" if "gluten-free" already exists.

(Also a good excuse to start a smol `tag_spec.rb`...)

anaulin (approved):
Very nice, thank you!

app/furniture/marketplace/tag.rb
zspencer: I like enforcing a case insensitive uniqueness scoped to bazaar!

zspencer:
> The justification is that Rails people have a bizarre kink for bridge tables. 😉

Is it a kink if it's inflicted against our will through trauma... oh wait.

spec/furniture/marketplace/product_tags_system_spec.rb
zspencer: I like the solution of a `marketplace` trait for making instantiating a `marketplace` in a space with an entrance + members.

Issue #2198, opened 2024-02-10
🐞 `Marketplace`: `Buying Products` - `Checkout` missing only one `DeliveryArea`


zspencer:
- When https://github.com/zinc-collective/convene/issues/1326, you cannot Checkout if there is only one `unarchived` https://github.com/zinc-collective/convene/issues/1136

zspencer:
I've confirmed that this works on production again!

PR #2199, opened 2024-02-10
🐞🔨`Marketplace`: `Buying Products` with one unarchived `DeliveryArea`

zspencer:
- https://github.com/zinc-collective/convene/issues/2198

When we added [archiving delivery areas], I didn't take into account that `Marketplace#cart_for_shopper` was checking against all the delivery areas.

This fixes that and adds a system spec to make sure we can still checkout after!

[archiving delivery areas]: https://github.com/zinc-collective/convene/pull/2024

### DIFF
--- a/app/furniture/marketplace/carts/_footer.html.erb
+++ b/app/furniture/marketplace/carts/_footer.html.erb
@@ -10,6 +10,7 @@
       <%= render "marketplace/carts/total", cart: cart %>
     </td>
   </tr>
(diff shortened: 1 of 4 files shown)


zspencer:
It's been 7 days since I requested review; and this is a significant bug so I am going to merge it without reviewers.

rosschapman:
@zspencer thanks for taking care of this!

Issue #2197, opened 2024-02-10
🥡 `Marketplace`: `Product#description`


zspencer:
- https://github.com/zinc-collective/convene/issues/2169

## Use Cases

- [x] `Product#description` is set when https://github.com/zinc-collective/convene/issues/1324
- [x] `Product#description` is shown when https://github.com/zinc-collective/convene/issues/1326
- [ ] `Product#description` is Rich Text

## Engineering Details

The `Product#description` field uses [`ActionText`] to provide a rich text editor. 

[`ActionText`]: https://guides.rubyonrails.org/action_text_overview.html

Issue #2153, opened 2024-01-26
📋 `Marketplace`: `Menu`


zspencer:
- The https://github.com/zinc-collective/convene/issues/2153 is a paginated list of every available https://github.com/zinc-collective/convene/issues/2169 that provides the #2043 with:
  - https://github.com/zinc-collective/convene/issues/1428
  - https://github.com/zinc-collective/convene/issues/2189
  - https://github.com/zinc-collective/convene/issues/2218
  - https://github.com/zinc-collective/convene/issues/2197

PR #2200, opened 2024-02-10
🌸✨🥡 `Marketplace`: `Products#description` is Rich Text

zspencer:
- https://github.com/zinc-collective/convene/issues/2197
- https://github.com/zinc-collective/convene/issues/2153

## After
Screenshot 2024-02-09 at 6 05 06 PM

### DIFF
--- a/spec/furniture/marketplace/menu/product_component_spec.rb
+++ b/spec/furniture/marketplace/menu/product_component_spec.rb
@@ -9,16 +9,16 @@ RSpec.describe Marketplace::Menu::ProductComponent, type: :component do
   let(:product) { create(:marketplace_product, description: product_description, marketplace:) }
   let(:product_description) do
     <<~DESC.chomp
-      A delicious fritter made with love
+      <p>A delicious fritter made with love</p>
 
-      Make sure to **eat it warm!**
+      <p>Make sure to <strong>eat it warm!</strong></p>
     DESC
   end
 
   it { is_expected.to have_content(vc_test_controller.view_context.humanized_money_with_symbol(product.price)) }
   it { is_expected.to have_no_content("Serves") }
 
-  it "renders the description as markdown" do
+  it "renders the description as rich text" do
     expect(output).to have_css("p", text: "A delicious fritter made with love")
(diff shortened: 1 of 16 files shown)


zspencer:
Look @anaulin ! No italics!

anaulin (approved):
This is ok, but (1) how will users know to use Markdown in the product descriptions, and (2) are we sure we want to double down on Markdown, as opposed to using Action Text and a more user-friendly WYSIWYG editor?

zspencer:
I'm down to try out `ActionText`!

spec/furniture/marketplace/menu/product_component_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      Make sure to **eat it warm!**
+    DESC
+  end
+  let(:product) { create(:marketplace_product, description: product_description, marketplace:) }
+
+  it { is_expected.to have_content(vc_test_controller.view_context.humanized_money_with_symbol(product.price)) }
+
+  it "renders the description as markdown" do
+    expect(output).to have_css("p", text: "A delicious fritter made with love")
+    expect(output).to have_css("strong", text: "eat it warm!")
```
rosschapman, line 24: Is this correct? Shouldn't there be a `<strong>` tag around this text?

rosschapman, line 24: Is there an advantage to expecting on each part rather than doing something like: 
```ruby
expect(output).to eq(<p>A delicious fritter made with love</p> Make sure to <strong>eat it warm!<strong>
```

rosschapman (approved):

rosschapman:
Whoa ActionText! Rails has it all! 🤣

spec/furniture/marketplace/menu/product_component_spec.rb
zspencer: Eh, mostly for "flexibility"; since if we switch from say, strong to italic; the A delicious fritter made with love bit shouldn't be sad.

rosschapman (approved):

PR #2211, opened 2024-02-16
⚙️ `Gizmos`: Support `ActionText`

zspencer:
- https://github.com/zinc-collective/convene/pull/2200#pullrequestreview-1879198110

Markdown is cool, but it's nerd-forward.

ActionText is a much more human-friendly way of implementing rich text in Rails applications.

I don't know enough about [Trix] to know if it will work in multi-user contexts; or how it will work with rich-embeds, ala Notion or Google Docs

But I think it's a good-enough-default for now; and if we get to the point where embeddables or multi-user editing of a field is important we can figure that out.

ETA: This is a direct output of `bin/rails action_text:install` and `bin/rails db:migrate`; since I didn't want to conflate structural changes with behavioral changes.

[Trix]: https://trix-editor.org/

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -1,7 +1,7 @@
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 require "spec_helper"
 ENV["RAILS_ENV"] ||= "test"
-require File.expand_path("../config/environment", __dir__)
+require_relative "../config/environment"
 # Prevent database truncation if the environment is production
 abort("The Rails environment is running in production mode!") if Rails.env.production?
 require "rspec/rails"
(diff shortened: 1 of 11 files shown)


spec/rails_helper.rb
```diff
@@ -20,19 +20,20 @@
 # directory. Alternatively, in the individual `*_spec.rb` files, manually
 # require only the support files necessary.
 #
-Dir[Rails.root.join("spec", "support", "**", "*.rb")].sort.each { |f| require f }
+# Rails.root.glob('spec/support/**/*.rb').sort.each { |f| require f }
```
zspencer, line 23: We technically still need this line.
```suggestion
  Rails.root.glob('spec/support/**/*.rb').sort.each { |f| require f }
```

rosschapman (approved):
Looks interesting, curious to see how riding this rail works out!

rosschapman:
Also now I'm having childhood flashbacks...
image

PR #2249, opened 2024-02-27
Cleans up Order and Order history UI

rosschapman:
Changes are the same for guest or logged in shoppers.

Before
image
image

After
image
image

### DIFF
--- a/app/furniture/layouts/marketplace.html.erb
+++ b/app/furniture/layouts/marketplace.html.erb
@@ -1,7 +1,4 @@
 <%- content_for :content do %>
-  <%- if room.hero_image&.upload.present? %>
-    <%= image_tag room.hero_image&.upload&.variant(resize_to_fill: Media::FULL_WIDTH_SHORT), class: "w-full" %>
-  <%- end %>
   <div class="flex flex-col content-between justify-end mx-auto max-w-7xl sm:px-6 lg:px-8">
     <div class="grow mx-auto w-full">
       <%= turbo_frame_tag(marketplace, data: { turbo_action: :advance }) do %>
--- a/app/furniture/layouts/marketplace.html.erb
+++ b/app/furniture/layouts/marketplace.html.erb
@@ -1,22 +1,28 @@
 <%- content_for :content do %>
-  <%= turbo_frame_tag(marketplace, data: { turbo_action: :advance }) do %>
-    <%- if policy(marketplace).edit? %>
-      <%= render Marketplace::ManagementComponent.new(marketplace: marketplace) do %>
-        <%= yield  %>
-      <%- end %>
-    <%- else %>
-      <section>
-        <header>
-          <h2>
-            <%= t('.header', space: marketplace.space.name, room: marketplace.room.name) %>
-          </h2>
-        </header>
-
-        <div>
-          <%= yield  %>
-        </div>
-      </section>
-    <%- end %>
+  <%- if room.hero_image&.upload.present? %>
+    <%= image_tag room.hero_image&.upload&.variant(resize_to_fill: Media::FULL_WIDTH_SHORT), class: "w-full" %>
   <%- end %>
-<%- end %>
+  <div class="flex flex-col content-between justify-end mx-auto max-w-7xl sm:px-6 lg:px-8">
+    <div class="grow mx-auto w-full">
+      <%= turbo_frame_tag(marketplace, data: { turbo_action: :advance }) do %>
+        <%- if policy(marketplace).edit? %>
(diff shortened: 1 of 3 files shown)


app/furniture/layouts/marketplace.html.erb
```diff
@@ -1,22 +1,28 @@
 <%- content_for :content do %>
-  <%= turbo_frame_tag(marketplace, data: { turbo_action: :advance }) do %>
-    <%- if policy(marketplace).edit? %>
-      <%= render Marketplace::ManagementComponent.new(marketplace: marketplace) do %>
-        <%= yield  %>
+  <%- if room.hero_image&.upload.present? %>
+    <%= image_tag room.hero_image&.upload&.variant(resize_to_fill: Media::FULL_WIDTH_SHORT), class: "w-full" %>
+  <%- end %>
```
rosschapman, line 4: I think this makes sense here?

app/furniture/layouts/marketplace.html.erb
```diff
@@ -1,22 +1,28 @@
...
+  <div class="flex flex-col content-between justify-end mx-auto max-w-7xl sm:px-6 lg:px-8">
+    <div class="grow mx-auto w-full">
```
rosschapman, line 6: Adds similar chrome styling as the main layout.

app/furniture/marketplace/orders/_order.html.erb
```diff
@@ -26,11 +23,9 @@
   <hr class="col-span-6 mb-2" />
   <div class="col-span-4">Total</div>
   <div class="col-span-2 text-right"><%= humanized_money_with_symbol(order.price_total) %></div>
-
-  <%- if policy(order.events).index? %>
+  <%- if policy(order.events).index? && order.events.present? %>
```
rosschapman, line 26: Hides the Timeline altogether if there are no events yet.

app/furniture/marketplace/orders/index.html.erb
```diff
@@ -2,5 +2,5 @@
 <%- @pagy, @records = pagy(orders.paid) %>
 <div class="mx-auto w-full max-w-prose">
   <%= render @records %>
-  <%== pagy_nav(@pagy) %>
+  <%== pagy_nav(@pagy) if @pagy.pages > 1 %>
```
rosschapman, line 5: I'm not sure there's an reason to show pagination for a single page?

zspencer: Agree, I noticed that in another spot; makes me think we need a partial or Component for rendering the Pagy nav so we follow the same rule of skipping when no page

zspencer (commented):
This looks great! Only objection I have is the adding of the `Section#hero_image` to `layouts/marketplace` rather than `layouts/application.html.erb` (and maybe doing that as it's own PR)

app/furniture/layouts/marketplace.html.erb
rosschapman: In practice I didn't notice any double rendering. With the way the templates are "stacked" currently this seems to work fine. I spot checked this everywhere I could.

I'm not even that strong on having this image here. But it seemed nicer than a stark order summary while I was going through the checkout flow. Although the more I sit with it I think presenting the shopper the summary without visual embellishment might be a more focused experience. So I'll take it out.

app/furniture/layouts/marketplace.html.erb
rosschapman: What is "main"? Do you mean `layouts/application.html.erb`? I suppose it could although if we raise it to that level I wonder what cascading effects it will have on other non-room, non-marketplace pages. 

I'm a little confused on the "stacking" of templates between Rooms and the Marketplace Gizmo. Would need some more time to trace how the hierarchy to understand how to reconcile this styling across both, and/or more screens.

zspencer: Yes, I mean `layouts/application.html.erb`

And yea, the way the layouts "stack" needs a bit of refinement.

Right now, everything (except the `Marketplace::Flyer`) renders within `layouts/application.html.erb`; which probably should have the framing we want for everything.


`layouts/marketplace.html.erb` is intended for wrapping pages within the `Marketplace`, but still inherits from `layouts/application.html.erb`, so it should only include things that are significant on a per-Marketplace basis like the `ManagementComponent` frame, the `turbo_frame_tag` for the `Marketplace`, maybe some `Marketplace` specific navigation (tho I am not sold on that)

zspencer (approved):
Looks good, let's ship it!

PR #2254, opened 2024-02-28
🧹 Ensure that factory for Journal::Keyword always uses a unique canonical keyword

anaulin:
This should prevent occasional test failures caused by trying to create a duplicate keyword, such as the one in https://github.com/zinc-collective/convene/actions/runs/8070888101/job/22049297163


zspencer (approved):
🤦 thank you!

PR #2256, opened 2024-02-29
🧹 Uncomment `PG*` default lines in `.env.example`

daltonrpruitt:
In general, people seem to need these to be uncommented more often than not. 
Discussion seemed to indicated this was a desirable change.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,9 +5,9 @@
 
 # If you are using docker to manage your postgresql database,
 # you will want to uncomment these
-# PGHOST=localhost
-# PGUSER=postgres
-# PGPASSWORD=password
+PGHOST=localhost
+PGUSER=postgres
+PGPASSWORD=password
 
 
 PORT=3000


anaulin (approved):
I wonder if this particular bit is a Linux vs Mac Docker difference. I use Docker on Mac and I keep these variables commented out. Might be worth adding a comment to that effect as part of this PR (if we believe that to be the case)?

I don't have a strong opinion on this, and I'm happy with whatever makes folks' lives easier, and this is definitely safe to merge!

rosschapman:
@daltonrpruitt, as @anaulin suggested, I think a short comment to make developers aware that this default is not always necessary or desirable depending on your machine  would be useful. What about adding to `CONTRIBUTING.md` at https://github.com/zinc-collective/convene/blob/main/CONTRIBUTING.md?plain=1#L76.

rosschapman (approved):

.env.example
```diff
@@ -5,9 +5,9 @@
 
 # If you are using docker to manage your postgresql database,
 # you will want to uncomment these
-# PGHOST=localhost
-# PGUSER=postgres
-# PGPASSWORD=password
+PGHOST=localhost
+PGUSER=postgres
+PGPASSWORD=password
```
zspencer, line 10: I think we only want to uncomment the first two; as `password` is not guaranteed to be their postgresql password.

anaulin:
I personally vote against adding new separate documentation pages, such as `SETUP_FAQ.md`. In my opinion spreading the documentation across different pages makes it even harder to find whatever you're looking for. I'd rather have better-organized longer documents where I can Ctrl-F and also see an idea of the contents at a glance.

zspencer:
That's fair; Never underestimate the power of 🍀+F!

rosschapman:
Some might argue that's a better way to write code as well. 😅

anaulin:
> Some might argue that's a preferred way to write code as well. 😅

🙀🙅🏼‍♀️🚫

anaulin:
@daltonrpruitt are you still planning to work on this PR, or shall we close it for now?

anaulin:
@daltonrpruitt I'm going to close this PR to keep our number of open PRs to a minimum. Please feel free to reopen if/when that makes sense.

Issue #2243, opened 2024-02-26
🐛  Marketplace: Order link in e-mail notification doesn't work


anaulin:
The link to an order in the order notification that goes to the vendor does not work. It looks like:
http://spaces/piikup/rooms/piikup-bundle/marketplaces/8a729075-9805-43c9-a314-c4cc74abdd1c/orders/e5cbd784-0fe2-4957-9ff5-1377b3f81677

Which is not a valid URL.

rosschapman:
When testing locally the domain segment of the url is there when I click the link. Seems to be rendered correctly in the source as well. 🤔

![image](https://github.com/zinc-collective/convene/assets/5185/7570a101-55cb-457a-ba61-ddf1bf9e1797)

rosschapman:
Considering the ActionMailer [configs](https://github.com/zinc-collective/convene/blob/main/config/initializers/action_mailer.rb#L14-L15) look ok , all I can think of is that the  `APP_ROOT_URL` env var is not getting set on boot.

anaulin:
@rosschapman if your hypothesis about `APP_ROOT_URL` being unset in production is correct, that would mean all link generated by Mailers have this problem. Do the links in other emails have this problem as well? (For example, the one-time password email when you try to log in.)

rosschapman:
Good idea. Let me verify other mailers have the same issue.

PR #2257, opened 2024-02-29
Fixes absolute paths in Marketplace mailers #2243 

rosschapman:
### Fixes Order Url
Because Mailers don't receive the request context, url generation in mailer templates is less straightforward than in templates backed by a controller. `*_url` helpers are recommended so that Rails can construct an absolute path. But In our case we need the `polymorphic_url` which works best with our `location` methods. 

See: https://guides.rubyonrails.org/action_mailer_basics.html#generating-urls-in-action-mailer-views. 

### (Slightly) better marketplace factories
In addition to fixing the url, this PR includes a bite-sized effort to harden the associations in the Marketplace factories. While we pair debugged the mailer previews in the browser, @anaulin and I noticed inconsistencies in the data which prevented us from easily verifying the url was correct -- trying to visit the link as a guest or logged-in as the shopper was resulting in 404s (effectively access denied). We're convinced the URL is fixed, but there's still remaining work to be done to make it easy to live test the mailers locally. For example, notice the mismatched order ids in the email subject and body:

image

### DIFF
--- a/spec/factories/furniture/marketplace.rb
+++ b/spec/factories/furniture/marketplace.rb
@@ -78,7 +78,7 @@ FactoryBot.define do
   end
 
   factory :marketplace_product, class: "Marketplace::Product" do
-    name { Faker::TvShows::DrWho.specie }
+    name { Faker::TvShows::DrWho.unique.specie }
     price_cents { Random.rand(1_00..999_99) }
 
     marketplace
@@ -131,7 +131,7 @@ FactoryBot.define do
   end
 
   factory :marketplace_order, class: "Marketplace::Order" do
-    marketplace
+    marketplace { association(:marketplace, :ready_for_shopping) }
     shopper { association(:marketplace_shopper) }
 
     status { :paid }
@@ -141,7 +141,7 @@ FactoryBot.define do
         product_count { 1 }
       end
 
-      ordered_products { Array.new(product_count) { association(:marketplace_ordered_product) } }
+      ordered_products { Array.new(product_count) { association(:marketplace_ordered_product, marketplace: marketplace) } }
     end
 
     trait :with_taxed_products do
@@ -177,8 +177,12 @@ FactoryBot.define do
   end
 
   factory :marketplace_ordered_product, class: "Marketplace::OrderedProduct" do
-    product factory: :marketplace_product
-    order factory: :marketplace_order
+    transient do
+      marketplace { association(:marketplace) }
+    end
+
+    product { association(:marketplace_product, marketplace: marketplace) }
+    order { association(:marketplace_order, marketplace: marketplace) }
     quantity { 1 }
   end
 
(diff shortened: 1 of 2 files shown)


anaulin (approved):
👍🏼

spec/factories/furniture/marketplace.rb
```diff
@@ -141,7 +141,7 @@
         product_count { 1 }
       end
 
-      ordered_products { Array.new(product_count) { association(:marketplace_ordered_product) } }
+      ordered_products { Array.new(product_count) { association(:marketplace_ordered_product, marketplace: marketplace) } }
```
zspencer, line 144: Doh, I always forget that you need to pass them down the chain.

spec/factories/furniture/marketplace.rb
```diff
@@ -131,7 +131,7 @@
   end
 
   factory :marketplace_order, class: "Marketplace::Order" do
-    marketplace
+    marketplace { association(:marketplace, :ready_for_shopping) }
```
zspencer, line 134: Do all `Orders` need a `ready_for_shopping` markeplace? FactoryBot encourages making the baseline factory as minimal as possible, and use a trait if you need something more.

zspencer (commented):
Nice find and thank you for improving the Factories!!

spec/factories/furniture/marketplace.rb
rosschapman: Considering that principle then I suppose no. But following that we'd need a variation like `marketplace_order_ready_for_shopping`whatever for use with the mailer in https://github.com/zinc-collective/convene/blob/main/spec/mailers/previews/marketplace/order/received_mailer_preview.rb#L5. I'm thinking it may be a worthwhile exercise to write a new factory stack used by the mailer to see if we can fully simulate the correct data model and compare to the base factories currently there. Maybe it's a special that is only used for the mailer preview, then we don't have to worry so much about recreating the entire data model with the base factories?

zspencer: I don't think it makes sense to add a `marketplace_order_for_mailer_preview` or rewrite the factory stack. 

We have two pretty easy options for how to do this without promoting `Order`'s default factory to include a Marketplace that is :ready_for_shopping.


Option A) Instantiate the Marketplace where we need it
```rb
class Marketplace
  class Order
    class ReceivedMailerPreview < ActionMailer::Preview
      def notification
        marketplace = create(:marketplace, :ready_for_shopping)
        Order::ReceivedMailer.notification(FactoryBot.build(:marketplace_order, :full, marketplace:))
      end
    end
  end
end
```

Option B) Update the [`:full` trait to include a `Marketplace` generated with the `:ready_for_shopping` trait.](https://github.com/zinc-collective/convene/blob/main/spec/factories/furniture/marketplace.rb#L168) (which apparently means finding out why `the `marketplace :full` trait isn't :ready_for_shopping...)

Issue #2271, opened 2024-03-02
Gap missing on pages


rosschapman:
Some layouts are squished against the viewport edge. I found at least one example on the Entry create/edit pages:

![image](https://github.com/zinc-collective/convene/assets/5185/233b1200-f600-4caf-9299-ef7c16893f30)

Issue #2253, opened 2024-02-28
🐞 Hero image: it should have a maximum height, so that it doesn't take up most of the space on very wide screen sizes


anaulin:
Currently Convene's Hero Image expands to take up the entire available width (as you'd expect from a Hero Image), but also proportionally grows vertically. This results in the image taking up most of the vertical space when using a very wide viewport:
![image](https://github.com/zinc-collective/convene/assets/6729309/6d04ff7f-d568-4a7a-ad75-4e11ac268ea3)

Compare with how SquareSpace handles the same ultra-wide viewport:
![image](https://github.com/zinc-collective/convene/assets/6729309/5ada32ba-4205-43f6-8f05-95901a834604)

We should update the CSS on the Hero Image to make sure that it doesn't exceed a certain percentage of the screen height (or maybe a set number of pixels?).

anaulin:
@rosschapman this might be Relevant To Your Interestes

rosschapman:
My kingdom for a designer! But yes, that is pretty egregious 😅

zspencer:
I'm curious how to do this with `ActiveStorage` / `libvips`. My brain struggles to figure out how the constraints for that work.

PR #2281, opened 2024-03-07
🌸 `DesignLibrary`: Layout is responsible for.... Layout...

zspencer:
- https://github.com/zinc-collective/convene/issues/2271
- https://github.com/zinc-collective/convene/issues/2253

One of the things we've kept bumping into is that each page has to decide how wide it wants to be. That doesn't really seem sustainable or particularly necessary; so now the Layout takes care of:

- Providing a Max-Width on every page
- Ensuring the Layout is centered horizontally
- Ensuring there is appropriate margin from the top of the screen

### After
Screenshot 2024-03-06 at 6 08 02 PM

### Before
Screenshot 2024-03-06 at 6 07 51 PM


anaulin (approved):
🚢

PR #2282, opened 2024-03-07
🧹 Unfutz the broken linter

zspencer:
I accidentally pushed right to `main`. It was not great. Now our linter is sad.

This makes it less sad.


anaulin (approved):
Oops.

PR #2283, opened 2024-03-07
🌸 Make cards in the section navigation all have the same height

anaulin:

A small beautification tweak to make section navigation cards have a uniform height.

I decided that a max-height wasn't necessary, since we constrain the height of the image and the description length is capped at 300 characters.

I also took the opportunity to expand the seeds file with some hero images and room descriptions.

### Before
image

### After
image


zspencer (approved):
Thank you! This was something I had been wanting to deal with!!!

PR #2285, opened 2024-03-07
🌸 Ensure that section card arrow maintains its minimum width

anaulin:

This prevents the section card arrow from being squozen to an inch of its life by a long section title.

### Before
image

### After
image


zspencer (approved):
Love it!

PR #2286, opened 2024-03-07
🌸 Make arrow in section navigation look more like a little button

anaulin:

This is closer to the styling we had earlier, but without the bright orange, and without a change during hover.

### Before
(journal card shows hover state)
image

### After
(journal card shows hover state)
image


zspencer (approved):
Nice!

PR #2287, opened 2024-03-07
Uses `ActionText` for `MarkdownBlockText` content

rosschapman:


app/models/furniture.rb
```diff
@@ -22,6 +22,8 @@ class Furniture < ApplicationRecord
 
   delegate :attributes=, to: :gizmo, prefix: true
 
+  has_rich_text :action_content
```
zspencer, line 25: Why can't we have the `has_rich_text` on the `MarkdownTextBlock` itself?

zspencer (commented):
I presume there's something weird happening when we try and just make the change within the `MarkdownTextBlock`; so I'll pull it down and futz around with it.

I'm a bit surprised tho because `ActionText` doesn't have a field on the table it belongs to; so I would expect it to work fine with our polymorphic gizmos, so long as it's attached to something that inherits from `ActiveRecord` (which our `Furniture` objects do).

zspencer:
So, for those following along at-home/in-the-future: We stalled on this because we realized the underlying architecture of Gizmo's needed to be adjusted.

Instead of all Gizmo's living in the Gizmos table, each Gizmo will define it's own "Trunk" in the Data Forest, just as it defines it's own Trunk in the Object Graph. It will likely implement an interface called "Slottable" or some such, but will ultimately be it's own plane-jane ActiveRecord object; rather than a child of all the wild.

I'm going to leave this open, but it's "blocked" until we finish that refactoring.

PR #2301, opened 2024-03-14
Operation slot machine

zspencer:
@rosschapman ran into a weird bug when `ActionText`ing the `MarkdownTextBlock` where the action text couldn't save itself. Basically we were using nested attributes with a faux-relationship; which meant the actual `MarktdownTextBlock` was not getting saved.

@anaulin pointed out that this is an example of the single-table-inheritance being annoying, and that maybe we should finally deal with the cruft and blurry lines from our initial `Furniture` implementation.

So we started down the path of a refactor where we make a `Slot` responsible for where in a particular `Section` a `Gizmo` should live; and which `Gizmo` should live in that spot.

So we wrote a System Spec, and then we started implementing it and realized we had gone down the wrong pathyway, we want to create the `Gizmo` first and then add the `Slot`!

So we're going to reset and move forward with making a `Card` for each `Gizmo`!


anaulin:
If I'm understanding the notes in the PR description correctly, the underlying issue that @rosschapman was running into was the STI-iness of MarkdownBlock. This could be improved by migrating MarkdownBlock to be a Gizmo based on its own table, and not using the shared `furnitures` table.

There is an adjacent, but separate, issue around improving how we (re)order Gizmos. After Ross' introduction of the positioning gem, it feels like a good next step in this direction would be to remove the RankedModel gem altogether.

(Writing this down because Ross and I were just discussing next steps on this.)

zspencer:
Agreed, I think a larger refactor is probably at play (and oh god, has it been since March?! :x)

PR #2302, opened 2024-03-14
Improve look of section navigation cards when at least one section has an image

anaulin:
* https://github.com/zinc-collective/convene/issues/1988

This PR refines the look of the section navigation component:
* when at least one of the sections has an image an others don't, it inserts a colored image placeholder in all the cards that don't have an image
* when at least one of the sections has a description text, it makes sure that all the cards have the same description height, even if they don't have a description text

### Before
image
image
image

### After
image
image
image

### DIFF
--- a/app/furniture/section_navigation/section_navigations/_section_navigation.html.erb
+++ b/app/furniture/section_navigation/section_navigations/_section_navigation.html.erb
@@ -1,17 +1,1 @@
-<div class="grid grid-cols-1 gap-5 sm:gap-6 sm:grid-cols-2 lg:grid-cols-4">
-  <% policy_scope(section_navigation.rooms).each do |room| %>
-    <%= link_to polymorphic_path(room.location), id: dom_id(room, :link_to), class: "group no-underline" do %>
-      <%= render CardComponent.new(media: room.hero_image, classes: "flex flex-col justify-between h-full") do |card| %>
-        <p><%= room.description %></p>
-        <footer class="text-2xl text-justify font-semibold flex justify-between items-center">
-          <h3>
-            <%= room.name %>
-          </h3>
-          <%= render SvgComponent.new(classes: "h-7 w-7 min-w-7 p-1 rounded-full bg-slate-100") do %>
-            <path stroke-linecap="round" stroke-linejoin="round" d="M8.25 4.5l7.5 7.5-7.5 7.5" />
-          <% end %>
-        </footer>
-      <%- end %>
-    <%- end %>
-  <% end %>
-</div>
+<%= render SectionNavigation::SectionGridComponent.new(section_navigation) %>
--- a/app/furniture/section_navigation/section_grid_component.html.erb
+++ b/app/furniture/section_navigation/section_grid_component.html.erb
@@ -1,0 +1,24 @@
+<div class="grid grid-cols-1 gap-5 sm:gap-6 sm:grid-cols-2 lg:grid-cols-4">
+  <% @rooms.each do |room| %>
+    <%= link_to polymorphic_path(room.location), id: dom_id(room, :link_to), class: "group no-underline" do %>
+      <%= render CardComponent.new(media: room.hero_image, classes: "flex flex-col justify-between h-full p-0") do |card| %>
+        <% if image_placeholder?(room) %>
+          <% card.with_header(variant: :no_padding) do %>
+           <%=  image_placeholder_div %>
+          <% end %>
+        <% end %>
+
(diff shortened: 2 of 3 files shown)


app/furniture/section_navigation/section_navigations/_section_navigation.html.erb
```diff
@@ -1,7 +1,17 @@
 <div class="grid grid-cols-1 gap-5 sm:gap-6 sm:grid-cols-2 lg:grid-cols-4">
+  <% has_section_images = section_navigation.rooms.any? { |r| r.hero_image.present? } %>
```
rosschapman, line 2: I don't love seeing this type of logic in templates because I like to swing toward logic-less templates generally, but...I also wouldn't elevate it to the model since it's quite view specific. Of course a ViewComponent class would be a nice place for this. (Just thinking out loud.)

app/furniture/section_navigation/section_navigations/_section_navigation.html.erb
```diff
@@ -1,7 +1,17 @@
...
   <% policy_scope(section_navigation.rooms).each do |room| %>
     <%= link_to polymorphic_path(room.location), id: dom_id(room, :link_to), class: "group no-underline" do %>
       <%= render CardComponent.new(media: room.hero_image, classes: "flex flex-col justify-between h-full") do |card| %>
+        <% if has_section_images && room.hero_image.blank? %>
+          <% card.with_header do %>
+            <div class="text-6xl tracking-wider h-36 bg-amber-50 text-amber-600/25 p-4 rounded-md text-clip overflow-hidden flex">
```
rosschapman, line 8: Tailwind is...flexible. 🤣

rosschapman (approved):

rosschapman:
@anaulin nice use of clamping! 

tumblr_739e7c4118f7faa703eb056660fd6874_2b5c42d9_500

zspencer (approved):
I'm not sure I'm sold on the amber (it makes it look a bit like a caution sign, maybe grey?) and it feels like the text is a bit large... (text-4xl perhaps is the right size?)

Anyway, I think it's fine as-is; and we can polish it up later.

anaulin:
Hmm, yeah, I sort of don't love this. I was envisioning something more "pictorial" that could be auto-generated as stand-ins for missing pictures, but my result looks more like A VERY LARGE TITLE THAT SOMETIMES DOESN'T FIT.

@zspencer @rosschapman would something like this be better? Or do you like it more with the text?

This version randomizes the background color for a more placeholder-y look, and removes the text:
image

(The height of the bottom row needs to be adjusted to match the top row with the description, but that's a different issue.)

app/furniture/section_navigation/section_navigations/_section_navigation.html.erb
anaulin: The other "Rails-y" place for this is a method in a helper -- we could start a `SectionNavigationHelper` and add this as a predicate there.

My feeling is that this probably wants to become a `ViewComponent` sooner rather than later, and in that case this will make sense as a method in there.

I generally avoid having complex logic in a view, but I also think that most views in the real world will have some kind of logic in them, so I'm not religious about it. In this case I feel like this is simple enough and enough of an area that we are likely still iterate on that this is ok for now.

zspencer:
ooo, I like that quite a bit! The pastels are lovely!

rosschapman:
@anaulin looks good. Will passing the `no_padding` variant option to `card.with_header` remove the spacing so that colored area is flush against the card container?

anaulin:
@rosschapman probably, removing that margin is one of several remaining TO-DOs in this PR (see the PR description for my full list so far)

anaulin:
@zspencer @rosschapman I've gone back and cleaned up the implementation, and updated the screenshots in the PR description. Please have a final look!

rosschapman (approved):
🌈

anaulin:
This is in production now. I notice that on Piikup's entrance, some of the sections aren't showing an image, but are also not getting the placeholder image:
image

So this needs some revisiting, to make sure that the check to see if there is a hero image for a room is correct.

I also noticed:
* there a lot more rooms visible in the section navigation if you log in
  * it would be nice to add a little indicator to the cards to signify that a card/room won't be listed for logged-out people, maybe something like a little icon with a crossed-out eye and a tooltip; that way Space administrators could see more clearly at a glance what will be visible to regular users
* we really should order the sections alphabetically (until we let users set an order, at least), right now it feels kinda haphazard

PR #2320, opened 2024-03-25
🧹 `rake db:seed` is idempotent

zspencer:
It turns out, running `rails db:seed` multiple times kept resulting in errors or duplicious data.

This makes it a bit easier to run multiple times, which can be helpful when making a change that also wants to update the seeds

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -4,39 +4,47 @@
 OPERATOR_EMAIL = "operator@example.com"
 MEMBER_EMAIL = "member@example.com"
 
-FactoryBot.create(:person, :operator, name: "Ollie Operator", email: OPERATOR_EMAIL)
+Person.find_by(email: OPERATOR_EMAIL) ||
+  FactoryBot.create(:person, :operator, name: "Ollie Operator", email: OPERATOR_EMAIL)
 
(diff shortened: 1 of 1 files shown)


anaulin (approved):
I guess, if you feel like you need this.

I don't love that this makes the seeds more brittle and harder to verify that they work as intended, since  they are no longer guaranteed to recreate everything in a clean state. Personally, I prefer to operate under the philosophy that the seeds should be good enough to run `db:seed:replant` as often as necessary.

zspencer:
Hmm, I don't think I've used `db:seeds:replant`, mostly because the projects I've been on in the last half decade have relied (perhaps _too_) heavily on anonymized production data; rather than a seed file. 

I can start working `db:seeds:replant` into my workflow and see if that gets me out of wanting them to be idempotent.

rosschapman:
Interesting, I believe I agree with @anaulin here because I would expect to only run the primary seed on a clean database. However, I ok with the idea of having secondary seeds for setting up the app for a particular use case. Like `rails db:seed:space:marketplace:ready` that would act similar to factories in test setup. Although I don't feel like our app is complicated enough yet to warrant that. We find more bugs doing it all manually 😅

Issue #2266, opened 2024-03-01
Redesign Header and Footer for Spaces


rosschapman:
## Background

It's near universal for commercial websites to have static headers and footers for providing basic site navigation or quick access to information and other resources. As noted by @anaulin in https://github.com/zinc-collective/convene/issues/2250#issuecomment-1970251259, Convene used to support a default header but this was removed because it wasn't considered visually useful when viewing  Marketplace screens.

It makes sense that displaying the website's "chrome" (some combination of header and/or footer) is not appropriate for every screen. But on more static informational pages it can be necessary visual brand statement and for site navigation. For example, when visiting  Moment Park pages that we have migrated into Convene, the typical header or footer site navigation (that us users of the web have become used to) feels obviously missing. See: https://convene.zinc.coop/spaces/moment-park/rooms/iphone-apps. Our only recourse as visitors is the single footer breadcrumb, which forces me to return to the home page before visiting another section of the website. 

### Space admins should be able to manage sitewide headers and footers in the following ways: 
- Enable and disable a sitewide Header
- Enable and disable a sitewide Footer
- Selectively disable the header or footer on a Section
- Manage what content goes in the header and footer (gizmos?)
- Pick the background color and text color of the header and footer

### What is a Header or a Footer? 

- Data:
  - `name` (string)
  - `content` ([Gizmo|ordered])
  - `sitewide_enabled` (boolean)
  - `excluded_from` ([`Section`])
    - Inversely we could add `show_header` and `show_footer` attributes to `Section`
- ❓ Is it a special kind of `Section`?
  - `access` attr could be utilized for enabled/disabled
  - Seems risky since they really serve different purposes. Header/Footer are more branding/navigation focused vs distinct _content areas_ of a `Space`. 

### Possible things Space admins could put in a Header or Footer
- Section Navigation
- Breadcrumb menu
- Markdown
- HTML
- `LinkList` (to add, discussed in https://github.com/zinc-collective/convene/issues/2036)

### Other

- The Convene branding footer should be visually distinct from the Space-enabled footer
  - Maybe a slightly darker orange?

zspencer:
> Convene used to support a default header but this was removed because it wasn't considered visually useful when viewing Marketplace screens.

It's more that the Header we had was an eye-sore; and the time invested to make it less of an eye-sore was less effective than removing it entirely 😅

rosschapman:
First spike up: https://github.com/zinc-collective/convene/pull/2338

[...]

PR #2338, opened 2024-04-02
Adds basic Space theming configuration: `show_header`, `header_bg_color`, `header_txt_color`

rosschapman:
This changeset enables Space admins to: 
- [x]  Enable and disable add a sitewide _header_ that will include Space name and Section Navigation
- [x]  Customize the _header_ background color
- [x]  Customize the _header_ text color

Additional implementation details and considerations: 
- The header is a common responsive design that is stacked vertically for small screens and expands horizontally for larger screens

https://github.com/zinc-collective/convene/assets/5185/7b063ee8-5571-42fd-b79f-55695b03c477

---
Reference 🎟️ : https://github.com/zinc-collective/convene/issues/2266


rosschapman:
@zinc-collective/convene-maintainers @zinc-collective/convene-contributors 

Here's a spike to add some Site Branding capability to Convene. I'm aiming for the simplest implementation at the start here: 

- Header is on/off, will display space-wide, is not configurable, only shows space name and section links with default styling (to come)
- Footer is customizable as action text
- Brand color only applies to header and footer

Does this feel like a reasonable place to start? 

NOTE: Spike is deliberately missing db bits.

app/views/spaces/_header.html.erb
```diff
@@ -0,0 +1,7 @@
+<div style='background-color: <%= current_space.brand_color %> '>
+  <h1><%= current_space.name %></h1>
+  <% policy_scope(current_space.rooms).each do |room| %>
+    <%= link_to room.name, polymorphic_path(room.location), id: dom_id(room, :link_to), class: "group no-underline" %>
+  <% end %>
+</div>
+</div>
```
rosschapman, line 7: This would be marked up and styled like the Moment Park header as a sensible default.

app/models/space.rb
```diff
@@ -45,6 +45,9 @@ class Space < ApplicationRecord
   # @returns {ActiveRecord::Relation<Utilities>}
   has_many :utilities, inverse_of: :space, dependent: :destroy
 
+  attribute :brand_header, :boolean
+  attribute :brand_color, :string
```
anaulin, line 49: Comments from group code review: it might be nice to use more specific names here, like `header_bg_color`. Feels like "brand_color" would imply a lot more than what we currently plan to do with this color.

app/views/spaces/_branding.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="mt-3 gap-y-3">
+  <%= render CardComponent.new do %>
+    <h3>Branding</h3>
```
anaulin, line 3: Maybe "Colors" or "Styling", instead of "Branding"?

"Branding" feels very oriented towards more commercial users, and also not as specific as the functionality that this includes right now.

app/components/brand_header_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<div style='background-color: <%= @space.brand_header_color %> ' class="flex flex-col md:flex-row flex-wrap justify-between sm:px-5 py-5 items-center">
```
zspencer, line 1: It would be nice if we used CSS for this by doing something like:

```erb
<!-- This lives in layout.html.erb -->
<style>
.SpaceHeader {
  background-color: <%= @space.header_background_color %>
}
</style>

<!-- this stays here -->
<div class="SpaceHeader">
...
</div>
```

zspencer (approved):
I think this is good-enough as-is, but I left some thoughts on how to make it even bettah.

app/views/spaces/_branding.html.erb
zspencer: lol maybe "Vibes"

app/components/brand_header_component.html.erb
rosschapman: I like _class-based_ css way better than Tailwind style utility classes, but I think mixing those patterns will get confusing.

app/views/layouts/application.html.erb
```diff
@@ -1,20 +1,17 @@
 <!DOCTYPE html>
 <html>
   <%= render 'layouts/head' %>
-
   <body>
     <%= render 'layouts/header' %>
-
+    <%- if current_space&.brand_header %>
```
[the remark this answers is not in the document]
rosschapman: Ah yeah, I had a partial in there before I made the component!

Issue #2340, opened 2024-04-04
🥡 `Marketplace::MenuGroup`


rosschapman:
### Use Cases

- [x] Setting `Marketplace::Product`'s `Marketplace::MenuGroup`
- [ ] https://github.com/zinc-collective/convene/issues/2378

### Execution Plan

April has pointed out - rightly so - that Product "menus" on Convene look "haphazard" because we don't allow sorting or categorization. Fortunately, @zspencer's implementation of #2189 laid a strong foundation for improving organization of product pages. 

**✅ DECISION:** As a first step, let's implement the capability for Marketplace administrators to group Products under  _sections_ (err...menus, aisles, taxonomy pending...). 

During spirited text-based Discord task planning, there was first some temptation to consider a separate model for _Product Section_ with a _many_ association to tags. This way Marketplace admins could manage these sections via a typical CRUD-y workflow, similar to managing other entities like Tags, Tax Rates, or Products. However, @anaulin encouraged us to consider enhancing `Tags` themselves without introducing any new entity relationships. Because tags and taxonomies are user-generated the use cases across customers will inherently evolve to be nuanced, various, and non-linear. In other words, hard to predict from our vantage point.  Ultimately we hoist a strong risk if we design a rigid relationship structure into the data model from the start.

**✅ DECISION:** After some deliberation, @anaulin proposed a (most reasonable) minimum set of requirements we need to deliver to support grouping of tags in sections on the page. Let's do these, show them to April, and start the feedback loop quickly.

**Phase 1**
- [x] A person can mark a tag as a "section"
- [x] We can retrieve a list of products that belong to a "section"
- [x] We can retrieve a list of all the "section" tags

**Phase 2**
- [x] A person can (re)order sections by drag and drop

**Further development tasks/considerations:** 
- [ ] `Tag` will be extended with a `boolean` field like `is_section`
	- [ ] Show a checkbox in the Add/Edit Tag UIs
- [ ] Tags should be sortable (we use `ranked-model`)
- [ ] Default section sort is by alpha
- [ ] 

**🔮 Considerations for the future:** 
- Adding the concept of a `TagGroup` or `TagCollection` would be necessary if we have a reason to display all the tags for a single  group, or inversely all the groups of a single tag. 
	- This could be a achieve with polymorphism (`TagGroupable`) although it won't be necessary if we only have a Tas and TagGroups
	- In this data model there's the possibility of needing to restrict certain tags (or all?) to be belong to one group

[...]

PR #2361, opened 2024-04-18
Space admins can create "Menu/Product Groups" and reorder them for display on their marketplace

rosschapman:
Foundational work for https://github.com/zinc-collective/convene/issues/2340. 

✋🏻✋🏻✋🏻 Let's get these PRs merged first to reduce cognitive load for reviewers: 
- https://github.com/zinc-collective/convene/pull/2390
- https://github.com/zinc-collective/convene/pull/2391

---

## Tasks

- [x] Add `is_section` attribute to `Marketplace::Tag` (db, model, controller, etc..)
- [x] Add "Section" label to Tags view
- [x] Add `Tag` editing (view, controller, etc...)
- [x] Add sections to Products view sorted by rank prop (`slot`)
- [x] Add tag sorting controller action
- [x] Add tag sorting UI (Stimulus Sortable)
- [x] Add system spec for marking group tag
- [x] Add system spec for ordering of group tags
- [x] 🌸 Update product tag display to pills

### Follow-up tasks (2/N)
- [ ] 🌸 Update tag selection UI to token input


rosschapman:
@zinc-collective/convene-maintainers I'm going to pause dev on this spike for the moment for a gut-check from the committee. 🙏🏻

app/furniture/marketplace/tags_controller.rb
```diff
@@ -12,6 +12,22 @@ def new
...
+      authorize(mtag)
+    end
+
+    def update
+      if authorize(mtag).update(mtag_params)
+        if request.xhr?
+          render json: {}, status: :ok
+        else
+          redirect_to marketplace.location(child: :tags)
+        end
```
rosschapman, line 25: Should I be using `respond_to |format|...` here? I tried by but for some reason after drag/drop I was still getting into the html block and it would try to redirect to `../<tag_id>/tags` incorrectly.

```ruby
# Example that redirects even for the ajax call 🤔
respond_to do |format|
	format.json { head :ok}
	format.html do 
		redirect_to marketplace.location(child: :tags)
	end
end 
```

db/migrate/20240422172820_add_position_to_marketplace_tag.rb
```diff
@@ -0,0 +1,9 @@
+class AddPositionToMarketplaceTag < ActiveRecord::Migration[7.1]
+  def change
+    add_column :marketplace_tags, :position, :integer
+
+    # Marketplace::Tag.order(:updated_at).each.with_index(1) do |tag, index|
+    #   tag.update_column :position, index
+    # end
```
rosschapman, line 7: I don't think this is necessary but we could.

app/furniture/marketplace/tags_controller.rb
zspencer: It depends on how the drag/drop request gets sent, I can't remember exactly how rails decides something is a "JSON" format, but I believe it's an intersection of xhr + content-type; which may not be being set by default on  `@stimulus-components/sortable`

db/migrate/20240422172820_add_position_to_marketplace_tag.rb
zspencer: This makes me think that a `Group` wants to be it's own table, rather than a flag on `marketplace_tags`, but also I don't want to eat that complexity right now. I think ordering by alpha is fine.

app/furniture/marketplace/tags/_form.html.erb
```diff
@@ -1,6 +1,10 @@
 <%= form_with(model: tag.location) do |form| %>
-
-<%= render "text_field", attribute: :label, form: form %>
-
-<%= form.submit %>
+  <div class="flex flex-col gap-5">
+    <div>
+      <%= render "text_field", attribute: :label, form: form %>
+      <%= form.label :is_group, "Is this a menu section?" %>
```
zspencer, line 5: In general, if there is drift between what we're calling things to a user and how we are naming them in the database, I prefer to start with them aligned.

zspencer (commented):
I think this is a great start! I left some comments (I had actually read this before I left last weekend but completely forgot to click the submit! Apologies!)

Gemfile
```diff
@@ -58,6 +58,8 @@ gem "active_record_extended", "~> 3.2"
 gem "activerecord-postgres_enum", "~> 2.0"
 # Support for models with "slots" or "positions"
 gem "ranked-model", "~> 0.4.9"
+gem "acts_as_list"
```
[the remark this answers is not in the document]
rosschapman: Yeah let's do that. `acts_as_list` was used in the examples for setting up Sortable but it's not a big deal to change since the api is the same.

app/furniture/marketplace/marketplace.rb
```diff
@@ -136,5 +136,12 @@ def square_order_notifications_enabled?
     def default_delivery_area
       (delivery_areas.unarchived.size == 1) ? delivery_areas.unarchived.first : nil
     end
+
+    # TODO: There must be a more Rails-y way to do this?
+    def products_with_no_group_tags
+      products.select do |p|
+        p.tags.with_group.empty?
+      end
+    end
```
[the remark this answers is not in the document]
rosschapman: Per Zoom sidebar with @zspencer I think we'll just plop this sql statement right into the code and avoid extra lookups and memory usage of the application loops.

app/furniture/marketplace/tags/_form.html.erb
rosschapman: Ah, I missed this doing s/section/group.

app/furniture/marketplace/tags_controller.rb
rosschapman: I'll double check the headers. I definitely don't mind `request.xhr?`.

db/migrate/20240422172820_add_position_to_marketplace_tag.rb
rosschapman: _:nod:_ I started going down the path of writing a `Group` model and I just couldn't come up with a strong enough justification at this time since the goal of this changeset feels achievable without it.

app/furniture/marketplace/menu_component.html.erb
```diff
@@ -1,5 +1,16 @@
...
+  <%- unless tag.products.empty? %>
+    <h1><%= tag.label %></h1>
+    <div class="grid lg:grid-cols-3 gap-3">
+      <%- tag.products.with_all_rich_text.unarchived.each do |product| %>
+        <%= render Marketplace::Menu::ProductComponent.new(product:, cart:) %>
+      <%- end %>
+    </div>
+  <%- end %>
+<%- end %>
+<h1>Other</h1>
```
rosschapman, line 11:
```suggestion
<h1>Everything</h1>
```

app/furniture/marketplace/tags_controller.rb
rosschapman: Ah, looks like Stimulus Sortable is actually sending an ajax request but with `Content-Type` as `multipart/form-data`.

app/furniture/marketplace/tags/_form.html.erb
rosschapman, line 5:
```suggestion
      <%= form.label :is_group, "Is this a menu group?" %>
```

app/furniture/marketplace/tags_controller.rb
```diff
@@ -12,6 +12,24 @@ def new
...
+      authorize(mtag)
+    end
+
+    def update
+      if authorize(mtag).update(mtag_params)
+        # POST requests to reorder tags (Stimulus Sortable) are sent with an
+        # html content-type so they cannot be handled by a JS responder.
+        if request.xhr?
+          render json: {}, status: :ok
+        else
```
zspencer, line 25: Do we need the if branch then?

app/furniture/marketplace/tags/index.html.erb
```diff
@@ -1,16 +1,47 @@
...
+  <h1>Tags</h1>
+  <p>Add tags to label your products as vegan, discounted, etc...</p>
+  <%- bazaar.tags.without_group.each do |tag| %>
+    <%-
+      # We must assign the marketplace object explicitly in the current scope
+      # because of the funky way we relate Tags to the Marketplace through a Bazaar.
+      # Yet our routing and navigation depends on a `marketplace` object being present
+      # on a Tag.
+      tag.marketplace = marketplace
+    %>
```
zspencer, line 40: This indicates to me that we should associate Tags to a Marketplace directly. I still think there's benefit for Marketplace::Tags to have some kind of relationship to the Tags in other Marketplaces in the same Space; but I don't think preserving that is that useful here.

app/furniture/marketplace/marketplace.rb
```diff
@@ -136,5 +136,44 @@ def square_order_notifications_enabled?
...
+            and mp.discarded_at is null
+        group by
+            mp.id
+        having not
+            't' = any(array_agg(mt.is_group));
+      SQL
+
+      # Step 3: Merge the above with all other products that are missing tags
+      # which won't be captured by the above query
+      with_no_group_tags + products.unarchived.where.missing(:tags)
```
zspencer, line 176: Oof this is a doozy but I'm glad y'all figured it out.

zspencer (approved):
This is great! Ship it!

app/furniture/marketplace/marketplace.rb
anaulin: This should have a spec. And it probably should be a scope on `Marketplace::Product`, instead of a class method on `Marketplace`.

app/furniture/marketplace/tag.rb
```diff
@@ -10,5 +10,10 @@ class Tag < Record
 
     attr_accessor :marketplace
     location(parent: :marketplace)
+
+    positioned
```
anaulin, line 14: Shouldn't this positioning be scoped to the `bazaar`? If I'm understanding correctly, this makes the position global across all models in the database, so there can only ever be a tag in position 1 across all of Convene bazaars and marketplaces.

app/furniture/marketplace/tag.rb
```diff
@@ -10,5 +10,10 @@ class Tag < Record
...
+
+    scope :without_group, -> { where(is_group: false) }
+    scope :ordered_tag_groups, -> { where(is_group: true).merge(order(position: :asc, updated_at: :desc)) }
```
anaulin, line 17: I don't think it is possible for two tags to have the same position, at least not within whichever we decide is the reasonable "scope". In which case it is not relevant to try to have a second ordering criterion.

app/furniture/marketplace/tags/index.html.erb
```diff
@@ -1,16 +1,47 @@
...
+  <h1>Product Groups</h1>
+  <p>Drag and drop to change the order in which groups will display on your Marketplace</p>
+  <ul data-controller="sortable" data-sortable-animation-value="150" data-sortable-resource-name-value="tag" class="flex flex-col gap-2 p-0">
+    <%- bazaar.tags.ordered_tag_groups.each do |tag| %>
+      <%-
+      # We must assign the marketplace object explicitly in the current scope
+      # because of the funky way we relate Tags to the Marketplace through a Bazaar.
+      # Yet our routing and navigation depends on a `marketplace` object being present
+      # on a Tag.
+      tag.marketplace = marketplace
```
anaulin, line 12: I don't understand this. I see that a Tag has an accessor for a Marketplace, but doesn't seem to have a direct association to one, and yet has a `location(parent: :marketplace)`, which is probably causing the need for this weird assignment here.

Seems like we are missing a `/bazaar` route namespace, and it is leading to this funkiness. If that's what we need, can't we just... add it?

anaulin:
FYI, I've added a couple of commits to try to tighten up those tag and product scopes, and also added specs to document what the intended behavior is and verify that the implementation matches it.

Since the original behavior wasn't specc'ed or explained, it is possible that my changes introduce some unintended behavior change, so please @rosschapman test it on your side to make sure it works as intended.

app/furniture/marketplace/menu_component.html.erb
```diff
@@ -1,5 +1,16 @@
...
+  <%- unless tag.products.empty? %>
+    <h1><%= tag.label %></h1>
+    <div class="grid lg:grid-cols-3 gap-3">
+      <%- tag.products.with_all_rich_text.unarchived.each do |product| %>
+        <%= render Marketplace::Menu::ProductComponent.new(product:, cart:) %>
+      <%- end %>
+    </div>
+  <%- end %>
+<%- end %>
+<h1>Everything</h1>
```
anaulin, line 11: I know @zspencer suggested for this to be "Everything", but that doesn't match what this section contains, which is "Everything that wasn't already listed above".

It could be "Everything else". But also I think "Other" was just fine for now.

anaulin:
As part of the follow-up UI polish work, I think we should remove the "group" tags from the product cards, otherwise it looks redundant:
image

app/furniture/marketplace/tag.rb
```diff
@@ -10,5 +10,13 @@ class Tag < Record
 
     attr_accessor :marketplace
     location(parent: :marketplace)
+
+    positioned
+
```
anaulin, line 15: The only merge-blocking issue I see in this PR right now is sorting out this `positioned` scope.

anaulin (changes requested):

app/furniture/marketplace/product.rb
```diff
@@ -41,6 +41,9 @@ class Product < Record
 
     before_commit :standardize_attachment_name
 
+    scope :with_group_tag, -> { joins(:tags).merge(Tag.group_tag) }
+    scope :without_group_tag, -> { where.not(id: with_group_tag) }
```
rosschapman, line 45: Cool composition of scopes to replace the raw query!

rosschapman: Ha, I definitely overthought the original query, because this one that AR produces with a subquery is concise: 

```sql
select
	"marketplace_products".*
from
	"marketplace_products"
where
	"marketplace_products"."id" not IN(
		select
			"marketplace_products"."id" from "marketplace_products"
			inner join "marketplace_product_tags" on "marketplace_product_tags"."product_id" = "marketplace_products"."id"
			inner join "marketplace_tags" on "marketplace_tags"."id" = "marketplace_product_tags"."tag_id"
		where
			"marketplace_tags"."is_group" = true)
limit 11;
```

rosschapman:
@anaulin [said](https://github.com/zinc-collective/convene/pull/2361#discussion_r1597321899):
> The only merge-blocking issue I see in this PR right now is sorting out this positioned scope.

Yeah, I was punting on that for the moment while we got the code in good shape. I think scoping to the bazaar is...fine. This means that every vendor menu April builds can/will share the same product tag group list and order. Let's see how that amount of restriction plays out?

I committed the [7906242](https://github.com/zinc-collective/convene/pull/2361/commits/7906242451ce7be0254fa5e5b9f06c626b87cf30) for bazaar scoping.

rosschapman:
@anaulin @zspencer Now that I've read through all the comments it sounds like creating an association between Product Tags and Marketplace makes more sense currently, considering the way we we've designed the user experience. We've put tags under the marketplace route space anyway. (The ick of the routing hack `tag.marketplace = marketplace` is obvious). As Zee muses,  right now the use case(es) for tag sharing across the bazaar don't feel strong enough to contort the code to serve tag administration at the marketplace level:

>  I still think there's benefit for Marketplace::Tags to have some kind of relationship to the Tags in other Marketplaces in the same Space; but I don't think preserving that is that useful here

I also suspect April (and other users) would expect menu control at the individual marketplace level before considering how to manage them across the entire Bazaar.

How about I rewind and establish the Tag<>Marketplace association first, then merge that into this implementation and we can see how the routing and position scoping becomes more robust.

app/furniture/marketplace/tags/index.html.erb
zspencer: Yea, I think the path @rosschapman suggested to ditch the `bazaar` relationship on tags is the right one; which I think was your suggestion earlier that I stubbornly refused.

app/furniture/marketplace/tags/index.html.erb
```diff
@@ -1,16 +1,40 @@
...
 <%= render CardComponent.new do |card| %>
-  <%- marketplace.tags.each do |tag| %>
-    <div id="<%= dom_id(tag)%>" class="flex flex-row gap-3">
-      <span class="flex-grow">
-        <%= tag.label %>
-      </span>
+  <div class="flex flex-col gap-4">
+    <div>
+      <h1>Menu Groups</h1>
+      <%- if marketplace.tags.group_tag.empty? %>
```
anaulin, line 6: I like how in the "Tags" section there is a helpful message ("Adding tags to your products like "vegan" or "discounted" will help shoppers more easily find what they want"). Let's add something similar here, e.g: "Use product groups like "featured" or "produce" to display some of your products together, and to control the order in which products are displayed." (Or some better wording.)

db/seeds.rb
```diff
@@ -30,7 +32,15 @@
   description: "A marvelous marketplace for magic merchandise.",
   hero_image: FactoryBot.create(:media)
 )
-FactoryBot.create(:marketplace, :full, room: marketplace_section)
+marketplace = FactoryBot.create(:marketplace, :ready_for_shopping, product_quantity: 16, room: marketplace_section)
```
anaulin, line 35: I think this code exists in `main`, and should not be appearing as a diff in this PR.

anaulin (approved):
Code-wise this looks good, thank you! I'm fine with this being merged once you can get the build green and the rebasing on latest `main` sorted out.

My main comment is that I think we should improve the naming and description of "tag groups" and "tags" in user-facing UIs and strings. The goal is to make this as clear to users as possible. But we can make these changes in follow-up PRs.

Calling it a "menu group" doesn't make sense to me (it is not a group of menus). I think some better names could be "product group" or "menu" or maybe "menu tag".

We could either differentiate "tags" from "menus" more clearly in the management UI (maybe even having two separate edit/create places for the two different kinds), or we could instead lean into it and call them "menu tags", and describe them as "special tags that allow you to group a set of your products and control the order in which they are displayed".

rosschapman:
@anaulin Agree about cleaning up the copy. April might even have some insight on what language feels more natural from a user perspective.

anaulin:
Asking April for input is a good idea!

db/seeds.rb
rosschapman: Ah, it's not actually. You made these changes in this branch with https://github.com/zinc-collective/convene/pull/2361/commits/cd5cd4b1e849fb37bf0384691b97e8f2cf2a4b37. ☺️

rosschapman: Ah, it's not actually. You made these changes in this branch with https://github.com/zinc-collective/convene/commit/cd5cd4b1e849fb37bf0384691b97e8f2cf2a4b37. ☺️

anaulin:
> Ah, it's not actually. You made these changes in this branch with https://github.com/zinc-collective/convene/commit/cd5cd4b1e849fb37bf0384691b97e8f2cf2a4b37

Oh, sorry, I forgot that was in this PR! 🤦🏼‍♀️

PR #2390, opened 2024-05-06
Adds positioning Gem 

rosschapman:
Part of work for https://github.com/zinc-collective/convene/pull/2361. Should be merged first.


zspencer (approved):
Thanks! Hopefully we can do a follow on to rip out `ranked-model` and not have both gems but for now this lets us move forward with the sortable categories work so ship it.

rosschapman:
> Thanks! Hopefully we can do a follow on to rip out ranked-model and not have both gems but for now this lets us move forward with the sortable categories work so ship it.

Yeah I'll make an issue for this.

PR #2391, opened 2024-05-06
Adds `is_group` and `position` column to Marketplace Tags

rosschapman:
Part of work for https://github.com/zinc-collective/convene/pull/2361. Should be merged first!


zspencer (approved):
go for it; we can always promote a `Marketplace::MenuGroup` later.

PR #2400, opened 2024-05-10
Remove obsolete docker-compose top-level `version` keyword

anaulin:
The top-level version keyword in the docker-compose.yml spec has been deprecated, and using now triggers a warning. See: https://github.com/compose-spec/compose-spec/blob/master/04-version-and-name.md

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3.1'
-
 services:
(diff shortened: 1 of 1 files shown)


rosschapman (approved):

zspencer (approved):
Nice!

PR #2408, opened 2024-05-15
Removes Bazaar association from Tags

rosschapman:
While working on https://github.com/zinc-collective/convene/pull/2361 it became clear to reviewers (and yours truly, the author) that the association of Tags to the Bazaar didn't quite fit the emerging use cases, particularly "menu groups" and interfaces for managing tags within the Marketplace scope. 

This changeset updates the association on tags to a Marketplace!


db/migrate/20240515215452_add_marketplace_id_to_tags.rb
```diff
@@ -0,0 +1,7 @@
+class AddMarketplaceIdToTags < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :marketplace_tags, :marketplace, index: {unique: true, algorithm: :concurrently}
```
anaulin, line 5: We probably don't want to have any empty foreign keys in this table, so it would be good to add that constraint as you add the reference.

This will also force you to consider how do you want to deal with any tag records that might already exist in the database (assign them to a random marketplace from the bazaar they belonged to? delete them? something else?).

spec/furniture/marketplace/tag_spec.rb
```diff
@@ -1,5 +1,7 @@
 require "rails_helper"
 
 RSpec.describe Marketplace::Tag, type: :model do
-  it { is_expected.to validate_uniqueness_of(:label).case_insensitive.scoped_to(:bazaar_id) }
+  subject { build(:marketplace_tag) }
```
rosschapman, line 4: This test started complaining because tags were being created without a marketplace_id foreign key. Not sure why it worked for bazaar before?

anaulin: The relationship to `bazaar` did not have a non-null constraint, but the new `marketplace` relationship does.

rosschapman: Ah yes, of course. Free floating tags.

anaulin (approved):

Issue #2426, opened 2024-05-27
Marketplace Menus: 🪩🥂🍾✨ and next steps


rosschapman:
## 🚀😀🥳👌🏽  Menus are a thing now!
Last week we released the first iteration of improving the ordering experience on the Marketplaces. This effort was initially kicked off with a request from our client Piikup for more control over how Marketplace Products were displayed, including product order and gouping: https://github.com/zinc-collective/convene/issues/2340. Effectively this meant allowing configuration of *menus* -- a common behavior of retail web applications. 

Here's all the bodacious, beautiful code from this initial release:

1. https://github.com/zinc-collective/convene/pull/2390
2. https://github.com/zinc-collective/convene/pull/2391
3. https://github.com/zinc-collective/convene/pull/2408
4. https://github.com/zinc-collective/convene/pull/2361

## ✨🤩🌿 Menus twisted our arm to refactor
Beyond enabling better segregation of products into menu "sections" (eh, terminology WIP) (ie sub-menus, or topical groupings), this work provided additional utility as a vessel for interrogations of our domain model and marketplace management UI. Like, the relationship of Tags to a Bazaar, and how tag and menu management can be more humanistic and less an expression of the datastore.

## ✨🌸🐞🛍️🥡 Coming up
There's more work to do. Here's a sketch of what's obviously missing and what questions still need answering.

### 🤗👉 Tasks ready to go 
- [x] ✨ Marketplace admins should be able to _delete_ tags (Done: https://github.com/zinc-collective/convene/pull/2431)
- [x] 🌸 _Menu tags_ are redundantly displayed on the product. See how the the `featured` tags is shown on the Banana product when already under the Featured section. (Done: https://github.com/zinc-collective/convene/pull/2431)
	![image](https://github.com/zinc-collective/convene/assets/5185/17da052a-5b5d-4a27-a5ba-61fe5164b010)
- [x] ✨ Let's add a sensible default sort order for products. Alphabetical should suffice. (Done: https://github.com/zinc-collective/convene/pull/2431)
- [x] 🌸 Consistent tag design. Right now tags are displayed as plain text and/or pills across admin and end-consumer interfaces. Should probably institute a common "pill" design (Done: https://github.com/zinc-collective/convene/pull/2431)

### 💭🤔🤨🧐🧠 Tasks that need more thought/design

- ❓Better interfaces for managing tags and menus, and possible better integration with Product admin
	- Clarify the distinct interfaces for managing menus and tags
[...]

PR #2431, opened 2024-06-01
Marketplace Tags enhancements

rosschapman:
A handful of useful, low-bar follow-ups to https://github.com/zinc-collective/convene/pull/2361 that was scoped out in https://github.com/zinc-collective/convene/issues/2426. 

- [x] Adds tag deletion
- [x] Renames `group*` -> `menu*`
	💬 Did my best to migrate this language in the code and for end users. Feedback very welcome.
- [x] Adds `sort_alpha` scope for Products
- [x] Adds new `Display` and `Admin` components for tags
	- [x] Adds product tag count to Tag admin, with pluralized string
image
image


app/furniture/marketplace/locales/en.yml
```diff
@@ -70,6 +70,7 @@ en:
         success: "Notification Method '%{contact_location}' Saved!"
       destroy:
         success: "Notification Method '%{contact_location}' Removed!"
+        link_to: "Remove Notification Method '%{contact_location}'"
```
rosschapman, line 73: Hitting a different bug when running testing `spec/furniture/marketplace/marketplace_tags_system_spec.rb` surfaced this missing translation. 🤷‍♀️

app/components/marketplace/tag/admin_component/admin_component.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= render CardComponent.new(dom_id: dom_id(tag)) do |card| %>
+  <div class="flex flex-col gap-2">
+    <h3><%= tag.label %></h3>
+    <span class="text-xs">Assigned to <%= assigned_product_count %> <%= "Product".pluralize(assigned_product_count) %></span>
```
rosschapman, line 4: Not sure about "assigned to" 🤔

app/components/marketplace/tag/admin_component/admin_component.yml
rosschapman, line 1: Do we care about this boilerplate being here?

app/furniture/marketplace/menu_component.html.erb
```diff
@@ -1,16 +1,16 @@
-<%- marketplace.tags.group_tag.by_position.each do |tag| %>
+<%- marketplace.tags.menu_tag.by_position.each do |tag| %>
   <%- unless tag.products.empty? %>
     <h1><%= tag.label %></h1>
     <div class="grid lg:grid-cols-3 gap-3">
-      <%- tag.products.with_all_rich_text.unarchived.each do |product| %>
+      <%- tag.products.with_all_rich_text.unarchived.sort_alpha.each do |product| %>
```
rosschapman, line 5: Is this type of scope chaining typical in templates? Would you put a helper on the ruby component to reduce logic in the template?

app/components/marketplace/tag/admin_component/admin_component.yml
zspencer: Is this for i18n? Or ? If we don't put anything in it I would rather we left it out.

app/furniture/marketplace/locales/en.yml
zspencer: Probably worth submitting independently so it can be merged without being hung up on feedback for this feature.

app/furniture/marketplace/locales/en.yml
```diff
@@ -142,6 +143,8 @@ en:
       new:
         link_to: "Add Product Tag"
       edit:
-        link_to: "Edit Product Tag"
+        link_to: "Edit Product Tag '%{label}'"
```
zspencer, line 146: 👍👍👍

app/furniture/marketplace/locales/en.yml
```diff
@@ -142,6 +143,8 @@ en:
...
       update:
         success: "Product Tag saved!"
+      destroy:
+        link_to: "Remove Product Tag '%{label}'"
```
zspencer, line 150: 👍👍👍

app/furniture/marketplace/menu_component.html.erb
zspencer: Well, it's not too unusual, but it does make for an N+1 query. It may be worth pulling this into the query on line 1; which may be expressed in the view directly, as a helper on `MenuComponent`, or a scope on `Marketplace`.

There's also an argument to be made for not doing *any* queries from within a `Component` or view; instead performing them at the controller layer.

app/furniture/marketplace/tags_controller.rb
```diff
@@ -42,6 +42,20 @@ def index
...
+      authorize(mtag).destroy
+
+      respond_to do |format|
+        format.turbo_stream do
+          if mtag.destroyed?
+            render turbo_stream: turbo_stream.remove(mtag)
+          else
+            render turbo_stream: turbo_stream.replace(mtag)
+          end
+        end
```
zspencer, line 55: Do we need a turbo stream? I thought we had gotten to the point where we can just render plain html and Turbo will do a diff-and-swap within the marketplace?

zspencer (approved):
This looks like an overall improvement from an ixd perspective, and tidies up some oddities; I'm down.

app/components/marketplace/tag/admin_component/admin_component.yml
rosschapman: Yeah it's auto-generated by `rails generate component ...` for i18n. I can remove them.

app/furniture/marketplace/tags_controller.rb
rosschapman: Ah right. I'll double check.

app/components/marketplace/tag/admin_component/admin_component.html.erb
anaulin: I think just "N products" or "on N products" might be clear enough.

app/components/marketplace/tag/admin_component/admin_component.yml
anaulin: Yeah, no sense in checking in blank lines. (Also, I had no idea this .yml file for a component was a thing, #TIL.)

app/components/marketplace/tag/display_component/display_component.yml
anaulin, line 1: Can be deleted.

app/furniture/marketplace/menu_component.html.erb
rosschapman: What's the best way to analyze whether this lookup would produce an n+1 query? Righ now if I render out the sql in a rails console I'm not seeing multiple selects for associative rows. 

```sql
select
	"marketplace_products"."id" as t0_r0,
	"marketplace_products"."marketplace_id" as t0_r1,
	"marketplace_products"."name" as t0_r2,
	"marketplace_products"."price_cents" as t0_r3,
	"marketplace_products"."price_currency" as t0_r4,
	"marketplace_products"."created_at" as t0_r5,
	"marketplace_products"."updated_at" as t0_r6,
	"marketplace_products"."discarded_at" as t0_r7,
	"marketplace_products"."servings" as t0_r8,
	"action_text_rich_texts"."id" as t1_r0,
	"action_text_rich_texts"."name" as t1_r1,
	"action_text_rich_texts"."body" as t1_r2,
	"action_text_rich_texts"."record_type" as t1_r3,
	"action_text_rich_texts"."record_id" as t1_r4,
	"action_text_rich_texts"."created_at" as t1_r5,
	"action_text_rich_texts"."updated_at" as t1_r6
from
	"marketplace_products"
	inner join "marketplace_product_tags" on "marketplace_products"."id" = "marketplace_product_tags"."product_id"
	left outer join "action_text_rich_texts" on "action_text_rich_texts"."record_type" = 'Marketplace::Product'
		and "action_text_rich_texts"."name" = 'description'
		and "action_text_rich_texts"."record_id" = "marketplace_products"."id"
where
	"marketplace_product_tags"."tag_id" = 'b8d56ff4-8326-4a9b-8576-9e4b677bc4c4'
	and "marketplace_products"."discarded_at" is null
	/* loading for pp */
order by
	"marketplace_products"."name" asc
limit 11;
```

rosschapman:
@anaulin I'm going to merge for the dopamine hit but if you're still chewing on anything please leave it here and I'll address it later.

app/furniture/marketplace/menu_component.html.erb
anaulin: An N+1 situation requires two separate queries (one that is repeated N times, plus the outer one), so you would not see an N+1 query by looking at the SQL of one query. You would need to have a look at the implementation of the `ProductComponent`, which is what you are rendering inside the loop, and see if it is going to create any additional queries.

Also, we can add https://github.com/flyerhzm/bullet to the project and help it guide us when it thinks there are N+1 queries (although note that it is not, err, bullet-proof?).

rosschapman: Oh oh oh right, the risk of n+1 occurs with potential queries for each item from this collection as we loop. `bullet` gem looks useful. Sounds like similar amount of caution/skepticism we reserve for `strong_migrations`'s suggestions 😅. 

I'll take a pass in another PR adding the bullet gem and lifting this heavy lifting to a better place for Law of Demeter and n+1 code hazards.

PR #2441, opened 2024-06-05
Adds Bullet gem and adds new Product scopes and associations on Tag

rosschapman:


app/furniture/marketplace/product.rb
```diff
@@ -44,6 +44,8 @@ class Product < Record
     scope :with_menu_tag, -> { joins(:tags).merge(Tag.menu_tag) }
     scope :without_menu_tag, -> { where.not(id: with_menu_tag) }
     scope :sort_alpha, -> { order(name: :asc) }
+    scope :menu_tag_product, -> { with_all_rich_text.merge(unarchived).merge(sort_alpha) }
+    scope :menu_product, -> { with_all_rich_text.merge(unarchived).merge(without_menu_tag).merge(sort_alpha) }
```
anaulin, line 48: A few things about these two new scopes raise questions for me:
* why are you `merge`ing scopes from the same model? i don't think you need nor want to do that; you can just do `{... unarchived.without_menu_tag.sort_alpha ...}`
* do we want to encode eager-loading as part of a scope? we don't typically do that, because it makes the scope only suitable for that one particular use case, and bloated for other purposes; in this example, if you wanted to count the menu_products, this scope would be unnecessarily eager-loading all rich text in order to do that count
* these are scopes on a Product model, so adding `product` to the scope names seems redundant (i.e. `Product.menu` is a better scope name than `Product.menu_product`)
* it seems like the `menu_tag_product` scope is defined as a superset of the `menu_product`, which doesn't seem right

app/furniture/marketplace/tag.rb
```diff
@@ -5,6 +5,8 @@ class Tag < Record
     belongs_to :marketplace, inverse_of: :tags
     has_many :product_tags, inverse_of: :tag, dependent: :destroy
     has_many :products, through: :product_tags, inverse_of: :tags
+    has_many :menu_tag_products, -> { menu_tag_product }, through: :product_tags, source: :product, inverse_of: :tags
+    has_many :menu_products, -> { menu_product }, through: :product_tags, source: :product, inverse_of: :tags
```
anaulin, line 9: I see the `menu_tag_products` relationship used in this PR, but not `menu_products`.

I definitely do not want to use a relationship that uses a scope that includes eager-loading, that doesn't seem right to me.

anaulin (changes requested):
If you separate this PR into two, we can merge the addition of Bullet right away. It feels like the scopes need more discussion.

app/furniture/marketplace/product.rb
rosschapman: > do we want to encode eager-loading as part of a scope

I think this makes sense to do now that I've moved the scope to methods on the component

app/furniture/marketplace/menu_component.rb
rosschapman, line 1: @anaulin Here's another stab where these queries (scope combos) live on the component since they are very specific to the menu interface. Notionally it feels like heavy lifting for a "view" component, but it would be the same if we hoisted this logic to the component. Maybe something like a `MarketplaceMenu` would be a domain object model to move toward to remove this concern from the view layer.

anaulin: I like this a lot. If these scope combinations are only used in this component, putting them in the component seems great; it makes the ERB for the component cleaner without cluttering the model scopes.

PR #2444, opened 2024-06-06
Adds Bullet gem

rosschapman:
Tools for lazy programmers! (I mean "lazy" in the exalted sense of not-too-clever.)

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -2,12 +2,12 @@ require "active_support/core_ext/integer/time"
 
 Rails.application.configure do
   config.after_initialize do
-    Bullet.enable        = true
-    Bullet.alert         = true
+    Bullet.enable = true
+    Bullet.alert = true
     Bullet.bullet_logger = true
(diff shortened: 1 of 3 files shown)


config/environments/development.rb
```diff
@@ -1,6 +1,15 @@
...
 
 Rails.application.configure do
+  config.after_initialize do
+    Bullet.enable = true
+    Bullet.alert = true
+    Bullet.bullet_logger = true
+    Bullet.console = true
+    Bullet.rails_logger = true
+    Bullet.add_footer = true
+  end
```
rosschapman, line 11: These are auto-gen default configs. I'm open to changing them/don't have a strong opinion (yet).

anaulin: Seems fine to me. We can tweak these as we work with the gem and learn more about it and develop Opinions ™️.

anaulin (approved):
Thank you! 🚢

PR #2570, opened 2024-09-12
Does this fix CI?

zspencer:


anaulin (approved):
This fixes one of the things that are wrong with CI, so I think it is def worth merging!

PR #2571, opened 2024-09-12
Revert "Bump shoulda-matchers from 6.2.0 to 6.4.0"

anaulin:
Reverts zinc-collective/convene#2563 because it broke the build, seemingly?


anaulin:
Well, that wasn't it.

PR #2572, opened 2024-09-12
For some reason, the Gemfile.lock file keeps changing

zspencer:
@anaulin said that upgrading Bundler and running bundle install should fix it, so that's what I did.


anaulin (approved):
If I said it, it must be plausible!

PR #2573, opened 2024-09-12
Ensure "Remember Me for Stripe Link" Checkbox is unchecked

zspencer:
At least one of the CI failures appears to be caused by changes to the Stripe Checkout flow which now defaults to asking them to save their Stripe Link contact info


anaulin (approved):
Thanks for fighting the good fight!

PR #2634, opened 2024-10-06
[2626] move actions after order payment to after_commit hook

Kroch4ka:


spec/furniture/marketplace/finalize_order_payment_job_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    it "create order events" do
+      expect { action }.to change { order.events.reload.size }.from(0).to(3)
+    end
+
+    it "call jobs" do
+      action
+      expect(Marketplace::SplitJob).to have_received(:perform_later)
+      expect(mailer_double).to have_received(:deliver_later).exactly(2).times
```
zspencer, line 26: `rspec-rails` has [`have_enqueued_mail`](https://www.rubydoc.info/gems/rspec-rails/RSpec%2FRails%2FMatchers:have_enqueued_mail) and [`have_enqueued_job`](https://www.rubydoc.info/gems/rspec-rails/RSpec%2FRails%2FMatchers%3Ahave_enqueued_job) matchers that will allow you to make these assertions without stubbing.

zspencer (commented):
I think this is a good first stab at this, but there's a few adjustments that will want to be made before merging:

I think the strategy of splitting the post-order update items into a background job is a reasonable idea; but doesn't resolve the issues that emerge when we can't enqueue background jobs.

I think instead what we want is to wrap [`Marketplace::StripeEventsController` lines 23 through 33](https://github.com/zinc-collective/convene/blob/f8bced0cff085e1b1eaca94ae93ceb47496bd164/app/furniture/marketplace/stripe_events_controller.rb#L23-L33) in a [transaction](https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html) and rollback if any exception is thrown within the transaction block.

This way the jobs won't get enqueued and the order won't get marked paid until we successfully process the webhook.

Thoughts?

spec/furniture/marketplace/finalize_order_payment_job_spec.rb
Kroch4ka: I think an alternative solution might be to use the after_commit_everywhere gem.
image

zspencer: I think for now the transaction should be sufficient; since any failing step will revert the transaction and rollback any database changes. We may want to get more thoughtful if things start going weird; but I am OK with a basic transaction / rollback for now.

zspencer: (That said, I really like how the `after_commit_anywhere` gem keeps the `after_commit` to this particular place, rather than relying on a global `after_commit` that checks for a status change!)

zspencer:
Hey @Kroch4ka , I'm going to close this for now, but feel free to re-open if you have an additional commit.

Issue #2641, opened 2024-10-10
Delivery fee can be a percentage of the order


zspencer:
For Catering, Piikup wants to charge a %age of the total order. For small orders, she wants a $10 minimum.

The way I think that makes sense to do this is to allow each delivery area to be either a $amount or a %age amount; and label the delivery areas "Catering to San Francisco" or "Personal Plates in East Bay"

zspencer:
I've let #832 know that this feature exists.

PR #2643, opened 2024-10-10
`Marketplace`: `DeliveryArea` fees as a percentage

zspencer:
- https://github.com/zinc-collective/convene/issues/2641

Todo:
- [x] A `DeliveryArea` may have a delivery fee as a percentage
- [x] `Orders` to a `DeliveryArea` with a Percentage Fee calculate the fee correctly.
- [x] `Shoppers` see a description of the Delivery Fee when building their Order.

April has adjusted her fee schedule, and catering orders are now 12% and regular orders are $10 flat. This simplifies things quite a bit; hooray! But alas, we had already built out stuff to support different fees for every vendor and across regions (i.e. SF it's $20, Oakland it's $10)

I think it's probably for the best to leave that flexibility in for now, and have updated it so we now have another attribute for how to calculate fees.

I've made the Executive Decision to treat these fees as cumulative (i.e. $10 plus 12%) if both the `DeliveryArea#price` and the `DeliveryArea#fee_as_percentage` are set; rather than trying to figure out how to break ties or validate only one is set or whatever. This may be a poor decision.

### DIFF
--- a/app/furniture/marketplace/cart/delivery_area_component.html.erb
+++ b/app/furniture/marketplace/cart/delivery_area_component.html.erb
@@ -9,6 +9,7 @@
           <%- if single_delivery_area? %>
             <div><small>Orders outside of this location will be subject to cancellation.</small></div>
           <% end %>
+          <div>Delivery Fee: <%= fee_description %></div>
         </div>
       </div>
       <div class="text-right">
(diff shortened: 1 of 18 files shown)


anaulin:
> I've made the Executive Decision to treat these fees as cumulative (i.e. $10 plus 12%) if both the DeliveryArea#price and the DeliveryArea#fee_as_percentage are set; rather than trying to figure out how to break ties or validate only one is set or whatever. This may be a poor decision.

This sounds like a sensible choice to me, and in fact I think in some cases that might be the case. (Sometimes for delivery there is a flat fee + percentage, if the order is large.)

anaulin (approved):
🚢 🇮🇹

Issue #2639, opened 2024-10-10
Set the reply-to on order emails to the seller / buyer / distributor


zspencer:
When I placed an order last week, the email included a reply-to that was to convene-support; so when April acknowledged the order it got sent to our support list instead of me-the-buyer.

It would probably be good to set the reply-to on our [`OrderPlaced`](https://github.com/zinc-collective/convene/blob/main/app/furniture/marketplace/order/received_mailer.rb) and [`OrderReceived`](https://github.com/zinc-collective/convene/blob/main/app/furniture/marketplace/order/received_mailer.rb) emails to the other party.

zspencer:
this should make order emails a bit more clear.

PR #2644, opened 2024-10-10
`Marketplace`: Set `reply_to` on `Order::Mailer`s

zspencer:
- https://github.com/zinc-collective/convene/issues/2639

When I placed my order last week, the replies started going to Convene Support, rather than the customer / seller.

This should reduce that happening.


anaulin (approved):
👍🏼

Issue #2664, opened 2024-10-17
The Auto split is not sending the full amount


zspencer:
<img width="185" alt="Screenshot 2024-10-16 at 6 46 37 PM" src="https://github.com/user-attachments/assets/a4eac881-9e90-47b0-9672-7358552944b3">

These should be dollars, not cents 🤦

PR #2678, opened 2024-10-24
🐞🔨 `Marketplace`: Transfer the `Vendor`'s share of the `Order#amount` correctly

zspencer:
Fixes https://github.com/zinc-collective/convene/issues/2664

When we built this initially, we didn't do enough integration testing to confirm that the amount being split was right.

So we've been sending the *dollar amount* as *cents* into the downstream accounts, which is not ideal.

This fixes that.


anaulin (approved):
Thanks for fixing!

PR #2822, opened 2025-02-06
Capture console errors in Sentry for debugging

zspencer:
The `TypeError: Load failed` error is a bit of  a mess to diagnose and debug. I could take some guesses (Turbo, for example, may be doing pre-fetching) but rather than (potentially) spinning my wheels guessing I'm going to get some more data before digging in.

See:
- https://github.com/zinc-collective/convene/issues/2705
- https://zinc-collective.sentry.io/issues/6036564075/
- https://sentry.io/answers/load-failed-javascript/

### DIFF
--- a/app/javascript/sentry.js
+++ b/app/javascript/sentry.js
@@ -8,7 +8,10 @@ export function initializeSentry() {
 
   Sentry.init({
     dsn: dsn,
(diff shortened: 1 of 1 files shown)


anaulin (approved):
Thank you!

PR #3058, opened 2025-08-26
Marketplace: Handle insufficient balances gracefully

zspencer:
Fixes https://github.com/zinc-collective/convene/issues/3057

Payments take time to clear, so if a big order comes in or the balance is low we can't actually split immediately.

It may also make sense for us to wait to do the split until end-of-day in general, rather than immediately upon receiving the payment...

In cases where the pending and available balance are high enough, we re-enqueue and don't cause an error (or tick the "5 tries until the job dies") counter.

If the pending and available balance together won't cover it, we raise an error and tick that counter up, which may result in un-split orders; but does mean we don't keep trying to split an order that will never be funded.


zspencer:
The failing test is resolved in this PR: https://github.com/zinc-collective/convene/pull/3059

anaulin (approved):

